PR #25, opened 2015-01-05
introduce ErrPipelineReplicationNotSupported for transports

stapelberg:
This is necessary for transports which don’t support raft pipelining —
otherwise, stderr will be flooded with error messages about that fact.

A good reason not to support explicit pipelining is basing your
transport on net/http, which internally re-uses connections, so explicit
pipelining doesn’t give you any advantages, it just needs more lines of
code to implement.


armon:
Makes sense! Thanks!

PR #29, opened 2015-01-12
add FileStore (like InmemStore, but persistent)

stapelberg:
This is a trivial implementation of a persistent logstore. It mostly follows InmemStore, except it writes to files (one file per log entry, one file per stablestore key).

As you write in the README, it may be desirable to avoid the cgo dependency which raft-mdb brings, which is why I wrote this ;).

With regards to stability, I’m running this code in a pet project of mine without noticing any issues. Additionally, I’ve swapped the InmemStore for my FileStore in all *_test.go files in the raft repository and successfully ran the tests.

With regards to performance, I’ve run a couple of benchmarks (end-to-end messages/s in the actual application I’m working on) using raft-mdb vs. this FileStore, with various little tweaks:

All tests were run by measuring the messages/s over 13 runs, then averaging the results. The underlying storage is an INTEL SSDSC2BP48 (480G consumer SSD).

```
raft-mdb+sync+cache:   557 msgs/s (recommended backend)
filestore:             736 msgs/s (no fsync!)
filestore+cache:      1131 msgs/s (no fsync!)
filestore+sync:        418 msgs/s
filestore+sync+cache:  516 msgs/s
filestore+rename:      718 msgs/s (!)
```

filestore+cache is FileStore, but with a proof-of-concept cache: every log entry is kept in memory and `GetLog()` just copies the log entry instead of reading from disk. This is obviously much faster, but a real cache would need to be developed. I’m thinking keeping `config.MaxAppendEntries` in a ringbuffer should fit the access pattern quite well.

filestore+sync is FileStore, but with `defer f.Sync()` after `defer f.Close()` in `StoreLogs()`.

filestore+sync+cache is the combination of both of the above.

filestore+rename is FileStore, but writing into a temporary file which is then renamed to its final path. I’m curious to hear what you have to say about that. The semantics this code guarantees are that a log entry is either fully present or not present at all (in case of power loss for example). AIUI, the raft protocol should be able to cope with this situation. With regards to performance, this outperforms the current raft-mdb (1.2x).

Perhaps it would even make sense to replace InmemStore by FileStore entirely — you’re saying InmemStore should only be used for unit tests, and FileStore can do that, too. That way, people wouldn’t even have the chance to abuse InmemStore and use it in production :).

But that can wait for follow-up commits. For now, I’m mostly interested to see whether you’d want to merge this? I don’t think every project (which doesn’t want to use raft-mdb) should need to implement this independently :).

### DIFF
--- a/file_store.go
+++ b/file_store.go
@@ -1,0 +1,208 @@
+package raft
+
+import (
+	"encoding/gob"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+// FileStore implements the LogStore and StableStore interface.
+type FileStore struct {
+	l         sync.RWMutex
+	lowIndex  uint64
+	highIndex uint64
+	dir       string
+}
+
+type uint64Slice []uint64
+
+func (p uint64Slice) Len() int           { return len(p) }
+func (p uint64Slice) Less(i, j int) bool { return p[i] < p[j] }
+func (p uint64Slice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+func NewFileStore(dir string) (*FileStore, error) {
+	if err := os.MkdirAll(filepath.Join(dir, "logs"), 0700); err != nil {
+		return nil, err
+	}
+
+	if err := os.MkdirAll(filepath.Join(dir, "stable"), 0700); err != nil {
+		return nil, err
+	}
+
+	return &FileStore{
+		dir: dir,
+	}, nil
+}
+
+// GetAll returns all indexes that are currently present in the log store. This
+// is NOT part of the LogStore interface — we use it when snapshotting.
+func (s *FileStore) GetAll() ([]uint64, error) {
+	var indexes []uint64
+	dir, err := os.Open(filepath.Join(s.dir, "logs"))
+	if err != nil {
+		return indexes, err
+	}
+	defer dir.Close()
+
+	names, err := dir.Readdirnames(-1)
+	if err != nil {
+		return indexes, err
+	}
+
+	for _, name := range names {
+		if !strings.HasPrefix(name, "entry.") {
+			continue
+		}
+
+		dot := strings.LastIndex(name, ".")
+		if dot == -1 {
+			continue
+		}
+
+		index, err := strconv.ParseInt(name[dot+1:], 0, 64)
+		if err != nil {
+			return indexes, fmt.Errorf("Unexpected filename, does not confirm to entry.%%d: %q. Parse error: %v", name, err)
+		}
+
+		indexes = append(indexes, uint64(index))
+	}
+
+	sort.Sort(uint64Slice(indexes))
+
+	return indexes, nil
+}
+
+// FirstIndex implements the LogStore interface.
+func (s *FileStore) FirstIndex() (uint64, error) {
+	s.l.RLock()
+	defer s.l.RUnlock()
+	return s.lowIndex, nil
+}
+
+// LastIndex implements the LogStore interface.
+func (s *FileStore) LastIndex() (uint64, error) {
+	s.l.RLock()
+	defer s.l.RUnlock()
+	return s.highIndex, nil
+}
+
+// GetLog implements the LogStore interface.
+func (s *FileStore) GetLog(index uint64, rlog *Log) error {
+	s.l.RLock()
+	defer s.l.RUnlock()
+	f, err := os.Open(filepath.Join(s.dir, "logs/entry."+strconv.FormatUint(index, 10)))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return ErrLogNotFound
+		}
+		return err
+	}
+	defer f.Close()
+
+	var elog Log
+	if err := gob.NewDecoder(f).Decode(&elog); err != nil {
+		return err
+	}
+	*rlog = elog
+	return nil
+}
+
+// StoreLog implements the LogStore interface.
+func (s *FileStore) StoreLog(log *Log) error {
+	return s.StoreLogs([]*Log{log})
+}
+
+// StoreLogs implements the LogStore interface.
+func (s *FileStore) StoreLogs(logs []*Log) error {
+	s.l.Lock()
+	defer s.l.Unlock()
+
+	for _, entry := range logs {
+		f, err := os.Create(filepath.Join(s.dir, "logs/entry."+strconv.FormatUint(entry.Index, 10)))
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


armon:
I think something like this is a good idea, but I think it should be a different project, like "raft-mdb". The only store in this project is just for testing purposes. I don't think production grade backends should necessarily live in this repo.

stapelberg:
> I think something like this is a good idea, but I think it should be a different project, like "raft-mdb". The only store in this project is just for testing purposes. I don't think production grade backends should necessarily live in this repo.

I’m curious about the rationale. Why should backends be kept separate from the rest of the code?

armon:
Since the backends are pluggable, there can be many implementations. I'd prefer not to put them all in this repo, and instead link to the various usable backends.

stapelberg:
> I'd prefer not to put them all in this repo, and instead link to the various usable backends.

In principle, I agree and don’t really care about where this code lives. However, there are a couple of things that I do care about, which I don’t yet understand how they should work:

1. The intention of submitting this pull request was that you, the upstream authors of hashicorp/raft, review it. That way, the code’s status would change from “how that guy implemented a logstore” to “an official logstore implementation”, thus encouraging re-use by other projects. Would you create e.g. hashicorp/raft-file so that I can send a pull request for that?
2. Furthermore, whenever there is an API change, I don’t want to have to update the logstore code, but it would be better if the same commit would just update all the logstores accordingly. This becomes harder when you have multiple repositories, but not infeasible. Would you take care of updating the logstores in such a case?
3. What is the workflow with regards to testing the various logstores after any changes to the raft library? Could we run the hashicorp/raft tests with each of the different logstore implementations? (That would make them a test-time dependency, which at least in Debian implies build-time dependency.)
4. If the answer to the previous question is that logstores should be so separate that raft changes don’t need to test them, how can we still ensure that all logstores pass a common set of tests? Could there be like a hashicorp/raft/testhelper which contains tests that the logstore repository can just import and run?

Thanks in advance for clarifying these points :).

armon:
1. In many ways, all this is exactly what we want to avoid. For now, `raft-mdb` is the hashicorp official implementation, so we don't want to be burdened by maintaining backends we don't actually use.
2. For the most part, the API is extremely stable so it's not really an issue. 
3+4. We could always add a meta-package that helps to test the well-known backends against the latest raft if that is something the community was interested in.

stapelberg:
Seems like you’re set on not picking this backend up :(.

Would you be willing to do a review of the backend code nevertheless in the future, when we publish it as part of our app? It’s only a 170 line file and I’d value your opinion on it.

rbg:
@Michael Why not create a separate repo that is just raft-filestore (akin to raft-mdb) ?? 

>

stapelberg:
Creating a raft-filestore repository is what I meant by “publishing it as part of our app”. Still, I’d like to get this reviewed.

file_store.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return ErrLogNotFound
+		}
+		return err
+	}
+	defer f.Close()
+
+	var elog Log
+	if err := gob.NewDecoder(f).Decode(&elog); err != nil {
```
armon: You can decode directly into `rlog` to avoid  a copy

file_store.go
```diff
@@ -0,0 +1,208 @@
...
+	return ioutil.WriteFile(filepath.Join(s.dir, "stable", "key."+string(key)), val, 0600)
+}
+
+// Get implements the StableStore interface.
+func (s *FileStore) Get(key []byte) ([]byte, error) {
+	s.l.RLock()
+	defer s.l.RUnlock()
+	b, err := ioutil.ReadFile(filepath.Join(s.dir, "stable", "key."+string(key)))
+	if err != nil && os.IsNotExist(err) {
+		return []byte{}, fmt.Errorf("not found")
```
armon: I think no error is expected when the key is not set

file_store.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// SetUint64 implements the StableStore interface.
+func (s *FileStore) SetUint64(key []byte, val uint64) error {
+	return s.Set(key, []byte(strconv.FormatUint(val, 10)))
+}
+
+// GetUint64 implements the StableStore interface.
+func (s *FileStore) GetUint64(key []byte) (uint64, error) {
+	b, err := s.Get(key)
```
armon: The not found case should be be handled by returning 0

file_store.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	return &FileStore{
+		dir: dir,
+	}, nil
+}
+
+// GetAll returns all indexes that are currently present in the log store. This
+// is NOT part of the LogStore interface — we use it when snapshotting.
+func (s *FileStore) GetAll() ([]uint64, error) {
```
armon: Not sure why this needed?

file_store.go
```diff
@@ -0,0 +1,208 @@
...
+	if err := os.MkdirAll(filepath.Join(dir, "logs"), 0700); err != nil {
+		return nil, err
+	}
+
+	if err := os.MkdirAll(filepath.Join(dir, "stable"), 0700); err != nil {
+		return nil, err
+	}
+
+	return &FileStore{
+		dir: dir,
```
armon: Both `lowIndex` and `highIndex` must be recovered from disk. Entirely possible they are not zero (e.g. data is already on disk).

armon:
Overall, I think this approach will work. I think there are some minor issues, like ensuring that the semantics match what LogStore expects (when to return errors, etc), as well as recovering existing data, but it should work! Feel free to ask for further feedback, but since we don't intend to merge, I'm going to close this PR.

file_store.go
stapelberg: The only downside from not recovering `lowIndex` and `highIndex` is that the node will receive all data again (via snapshots and log replay), i.e. higher bandwidth usage, disk usage and longer recovery time, right?

file_store.go
stapelberg: True, this is equal to just ranging over `FirstIndex()` to `LastIndex()`. Only realized that afterwards :).

file_store.go
stapelberg: Sent https://github.com/hashicorp/raft/pull/33 to clarify that in the documentation.

stapelberg:
Thanks a bunch for the review! I posted one follow-up question with regards to recovering the lowIndex and highIndex values in the appropriate line note.

file_store.go
armon: If all the nodes shutdown and restart (e.g. power loss to DC), then all of them will assume they have no data (high index is zero). This would be a "data loss" scenario which is easily avoidable. The indexes must be restored to prevent data loss and electing the correct leader with the most recent data.

PR #30, opened 2015-01-13
Add a ringbuffer as a cache for log entries.

stapelberg:
Raft first writes n entries to the LogStore, waits for them to be committed, then reads them again. By storing the last config.MaxAppendEntries in a cache, we get a 1.5x performance win in throughput.

I’ve tested the performance of this change by sending 1000 messages through my toy raft application and measuring how long they took to arrive. Without this cache, I’m seeing 718 messages/s (averaged over 13 runs), whereas with the cache I’m seeing 1105 messages/s (averaged over 13 runs).

### DIFF
--- a/log_cache.go
+++ b/log_cache.go
@@ -44,19 +44,21 @@ func (c *logCache) getLogFromCache(logidx uint64) (*Log, bool) {
 	return log, log.Index == logidx
 }
 
-// cacheLog should be called with strictly monotonically increasing logidx
+// cacheLogs should be called with strictly monotonically increasing logidx
 // values, otherwise the cache will not be effective.
-func (c *logCache) cacheLog(logidx uint64, log *Log) {
+func (c *logCache) cacheLogs(logs []*Log) {
 	c.l.Lock()
 	defer c.l.Unlock()
 
-	if len(c.cache) < cap(c.cache) {
-		c.cache = append(c.cache, log)
-	} else {
-		c.cache[c.current] = log
+	for _, l := range logs {
+		if len(c.cache) < cap(c.cache) {
+			c.cache = append(c.cache, l)
+		} else {
+			c.cache[c.current] = l
+		}
+		c.lastlogidx = l.Index
+		c.current = (c.current + 1) % cap(c.cache)
 	}
-	c.lastlogidx = logidx
-	c.current = (c.current + 1) % cap(c.cache)
 }
 
 func (c *logCache) GetLog(logidx uint64, log *Log) error {
@@ -72,9 +74,7 @@ func (c *logCache) StoreLog(log *Log) error {
 }
 
 func (c *logCache) StoreLogs(logs []*Log) error {
-	for _, l := range logs {
-		c.cacheLog(l.Index, l)
-	}
+	c.cacheLogs(logs)
 	return c.store.StoreLogs(logs)
 }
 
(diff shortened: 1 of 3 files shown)


armon:
Great idea! I will add some suggestions to the PR.

raft.go
```diff
@@ -201,7 +201,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		leaderCh:      make(chan bool),
 		localAddr:     localAddr,
 		logger:        log.New(conf.LogOutput, "", log.LstdFlags),
-		logs:          logs,
+		logs:          newLogCache(conf.MaxAppendEntries, logs),
```
armon: Don't do this automatically, I think leave this to the caller.

log_cache.go
```diff
@@ -0,0 +1,98 @@
+package raft
+
+import (
+	"sync"
+)
+
+// logCache wraps a logstore with a ring buffer providing fast access to the
+// last n raft log entries.
+type logCache struct {
```
armon: I think we should expose this and allow a caller to wrap a LogStore of their choice. It might be the case that the LogStore is doing this internally already and there is no need to double cache.

log_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	return c.store.GetLog(logidx, log)
+}
+
+func (c *logCache) StoreLog(log *Log) error {
+	return c.StoreLogs([]*Log{log})
+}
+
+func (c *logCache) StoreLogs(logs []*Log) error {
+	for _, l := range logs {
+		c.cacheLog(l.Index, l)
```
armon: Why not make this `cacheLogs` and avoid having to acquire/release the lock multiple times?

armon:
I think we can simplify this a fair bit by initializing the ring length == capacity instead of trying to juggle the length vs capacity. Especially since capacity will be small (< 1K), it's probably faster to avoid all the branching logic anyways and spare the few extra KB of memory ahead of time. Likely we will reach the full capacity in only a few seconds anyways so the effort to save the memory is wasted anyways.

log_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// 'last' is the index of the element we cached last,
+	// its raft log index is 'lastlogidx'
+	last := (c.current - 1)
+	m := last - int(c.lastlogidx-logidx)
+
+	// See https://golang.org/issue/448 for why (m % n) is not enough.
+	n := cap(c.cache)
+	log := c.cache[((m%n)+n)%n]
```
armon: As an example, all of this juggling could be simplified if you can assume length == capacity

armon:
Good idea! This is a great optimization.

raft.go
stapelberg: Done.

log_cache.go
stapelberg: Done.

log_cache.go
stapelberg: Done.

log_cache.go
stapelberg: Done.

armon:
Thanks!

armon:
@stapelberg I pushed some changes to simplify the implementation. We can actually just use the index of the log itself to make this all much easier and faster.

PR #34, opened 2015-01-21
Use strings instead of net.Addr objects

stapelberg:
…on which only String() is ever called.

This frees callers from the need to implement their own type which wraps
a string, and also reduces the amount of code in the raft package.

I think the EncodePeer() and DecodePeer() methods may now also be
obsolete.

### DIFF
--- a/future.go
+++ b/future.go
@@ -1,7 +1,6 @@
 package raft
 
 import (
-	"net"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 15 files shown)


armon:
This makes a lot of sense. I'm going to let this sit until we can ship Consul 0.5 since I don't want to change it for now, but let's revisit soon.

krobertson:
@armon with 0.5 shipped, you think this is mergeable? I was looking at using your raft library with Apcera's NATS. It makes it easier with this merged.

armon:
@krobertson @stapelberg Merged! I did it manually to fix a type conflict, but its in master now.

PR #35, opened 2015-01-23
Allow users to pass in their own logger.

stapelberg:
For making raft log to github.com/golang/glog, it is necessary to
specify a custom logger, because glog expects you to set the Lshortfile
flag.

Aside from this specific example, I think in general it’s better to give
users more control over the logger instead of creating it for the user —
after all, it’s still just a single line of code.

### DIFF
--- a/config.go
+++ b/config.go
@@ -3,6 +3,7 @@ package raft
 import (
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


armon:
Thanks!

PR #37, opened 2015-01-31
Reusable benchmark package

ryanuber:
Adds a `github.com/hashicorp/raft/bench` package, which exposes some public methods which can be hooked into from any external raft backend store. This will make it much easier to compare backend performance apples to apples. All that is required is that the caller passes a valid `StableStore` / `LogStore`.


armon:
LGTM

PR #38, opened 2015-02-02
Add a simple HTTP/JSON-based transport.

stapelberg:


armon:
Anything that is in this repository will implicitly be stamped with HashiCorps approval / recommendation (this is how people perceive anything in our repos). I don't believe HTTP is a good transport for this, and would not want people to have that impression. The provided NetTransport makes much more sense for this use case.

PR #40, opened 2015-04-10
Constant error for ErrNotFound from the stable store

ryanuber:
This exposes the `ErrNotFound` var publicly, like we do for the rest of the common errors. This makes it easier for apps to detect a key not found error.


ryanuber:
Actually, it looks like we share the error value in the raft-mdb back-end, and do string comparison to check the error type. I'm not sure it makes sense for the raft back-ends to return an error from the raft package, so this might just be moot. Closing for now.

PR #46, opened 2015-06-29
Some minor doc comment fixes

cespare:
* Add many missing periods
* Correct some misspellings
* Fix references to nil strings


armon:
Thanks!

PR #47, opened 2015-06-30
More doc comment fixes

cespare:
Just applying the same changes as #46 to the rest of the package.

Should've done this before sending #46; sorry for the noise.


armon:
Thanks!

Issue #50, opened 2015-07-22
seeing lots of DuplicateVotes and elections with no outcome


superfell:
I'm seeing issues where if i start 3 nodes at the same time (e.g. all on the same host using goreman or similar), it takes a long time to get to an election that actually elects a leader, i see this pattern repeated

10:12:26 n3 | 2015/07/22 10:12:26 [WARN] raft: Election timeout reached, restarting election
10:12:26 n3 | 2015/07/22 10:12:26 [INFO] raft: Node at 127.0.0.1:5003 [Candidate] entering Candidate state
10:12:26 n1 | 2015/07/22 10:12:26 [WARN] raft: Election timeout reached, restarting election
10:12:26 n1 | 2015/07/22 10:12:26 [INFO] raft: Node at 127.0.0.1:5001 [Candidate] entering Candidate state
10:12:26 n2 | 2015/07/22 10:12:26 [WARN] raft: Election timeout reached, restarting election
10:12:26 n2 | 2015/07/22 10:12:26 [INFO] raft: Node at 127.0.0.1:5002 [Candidate] entering Candidate state
10:12:26 n1 | 2015/07/22 10:12:26 [DEBUG] raft: Votes needed: 2
10:12:26 n1 | 2015/07/22 10:12:26 [DEBUG] raft: Vote granted. Tally: 1
10:12:26 n3 | 2015/07/22 10:12:26 [DEBUG] raft: Votes needed: 2
10:12:26 n3 | 2015/07/22 10:12:26 [DEBUG] raft: Vote granted. Tally: 1
10:12:26 n3 | 2015/07/22 10:12:26 [INFO] raft: Duplicate RequestVote for same term: 77
10:12:26 n1 | 2015/07/22 10:12:26 [INFO] raft: Duplicate RequestVote for same term: 77
10:12:26 n2 | 2015/07/22 10:12:26 [DEBUG] raft: Votes needed: 2
10:12:26 n2 | 2015/07/22 10:12:26 [DEBUG] raft: Vote granted. Tally: 1
10:12:26 n1 | 2015/07/22 10:12:26 [INFO] raft: Duplicate RequestVote for same term: 77
10:12:26 n3 | 2015/07/22 10:12:26 [INFO] raft: Duplicate RequestVote for same term: 77
10:12:26 n2 | 2015/07/22 10:12:26 [INFO] raft: Duplicate RequestVote for same term: 77
[...]

PR #51, opened 2015-07-22
seed timeout randomness from crypto/rand

superfell:
to ensure different nodes use different sequences of random timeouts. Fixes issue #50

### DIFF
--- a/util.go
+++ b/util.go
@@ -15,6 +15,8 @@ import (
 
 var rnd = rand.New(rand.NewSource(newSeed()))
 
+// returns an int64 from a crypto random source
+// can be used to seed a source for a math/rand.
 func newSeed() int64 {
 	r, err := crand.Int(crand.Reader, big.NewInt(math.MaxInt64))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


util.go
```diff
@@ -5,18 +5,29 @@ import (
...
 	"math/rand"
 	"time"
 
 	"github.com/hashicorp/go-msgpack/codec"
 )
 
+var rnd = rand.New(rand.NewSource(newSeed()))
+
+func newSeed() int64 {
+	r, err := crand.Int(crand.Reader, big.NewInt(2^63))
```
armon: Instead of `2^63` lets use `math.MaxInt64`

util.go
```diff
@@ -5,18 +5,29 @@ import (
...
+	"math/big"
 	"math/rand"
 	"time"
 
 	"github.com/hashicorp/go-msgpack/codec"
 )
 
+var rnd = rand.New(rand.NewSource(newSeed()))
+
+func newSeed() int64 {
```
armon: Can you add a doc string and a test to verify that the first 1000 results are unique.

superfell:
Are you ok with this in multiple commits, or do you want me to squash it down to one ?

armon:
Thanks! This is great!

armon:
We may be able to use this in other places to use `rnd` instead of `rand`

PR #53, opened 2015-07-23
fix #52: race in randomTimeout()

### LINKED ISSUE #52: race in `randomTimeout()`
Looks like commit cda1ee83b012d9eb41c620123fcd1bb6163b4fc8 creates a race in `randomTimeout()`.  Showed up in InfluxDB CI.  https://gist.github.com/dgnorton/c11e59fecf2bcf367e76

dgnorton:


armon:
@dgnorton Good catch. I wonder if this is a better approach: http://golang.org/src/math/rand/rand.go?s=1010:1042#L230

armon:
My worry here is that we are re-seeding every time we call `randomTimeout` which will use up a lot of the entropy

dgnorton:
@armon Agreed.  I thought maybe there was a reason for using crypto/rand.  Want me to switch it?

armon:
@dgnorton We want to seed the pseudo-random function with real entropy from crypto/rand, but only once at start

dgnorton:
@armon I didn't understand what you meant by the link to `type Rand` in the stdlib.  Pushed another commit that makes the rng global again and added a mutex.  Let me know if you had something different in mind.

superfell:
The docs for math/rand say the default source is safe for concurrent use, its not clear that NewSource returns something with different characteristics. perhaps we should just seed the default source instead of creating a new one.

armon:
@dgnorton It was easier I think to just seed the global rand, which will help anywhere in the library we depend on it. It also solves the safety issue as that one is locked.

PR #72, opened 2015-11-19
Close transport on shutdown

josephglanville:
My application needs to re-use the bind address that the Raft transport is using after it's been shutdown.
The transport seems to eventually shutdown however waiting on the shutdown Future is not enough to ensure that it will be available to bind to. This patch explicitly closes the transport.

If there is a better way of achieving the behaviour I need would be happy to implement.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -411,6 +411,7 @@ func (r *Raft) Shutdown() Future {
 	if !r.shutdown {
 		close(r.shutdownCh)
 		r.shutdown = true
+		r.trans.Close()
 		r.setState(Shutdown)
 	}
 
--- a/transport.go
+++ b/transport.go
@@ -58,6 +58,9 @@ type Transport interface {
 	// disk IO. If a Transport does not support this, it can simply
 	// ignore the call, and push the heartbeat onto the Consumer channel.
 	SetHeartbeatHandler(cb func(rpc RPC))
+
+	// Close shuts down the transport
+	Close() error
 }
 
 // AppendPipeline is used for pipelining AppendEntries requests. It is used
--- a/inmem_transport.go
+++ b/inmem_transport.go
@@ -208,6 +208,11 @@ func (i *InmemTransport) DisconnectAll() {
 	i.pipelines = nil
 }
 
+func (i *InmemTransport) Close() error {
+	i.DisconnectAll()
+	return nil
+}
+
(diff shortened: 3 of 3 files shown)


transport.go
```diff
@@ -58,6 +58,9 @@ type Transport interface {
 	// disk IO. If a Transport does not support this, it can simply
 	// ignore the call, and push the heartbeat onto the Consumer channel.
 	SetHeartbeatHandler(cb func(rpc RPC))
+
+	// Close shuts down the transport
+	Close() error
```
conslo: You're adding a requirement to an exported interface, that breaks compatibility gaurantees (and I don't know how this project intends on handling breaking changes).

raft.go
```diff
@@ -411,6 +411,7 @@ func (r *Raft) Shutdown() Future {
 	if !r.shutdown {
 		close(r.shutdownCh)
 		r.shutdown = true
+		r.trans.Close()
```
conslo: Could you instead to a two-value type assertion `closeable, ok := r.trans.(closeableInterface)` and close if it worked? That would alleviate the need to change the exported interface.

If you do this, please update the appropriate comments/documentation to reflect "transport will be closed in x circumstances if closeable"

armon:
@josephglanville I agree with the other comments, can you just add a type assertion to check if Close is implemented so the interface does not need to change. Also why not just close the transport after the Raft shutdown?

jefferai:
If this issue is caused by using a custom `http.Transport` it may be useful to see what `go-cleanhttp` does: https://github.com/hashicorp/go-cleanhttp/blob/master/cleanhttp.go

abligh:
This is not the right approach (having just tried it).

Adding the `closeable` stuff (and type check) works fine. However, closing the transport within `shutdown` is not correct, and cause `TestRaft_RemoveLeader` to fail about 50% of the time. The reason is that the `RemovePeer` called with our ourself as the peer causes `Shutdown` to be called, which starts disabling the transport before the peer has actually been properly removed (i.e. before the config change has been committed across the network). Whilst I'm sure things sort themselves out eventually, the test fails as a result.

The correct time to shutdown the transport is after the shutdown RPC has returned. I'll see if I can knock something up.

abligh:
See pull request #92 for what I think is a fix.

josephglanville:
Closing in favour of #92

Issue #75, opened 2015-12-15
Include voter in vote debug logging


superfell:
The current logging around voting/elections is typically 

16:57:19.062392 [INFO] raft: Node at sv_4 [Candidate] entering Candidate state 
16:57:19.062889 [DEBUG] raft: Votes needed: 3 
16:57:19.062901 [DEBUG] raft: Vote granted. Tally: 1 
16:57:19.063332 [DEBUG] raft: Vote granted. Tally: 2 
16:57:19.063595 [DEBUG] raft: Vote granted. Tally: 3 
16:57:19.063632 [INFO] raft: Election won. Tally: 3 

It'd be useful if the Vote granted lined include the name of the node that granted the vote

PR #76, opened 2015-12-15
add voter to granted vote debug logging. fixes hashicorp/raft#75

superfell:
This adds the node that granted a vote to the candidate's debug log. This helps when investigating election related issues.


armon:
Thanks!

Issue #79, opened 2016-01-18
Removing the leader can lead to different log's applied to the remaining nodes


superfell:
While doing some stress testing we ran into this issue. If you do a RemovePeer of a node who happens to be the leader at the time, and is also working on other outstanding Apply'd entries, the remaining nodes in the cluster can end up seeing a mismatch in the entries that are applied to the FSMs. In particular some nodes will see one entry applied and other node never see this entry (depending on the entry this can lead to serious data issues).

The entry in question always appears to be an inflight entry that is the log entry after the RemovePeers log entry. e.g. in this run log entry 3827 was the RemovePeer entry, node 0's FSM saw this sequence of Apply's (term/index/data)
1.    3825: 7945C0282719297129F2A5F85B89E9D55373600FA459F8BA7BFD9D443488635A67
1.    3826: BF921247A73A9AFDAD5D2A37D34C4C49636C6353D0173AE6D29682C11685FD6E07
3.    3829: A74CC536489C8AFD48EAA6A5FB2063614E16E15EA13754B362A8A448A7A49CBB75
3.    3830: 6E98BE940FF0CFD71960DBA7C9B3674D18BA97ECAACC8B80DF7211F9702AFF77D7

but node 3's FSM saw these sequence of applies
1.    3825: 7945C0282719297129F2A5F85B89E9D55373600FA459F8BA7BFD9D443488635A67
1.    3826: BF921247A73A9AFDAD5D2A37D34C4C49636C6353D0173AE6D29682C11685FD6E07
__1.    3828: 7340F10D4B683B02B08A37077C90CC4B39A7812C4A87EF986190679031DDDB9526__
3.    3829: A74CC536489C8AFD48EAA6A5FB2063614E16E15EA13754B362A8A448A7A49CBB75
3.    3830: 6E98BE940FF0CFD71960DBA7C9B3674D18BA97ECAACC8B80DF7211F9702AFF77D7

Note the extra log entry at 3828 that Node 3 saw that Node 0 didn't. (Node 1 was the leader which was the node removed)

In addition, the client request that created log entry 3828 was given a success result.

We added some instrumentation of AppendEntry RPC calls to our transport, and from that log we see this sequence of AE's from the leader at this time
node from -> to, first/last index in message, leaderCommitIndex
rl_1 -> rl_2	    3823 -     3824 :     3820
rl_1 -> rl_4	    3825 -     3826 :     3822
rl_1 -> rl_2	    3825 -     3826 :     3822
[...]

PR #80, opened 2016-01-18
test LeaderRemoval while doing Apply's

superfell:
This change to the test TestRaft_RemoveLeader_NoShutdown demonstrates the issue raised in #79 

there's something of a timing issue, so you need to repeatedly run it until it fails, e.g.

`while go test -run RemoveLeader_ ; do :; done`


superfell:
When this fails you'll see something like
`raft_test.go:255: FSM log length mismatch: 86 84`

armon:
@superfell Merging, since I'm using this to try to work on the bug.

PR #81, opened 2016-01-18
Step down if we are processing a RemovePeer of the active leader

### LINKED ISSUE #79: Removing the leader can lead to different log's applied to the remaining nodes

armon:
Fixes #79. Currently if there is an inflight RemovePeer(Leader), the active leader will set the quorum size to 1 which causes all future transactions to immediately Apply and replicate to the existing peers. This causes undefined behavior. In a completely serial world, we would expect all transactions on the current leader after the RemovePeer(Leader) to fail, as its no longer a valid peer or leader.

This PR changes the behavior to track if we have an inflight command that would cause us to step down. If so, we preemptively fail all future Apply's with an `ErrNotLeader` which is used to convey that the current node is not the leader and that we have not even attempted replication of the entry.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -918,7 +918,7 @@ func (r *Raft) leaderLoop() {
 			for i := 0; i < n; i++ {
 				// Fail all future transactions once stepDown is on
 				if stepDown {
-					ready[i].respond(ErrLeadershipLost)
+					ready[i].respond(ErrNotLeader)
 					ready[i], ready[n-1] = ready[n-1], nil
 					n--
 					i--
--- a/raft.go
+++ b/raft.go
@@ -849,6 +849,14 @@ func (r *Raft) startReplication(peer string) {
 // leaderLoop is the hot loop for a leader. It is invoked
 // after all the various leader setup is done.
 func (r *Raft) leaderLoop() {
+	// stepDown is used to track if there is an inflight log that
+	// would cause us to lose leadership (specifically a RemovePeer of
+	// ourselves). If this is the case, we must not allow any logs to
+	// be processed in parallel, otherwise we are basing commit on
+	// only a single peer (ourself) and replicating to an undefined set
+	// of peers.
+	stepDown := false
+
 	lease := time.After(r.conf.LeaderLeaseTimeout)
 	for r.getState() == Leader {
 		select {
@@ -908,13 +916,24 @@ func (r *Raft) leaderLoop() {
 			// Handle any peer set changes
 			n := len(ready)
 			for i := 0; i < n; i++ {
+				// Fail all future transactions once stepDown is on
+				if stepDown {
+					ready[i].respond(ErrLeadershipLost)
+					ready[i], ready[n-1] = ready[n-1], nil
+					n--
+					i--
+					continue
+				}
+
 				// Special case AddPeer and RemovePeer
 				log := ready[i]
 				if log.log.Type != LogAddPeer && log.log.Type != LogRemovePeer {
 					continue
 				}
 
-				// Check if this log should be ignored
+				// Check if this log should be ignored. The logs can be
+				// reordered here since we have not yet assigned an index
+				// and are not violating any promises.
 				if !r.preparePeerChange(log) {
 					ready[i], ready[n-1] = ready[n-1], nil
 					n--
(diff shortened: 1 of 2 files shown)


armon:
@superfell Just pushed the extra guard. Running this for a few hundred iterations to see if anything crops up.

superfell:
great, I'll run it through our stress-test's as well.

armon:
@superfell I'm merging, but please re-open if there is an issue.

superfell:
Will do, thanks for jumping on this one so quickly.

PR #82, opened 2016-01-19
allow a node to vote for the current leader, fixes #74

### LINKED ISSUE #74: Node doesn't vote for current leader
Part of the voting rules is that a node won't vote for another node if it thinks there's an active leader. However it applies this rule even when the node asking for the vote is the node that the voter thinks is the leader. e.g. (here node lp_0 refuses to vote for lp_3 because it thinks lp_3 is the leader)

lp_0:17:03:33.954809 [WARN] raft: Rejecting vote from lp_3 since we have a leader: lp_3

Seems like if the node asking for the vote is the node we think is the leader we should go ahead and vote for it.

superfell:


armon:
Thanks!

PR #83, opened 2016-01-20
Set last contact on leader step down

armon:
Currently when the leader steps down, its last contact time will be from when it was last contacted as a follower, which may be either zero or a very long time ago. This fix updates the last contact time to when we stepped down, better reflecting the recency of the data.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -755,6 +755,13 @@ func (r *Raft) runLeader() {
 
 	// Cleanup state on step down
 	defer func() {
+		// Since we were the leader previously, we update our
+		// last contact time when we step down, so that we are not
+		// reporting a last contact time from before we were the
+		// leader. Otherwise, to a client it would seem our data
+		// is extremely stale.
+		r.setLastContact()
+
 		// Stop replication
 		for _, p := range r.leaderState.replState {
 			close(p.stopCh)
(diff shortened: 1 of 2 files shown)


raft.go
```diff
@@ -792,6 +799,11 @@ func (r *Raft) runLeader() {
 			select {
 			case notify <- false:
 			case <-r.shutdownCh:
+				// On shutdown, make a best effort but do not block
```
slackpad: Is this related to the last contact change or something else?

raft.go
```diff
@@ -755,6 +755,13 @@ func (r *Raft) runLeader() {
 
 	// Cleanup state on step down
 	defer func() {
+		// Since we were the leader previously, we update our
+		// last contact time when we step down, so that we are not
+		// reporting a last contact time from before we were the
+		// leader. Otherwise, to a client it would seem our data
+		// is extremely stale.
+		r.setLastContact()
```
slackpad: The only other thing I could think of that this would interact with is https://github.com/hashicorp/raft/blob/master/raft.go#L636, which is the heartbeat timeout calculation as a follower. The intent of that code is to wait for the heartbeat amount of time, so I don't think it matters if the time gets updated (since the heartbeat timer starts when we become a follower).

slackpad:
@armon noted two small comments, otherwise looks good.

superfell:
Looks good to me.

raft.go
armon: Its a race condition in one of the unit tests, basically during a shutdown we may fail to push to notify channel, so this tries to fix it without blocking.

Issue #84, opened 2016-02-09
Cleanup Meta Ticket


armon:
Here are the list of issues, grouped together if they might make sense as a single PR.

State Races
- [x] Raft state should use locks for any fields that are accessed together (e.g. index and term)

Multi-Row Fetching
- [ ] ~~Replace single row lookups with multi row lookups (LogStore / LogCache) (look at cases around log truncation)~~
- [x] Verify the current term has not changed when preparing/processing the AppendEntries message #136

Follower Replication:
- [x] `replicateTo` should verify leadership is current during looping
- [x] Check for any hot loops that do not break on `stopCh`

Change Inflight Tracking
- [x] Remove majorityQuorum
- [x] Inflight tracker should map Node -> Last Commit Index (match index)
- [x] Votes should be ignored from peers that are not part of peer set
- [x] `precommit` may not be necessary with new inflight (likely will be cleaned up via #117)

Improve Membership Tracking
- [x] Peer changes should have separate channel and do not pipeline (we don't want more than one peer change in flight at a time) #117 
- [x] Peers.json should track index and any AddPeer or RemovePeer are ignored from older indexes - #117

Crashes / Restart Issues
- [ ] ~~Panic with old snapshots #85~~
- [ ] ~~TrailingLogs set to 0 with restart bug #86~~

New Tests
- [x] Config change under loss of quorum: #127
- Setup cluster with {A, B, C, D}
- Assume leader is A
- Partition {C, D}
- Remove {B}
- Test should fail to remove B (quorum cannot be reached)

/cc: @superfell @ongardie @sean- @ryanuber @slackpad

ongardie:
Thanks for writing that up, @armon. I'll take a shot at "Change inflight tracking".

ongardie:
I'll start on 'precommit may not be necessary with new inflight' (though I only have about 20min right now, ugh).

slackpad:
Ok I'm going to go ahead and merge #117. Here's the list of remaining tasks I identified from that. Most of these are already marked inline with TODO comments:

* [ ] Fixup `NewRaft()` to understand legacy `LogAddPeer` and `LogRemovePeer` entries
* [ ] Work through live upgrade scenario now that we have a new config format and ignore old config messages (will we need to support the legacy `LogAddPeer` and `LogRemovePeer` to make the transition work?)
* [ ] Clean up observer interface to use new types (we could probably just use the new server structure here)
* [ ] Add staging -> voter promotion logic
[...]

PR #87, opened 2016-02-18
Replace inflight tracker with commitment tracker

ongardie:
Towards hashicorp/raft#84.

While I think this is the approach we want to take for commitment, I'm worried that applying this to master right away may result in a temporary loss in stability. This change depends a lot on how cluster membership changes are done, and I don't think those are where we want them to be yet. We have more changes planned for hashicorp/raft#84, so should we start a new branch? That'd keep master in its seemingly stable state with its known flaws, until we're more confident the new pieces fit in well together. With that in mind, I've updated and run the unit tests but haven't done any further system-wide testing.

## Commit message

inflight.go used to have two jobs: it buffered log futures until they were
committed, and it marked new log entries committed. This second job is what
commitment.go does, while raft.go now tracks its own uncommitted log futures.

Inflight used to track the replication factor of each log future/entry
independently. It kept a count of how many servers stored that entry, and how
many were needed to mark it committed. The primary issue is that a simple count
of how many servers store an entry is fragile with respect to membership changes,
where the identity of the voting servers might change over time. It was also a
fair amount of bookkeeping to track this information per entry.

The new 'commitment' module instead tracks the match index for each server: the
last entry that the server has on disk that agrees with this leader. This has
several advantages:
 - No information is kept per log entry.
 - Only the latest known membership configuration is used to mark entries
   committed, as described in the Raft paper/dissertation.
 - Servers that aren't voters are simply ignored for the purpose of commitment.
   This is useful for leaders that remove themselves today but would also be
   useful for more general non-voting members in the future.

Because no information is kept per log entry, the log futures can now be state
kept local to the leader goroutine, without any need for locking. This
simplified the interface to commitment.go (relative to inflight.go)
significantly, while only adding a few lines here and there to raft.go.


slackpad:
Hi @ongardie making my way through here now. I definitely agree we should integrate in a branch for now since the changes are somewhat interdependent, and there are folks who pull from master and won't appreciate the intermediate state. I created https://github.com/hashicorp/raft/tree/issue-84-integration so you can retarget this and related changes to that. I'll work to rebase that with any small changes that come in to master.

inflight.go
```diff
@@ -1,213 +0,0 @@
...
-		// Ignore if not in the map, as it may be committed already
-		return
-	}
-
-	// Check if we've satisfied the commit
-	if !op.policy.Commit() {
-		return
-	}
-
-	// Cannot commit if this is not the minimum inflight. This can happen
```
slackpad: So this was a consequence of coupling the quorum data with the inflight log entries, and has no manifestation in the new scheme, right?

commitment_test.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"testing"
+)
+
+// Returns a slice of server names of size n.
+func voters(n int) []string {
+	if n > 7 {
+		panic("only up to 7 servers implemented")
+	}
+	return []string{"s1", "s2", "s3", "s4", "s5", "s6", "s7"}[:n+1]
```
slackpad: This should just be `[:n]`

commitment_test.go
```diff
@@ -0,0 +1,159 @@
...
+	c.match("s2", 4)
+	c.match("s3", 4)
+
+	if c.getCommitIndex() != 4 {
+		t.Fatalf("should be able to commit startIndex once replicated to a quorum")
+	}
+}
+
+// With no voting members in the cluster, the most sane behavior is probably
+// to not not mark anything committed.
```
slackpad: "to not not"

commitment_test.go
```diff
@@ -0,0 +1,159 @@
...
+func voters(n int) []string {
+	if n > 7 {
+		panic("only up to 7 servers implemented")
+	}
+	return []string{"s1", "s2", "s3", "s4", "s5", "s6", "s7"}[:n+1]
+}
+
+// Tests setVoters() keeps matchIndexes where possible.
+func TestCommitment_setVoters(t *testing.T) {
+	commitCh := make(chan struct{}, 1)
```
slackpad: Would be good in these tests to assert that the `commitCh` got hit correctly (or at least the correct number of times).

commitment_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+// With no voting members in the cluster, the most sane behavior is probably
+// to not not mark anything committed.
+func TestCommitment_noVoterSanity(t *testing.T) {
+	commitCh := make(chan struct{}, 1)
+	c := newCommitment(commitCh, []string{}, 4)
+	c.match("s1", 10)
+	c.setVoters([]string{})
```
slackpad: In the case where you had voters and then lost them all I think the `commitIndex` will just stay where it was - might be worth adding that case in here as well.

raft.go
```diff
@@ -861,6 +868,7 @@ func (r *Raft) startReplication(peer string) {
 // leaderLoop is the hot loop for a leader. It is invoked
 // after all the various leader setup is done.
 func (r *Raft) leaderLoop() {
+	// TODO: reconsider
```
slackpad: What were you thinking here?

raft.go
```diff
@@ -748,7 +750,10 @@ func (r *Raft) runLeader() {
 
 	// Setup leader state
 	r.leaderState.commitCh = make(chan struct{}, 1)
-	r.leaderState.inflight = newInflight(r.leaderState.commitCh)
+	r.leaderState.commitment = newCommitment(r.leaderState.commitCh,
+		append([]string{r.localAddr}, r.peers...),
+		r.getLastIndex()+1)
```
slackpad: I'm probably not understanding something subtle - why does this increment by 1 here?

raft.go
```diff
@@ -879,17 +887,23 @@ func (r *Raft) leaderLoop() {
 			r.setState(Follower)
 
 		case <-r.leaderState.commitCh:
-			// Get the committed messages
-			committed := r.leaderState.inflight.Committed()
-			for e := committed.Front(); e != nil; e = e.Next() {
-				// Measure the commit time
+			// Process the newly committed entries
+			commitIndex := r.leaderState.commitment.getCommitIndex()
+			r.setCommitIndex(commitIndex)
```
slackpad: Ugh all these index helpers use the name `term` for the argument. We should probably fix those use.

raft.go
slackpad: Ok talked though this with @sean- and it makes sense now - this will be the first index that can be committed _under this term_. This is probably worth a comment here as well because it's super important to how this operates (I see that it is documented over in commitment.go).

commitment_test.go
ongardie: Nice catch, as this off-by-one masked a critical off-by-one in the code under test.

commitment.go
```diff
@@ -0,0 +1,100 @@
...
+	"sort"
+	"sync"
+)
+
+// Commitment is used to advance the leader's commit index. The leader and
+// replication goroutines report in newly written entries with Match(), and
+// this notifies on commitCh when the commit index has advanced.
+type commitment struct {
+	// protectes matchIndexes and commitIndex
+	sync.Mutex
```
armon: We can make this an `RWMutex`

commitment.go
```diff
@@ -0,0 +1,100 @@
...
+	c.matchIndexes = make(map[string]uint64, len(voters))
+	for _, voter := range voters {
+		c.matchIndexes[voter] = oldMatchIndexes[voter] // defaults to 0
+	}
+	c.recalculate()
+}
+
+// Called by leader after commitCh is notified
+func (c *commitment) getCommitIndex() uint64 {
+	c.Lock()
```
armon: `RLock` and `RUnlock` safe here

commitment.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Internal helper to calculate new commitIndex from matchIndexes.
+// Must be called with lock held.
+func (c *commitment) recalculate() {
+	if len(c.matchIndexes) == 0 {
+		return
+	}
+	var quorumMatchIndex uint64
+	{
+		matched := make([]uint64, 0, len(c.matchIndexes))
```
armon: Style: You can unindent this whole block, and then just define-assign `quorumMatchIndex` instead of pre-declaring it.

commitment.go
ongardie: Seems like a pre-mature optimization. Do you think there'd be any measurable gain?

raft.go
ongardie: Well, that comment doesn't make any sense to me. If the leader is removing itself, I don't think outstanding requests have to be treated any differently (they should just be aborted but that's already part of the step down path). I don't think this PR is the right place to handle this, so I'll start a separate GitHub issue.

commitment.go
armon: Reading through it'll be minor, doesn't look like a caller can spend much time in a critical section

raft.go
ongardie: PR #101 fixes these.

ongardie:
Made obsolete by #102, which retargets this onto the issue-84-integration branch.

PR #88, opened 2016-02-26
Index/Term's in raft state should be locked/get/set as a single thing

superfell:
fixes Issue #84 item 1

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1369,7 +1369,7 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 		last := a.Entries[n-1]
 
 		// Delete any conflicting entries
-		lastLogIdx, _ := r.getLastLog()
+		lastLogIdx := r.getLastLogIndexOnly()
 		if first.Index <= lastLogIdx {
 			r.logger.Printf("[WARN] raft: Clearing log suffix from %d to %d", first.Index, lastLogIdx)
 			if err := r.logs.DeleteRange(first.Index, lastLogIdx); err != nil {
@@ -1821,7 +1821,7 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
 	}
 
 	// Check if we have enough logs to truncate
-	lastLogIdx, _ := r.getLastLog()
+	lastLogIdx := r.getLastLogIndexOnly()
 	if lastLogIdx <= r.conf.TrailingLogs {
 		return nil
 	}
--- a/raft.go
+++ b/raft.go
@@ -235,8 +235,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 
 	// Restore the current term and the last log
 	r.setCurrentTerm(currentTerm)
-	r.setLastLogIndex(lastLog.Index)
-	r.setLastLogTerm(lastLog.Term)
+	r.setLastLog(lastLog.Index, lastLog.Term)
 
 	// Attempt to restore a snapshot if there are any
 	if err := r.restoreSnapshot(); err != nil {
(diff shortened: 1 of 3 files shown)


raft.go
```diff
@@ -1821,7 +1818,8 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
 	}
 
 	// Check if we have enough logs to truncate
-	if r.getLastLogIndex() <= r.conf.TrailingLogs {
+	lastLogIdx, _ := r.getLastLog()
```
ongardie: Nit: I think this could be r.getLastLogIndexOnly()

raft.go
```diff
@@ -1367,7 +1367,7 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 		last := a.Entries[n-1]
 
 		// Delete any conflicting entries
-		lastLogIdx := r.getLastLogIndex()
+		lastLogIdx, _ := r.getLastLog()
```
ongardie: Nit: getLastLogIndexOnly() here too?

ongardie:
This PR looks to me like a step in the right direction. I'm still worried about shared state across the library, but this PR will help with some issues. +1

state.go
```diff
@@ -83,20 +87,39 @@ func (r *raftState) setCurrentTerm(term uint64) {
...
-	return atomic.LoadUint64(&r.LastLogIndex)
+func (r *raftState) getLastLog() (index, term uint64) {
+	r.lastLock.Lock()
+	index = r.lastLogIndex
+	term = r.lastLogTerm
+	r.lastLock.Unlock()
+	return
+}
+
+func (r *raftState) getLastLogIndexOnly() uint64 {
```
slackpad: This one is a little funky - I'd prefer just having all callers call `log, _ := getLastLog()`.

raft.go
```diff
@@ -1829,7 +1827,7 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
 	// back from the head, which ever is further back. This ensures
 	// at least `TrailingLogs` entries, but does not allow logs
 	// after the snapshot to be removed.
-	maxLog := min(snapIdx, r.getLastLogIndex()-r.conf.TrailingLogs)
+	maxLog := min(snapIdx, lastLogIdx-r.conf.TrailingLogs)
```
slackpad: This was a good catch.

slackpad:
Just one tiny comment on this one. We should target this to https://github.com/hashicorp/raft/tree/issue-84-integration, and a rebase should make Travis CI less flaky with ancient Go 1.2 removed.

state.go
slackpad: Looking through the history now I realize you had it this way originally :-)

@ongardie were you worried about being able to more easily find places that don't fetch both together or was this just a style thing? It's not a huge deal either way but `idx, _` seems more idiomatic.

superfell: Yeah,i started with just the 2 item return version, but adding index only simplified its usage in replication.go. am happy to go back to just the 2 item version if you'd rather have just that.

slackpad: I'd prefer only the 2 item version. I think for future maintenance it's probably better to make people think about fetching them together as a unit as well.

ongardie:
I think this one should go into master. It's pretty harmless, so there's no need to stage it in the issue-84-integration branch only.

ongardie:
Heads up: this'll probably conflict with #101 but should be easy to resolve.

armon:
Reviewed. I agree with the existing comments, then LGTM!

superfell:
rebased to current master, lastLogIndexOnly() removed.

armon:
LGTM!

Issue #90, opened 2016-03-15
Various raft tests fail occasionally


abligh:
Background: I am testing a new transport and to get this to work need it to pass all tests. Unfortunately some tests fail occasionally with the existing InmemoryTransport.

Running
```
for i in `seq 1 50` ; do go test -run TestRaft_Voting ; done
```

I see the occasional failure, like:

```
--- FAIL: TestRaft_Voting (0.09s)
	raft_test.go:189: [WARN] Fully Connecting
	raft_test.go:93: [INFO] raft: Node at efecfb3b-81c1-a503-46c3-05fbf14221b6 [Follower] entering Follower state
	raft_test.go:93: [INFO] raft: Node at e6ba3ba4-a326-f7d0-e260-9d1cc55719d0 [Follower] entering Follower state
	raft_test.go:93: [INFO] raft: Node at 2696da81-b3ca-69e1-b09e-ef3f91fed6ea [Follower] entering Follower state
	raft_test.go:93: [WARN] raft: Heartbeat timeout reached, starting election
	raft_test.go:93: [INFO] raft: Node at e6ba3ba4-a326-f7d0-e260-9d1cc55719d0 [Candidate] entering Candidate state
	raft_test.go:93: [DEBUG] raft: Votes needed: 2
	raft_test.go:93: [DEBUG] raft: Vote granted from e6ba3ba4-a326-f7d0-e260-9d1cc55719d0. Tally: 1
	raft_test.go:93: [DEBUG] raft: Vote granted from efecfb3b-81c1-a503-46c3-05fbf14221b6. Tally: 2
	raft_test.go:93: [INFO] raft: Election won. Tally: 2
	raft_test.go:93: [INFO] raft: Node at e6ba3ba4-a326-f7d0-e260-9d1cc55719d0 [Leader] entering Leader state
	raft_test.go:93: [WARN] raft: Heartbeat timeout reached, starting election
	raft_test.go:93: [INFO] raft: Node at 2696da81-b3ca-69e1-b09e-ef3f91fed6ea [Candidate] entering Candidate state
	raft_test.go:93: [DEBUG] raft: Votes needed: 2
	raft_test.go:93: [DEBUG] raft: Vote granted from 2696da81-b3ca-69e1-b09e-ef3f91fed6ea. Tally: 1
	raft_test.go:93: [ERR] raft: peer 2696da81-b3ca-69e1-b09e-ef3f91fed6ea has newer term, stopping replication
[...]

PR #91, opened 2016-03-17
Reset heartbeat on a RequestVote RPC call

abligh:
Heartbeat timers are randomized. This means that (all other things being equal) they may elapse and cause a node to enter candidate state at any time. The heartbeat timer is currently reset when a candidate receives a valid `AppendEntries` RPC call.

This causes an issue as a node that has just voted for a candidate may immediately declare itself a candidate. For instance in a new 3 node cluster, Node A declares itself a candidate, Node A votes
for itself, and Node B votes for node A too. Shortly afterwards (but before an `AppendEntries` heartbeat is sent by Node A), Node B's heartbeat timer expires, and Node B starts an election.

This causes an issue within the test suite, characterised by about 1%-2% of test runs failing. In particular A is detected as the leader, and B is subsequently detected as the leader. Issue #90 describes this in more detail.

The raft specification (section 5.2 of the paper, section 3.4 of the thesis), says (my emphasis):

> A server remains in follower state as long as it receives valid RPCs from a leader **or candidate** 

It is thus intended that a node in receipt of a 'valid' `RequestVote` RPC from a candidate should not declare itself a candidate (but should remain in follower state) for at least the Heartbeat interval.

It isn't entirely clear what a 'valid' `RequestVote` RPC means, but certainly one that is voted on is valid, and this is the smallest change.

This one line change thus resets the heartbeat timer if a vote is cast in favour, and thus removes the 1%-2% of tests failing for this reason.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: go
 
 go:
-    - 1.2
+    - 1.4
+    - 1.5
+    - 1.6
     - tip
 
 install: make deps
--- a/raft.go
+++ b/raft.go
@@ -1481,6 +1481,7 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 	}
 
 	resp.Granted = true
+	r.setLastContact()
 	return
 }
 


abligh:
Rebased onto #93 to fix travis tests.

ongardie:
+1

armon:
@abligh Looks great!
@ongardie Thanks for reviewing.

PR #92, opened 2016-03-17
Close transport on shutdown

abligh:
(based on branch reset-heartbeat-on-requestvote which is in an earlier PR)

When a peer shuts down, close the transport. This needs to be done in the future (i.e. when all the routines have ended), else the remove peer does not propagate correctly - this was an issue with pull request #72.

Unfortunately we cannot simply add a `Close()` to the `Transport` interface as we would break back compatibility. Therefore add a type assertion to ensure that the interface supports `Close` before closing it.

Add a `Close()` method to `InmemTransport` so this actually gets tested. `InmemTransport` has nothing much to close (though it does call `DisconnectAll`), but other transports may do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: go
 
 go:
-    - 1.2
+    - 1.4
+    - 1.5
+    - 1.6
     - tip
 
 install: make deps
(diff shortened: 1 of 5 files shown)


abligh:
Rebased onto #93 in the hope that fixes travis tests.

armon:
Thanks!

Issue #94, opened 2016-03-17
vote requests denied incorrectly


ongardie:
The log comparison in the RequestVote request handler is incorrect. The current code, https://github.com/hashicorp/raft/blob/057b893/raft.go#L1463-L1475 , rejects the request if the voter's last log term is greater than the candidate's OR if the voter's last log index is greater than the candidate's.

The Raft paper describes the intended comparison (emphasis added):
> Raft determines which of two logs is more up-to-date
> by comparing the index and term of the last entries in the
> logs. If the logs have last entries with different terms, then
> the log with the later term is more up-to-date. **If the logs
> end with the same term**, then whichever log is longer is
> more up-to-date.

I believe this can lead to stuck clusters where no leader can be elected. For example:
```
S1: [1,1,1,1]
S2: [1,1,1,1]
S3: [1,1,2]
```
Suppose S2 fails, but S1 and S3 are still operational. S1 will not vote for S3, and S3 will not vote for S1.

abligh:
See pull request #95 - I am pretty sure that's what you meant, @ongardie

ongardie:
Yeah, #95 looks like the right fix.

armon:
Fixed by #95. Thanks!

PR #95, opened 2016-03-17
Compare logs correctly on a RequestVote

abligh:
Per issue #94:

> The log comparison in the RequestVote request handler is incorrect. The current code,
> rejects the request if the voter's last log term is greater than the candidate's OR
> if the voter's last log index is greater than the candidate's.

It should actually only compare the log index if the last terms are equal.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1468,7 +1468,7 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


ongardie:
+1. Thanks, saved me from having to open up my editor :)

armon:
@abligh Could you please add a test case for this? Thanks!

armon:
@abligh I've merged it, but a test case in a new PR would be lovely.

PR #97, opened 2016-03-18
Ensure we don't call the shutdown future more than once

abligh:
When shutting down a raft object, we should not run run anything significant in the shutdown future more than once. This was not previously a problem, but now we `Close()` the transport if it is closeable, and we shouldn't close it more than once.

### DIFF
--- a/future.go
+++ b/future.go
@@ -97,6 +97,9 @@ type shutdownFuture struct {
 }
 
 func (s *shutdownFuture) Error() error {
+	if s.raft == nil {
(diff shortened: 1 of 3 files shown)


slackpad:
@abligh this looks good. Would you mind adding a couple lines to `TestRaft_AfterShutdown` to get the future and call `Error()` on it?

abligh:
@slackpad thanks, and done.

slackpad:
Thanks!

PR #98, opened 2016-03-18
Add a datagram transport layer

abligh:
This **experimental** commit adds a datagram transport layer to raft. Examples of a datagram transport layer would include ZeroMQ used in datagram mode, or UDP.

This is submitted for comment. You would be brave to merge this into master as-is, and that is not expected.

Architecture
=========

The datagram support is broken into two layers:

DatagramTransport
--------------------------

`DatagramTransport` provides a simple transport where users can implement a datagram RPC interface that needs know nothing about Raft. The RPC interface is expected to transport datagrams of large sizes (no fragmentation is provided). A Codec interface is provided (the use of which is optional) which also serializes the `IOReader` in requests. Peers remain as strings, i.e. no address lookup is performed.

A typical user might be ZeroMQ.

An example usage is provided through `InmemDatagramTransport` which provides a trivial loopback transport.

PacketconnTransport
-----------------------------

`PacketconnTransport` provides a transport interface for things implementing the `net.Packetconn` interface. `PacketconnTransport` itself is small, with the work being done by the RPC provider, which provides full support for fragmentation. Peer addresses are resolved (by a supplied helper) into `net.Addr` addresses. These would normally be done with `Connect`/`Disconnect` which is basically registering / deregistering a peer address.

An example usage is provided through `UnixgramTransport` which provides a loopback transport through unix domain sockets.

These have not (yet) had a large amount of testing. Tests occasionally (~1% of the time) fail. However, I see test failures about 0.5% of the time with the existing `InmemTransport`. I believe this is down to the tests relying on timing, and that's sometimes put out by (e.g.) concurrent elections. The fact that these transports are slower (having to copy data) means they are probably more sensitive.

Other Notes
=========

Changes to existing files which are non-obvious:

* `transport.go`: Add a `LoopbackTransport` interface, which is in essence anything we can run `TestRaft_*` on

* `transport.go`: Add `WithPeers` interface to specify that `Connect` / `Disconnect` / `DisconnectAll` are present.

* Make `InmemTransport` optionally take a specified address. This is to provide compatibility with the other transports. It's necessary on other `Loopback` transports as sometimes we Close() a transport, but then wish to recreate a new peer. As the transport has been `Close()`d we can't reuse the same transport with the new peer, so we need to create a fresh transport with a specified address (being the address of the old disconnected peer). This is in any case a better test.

* `raft_test.go`: run tests which each transport we can test

* `raft_test.go`: don't hardcode time values; instead use a multiple of the heartbeat. This makes slowing things down easier.

* `raft_test.go`: do not reuse transports which have been shutdown (see `testRaft_SnapshotRestore` etc.)

* `transport_test.go`: Now runs tests from `net_transport_test.go` on the three loopback transports.


armon:
@abligh I think this is better maintained in a separate repository, much like `raft-mdb` and `raft-boltdb` so that the core logic can remain separate. If we split it that way, happy to review a PR to make the changes necessary here to support the datagram transport.

abligh:
@armon fair enough. I thought that might be the response! For completeness one could take only the datagram stuff (not the packetconn stuff) on the grounds that the datagram stuff is actually obviously useful (well, at least obvious to me), whereas it is not obvious why one would use UDP / Unixgram transport rather than TCP / Unix domain streams (I only wrote the to prove it could work).

But in any case, I'll work on splitting this out. In terms of what 'needs' to be taken, the new files `datagram_*.go`, `inmem_datagram_transport.go`, `packetconn_transport.go` don't need to be taken (obviously), but taking the other stuff in the order mentioned above, here's what I think you should take (if you agree I will separate them out):

> * `transport.go`: Add a `LoopbackTransport` interface, which is in essence anything we can run `TestRaft_*` on

I think you should take this so it's possible to have a single testing interface for any loopback transport. If you agree I will put it into a separate patch.

> * `transport.go`: Add `WithPeers` interface to specify that `Connect` / `Disconnect` / `DisconnectAll` are present.

I think you should take this so it's possible to have a single testing interface for any loopback transport. If you agree I will put it into a separate patch.

> * Make `InmemTransport` optionally take a specified address. This is to provide compatibility with the other transports. It's necessary on other `Loopback` transports as sometimes we Close() a transport, but then wish to recreate a new peer. As the transport has been `Close()`d we can't reuse the same transport with the new peer, so we need to create a fresh transport with a specified address (being the address of the old disconnected peer). This is in any case a better test.

I think you should take this so it's possible to have a single testing interface for any loopback transport. If you agree I will put it into a separate patch.

> `raft_test.go`: run tests which each transport we can test

I suppose you don't *need* this. However, having `raft_test` somehow easily reusable so it can be used to test a new transport someone writes is helpful. I can't think of an obvious way to do this if the transport is in another package beyond copying the file, applying all the changes necessary now it's outside the package `raft`, and continuously keeping it in sync. That's a significant pain for all authors of other protocols.

Perhaps I should just rewrite this into a separate `rafttest` package that can be imported (separately) into a small test for these transports, and that can live in a separate package within the same repository, and test any protocol that implements `Loopback`.

Your thoughts would be helpful.

> `raft_test.go`: don't hardcode time values; instead use a multiple of the heartbeat. This makes slowing things down easier.

The more I think about it the more I think this is probably not the right way to go anyway.

I've seen lots of instances where the current tests are fragile as far as timing is concerned. I think the correct way is (sadly) to remove all references to timeouts apart from the purpose of stopping tests after things really have gone wrong, and implement some sort of `RaftObserver` class, which gives (e.g.) the ability to wait for 3 heartbeat periods, wait for a leader, wait for an election, or whatever. This would have obvious production uses as well as test uses. We'd specify a large timeout (e.g. 5 seconds) on the basis that the timeout would only ever be hit if the tests failed. I'd implement a function to register/deregister an observer, and the main raft process would send events through a channel to the observer. Normally the channel would be (a) buffered and (b) non-blocking, so it couldn't get in the way of the algorithm.

This would obviously allow the protocol to run really slowly, which would make debugging test failures easier, as well as making them more repeatable. Would you take a patch to do that? I think it might clear up some of the persistent but rare test failures.

Your thoughts would be helpful.

> `raft_test.go`: do not reuse transports which have been shutdown (see `testRaft_SnapshotRestore` etc.)

I think you should take this. If you agree I will put it into a separate patch.

> `transport_test.go`: Now runs tests from `net_transport_test.go` on the three loopback transports.

This is basically `net_transport_test.go` adapted to run with any of the loopback transports. It would only be testing `inmem_transport.go` though. I found it easier to write a transport when it was subject to the same tests as the ones `raft_test.go` used. I have no strong feelings, but I suppose I'd suggest either you take this (so `inmem_transport.go` gets tested as a transport) or if I'm moving the `raft_test` stuff out to a separate package so it can be used with any transport, this should go with it.

slackpad:
Closing this out since pieces have been merged elsewhere. Thanks!

PR #102, opened 2016-03-21
Replace inflight tracker with commitment tracker (retargeted)

ongardie:
Retargeted #87 onto the issue-84-integration branch. The original commit (259e54d) is the same as before, only rebased. The subsequent commits address the comments from PR #87. This should probably be squashed into one commit when merging.

The only comment I didn't take from #87 was @armon's about switching to an RWMutex. (I don't plan to do that unless you insist.)


raft.go
```diff
@@ -750,7 +752,10 @@ func (r *Raft) runLeader() {
 
 	// Setup leader state
 	r.leaderState.commitCh = make(chan struct{}, 1)
-	r.leaderState.inflight = newInflight(r.leaderState.commitCh)
+	r.leaderState.commitment = newCommitment(r.leaderState.commitCh,
+		append([]string{r.localAddr}, r.peers...),
+		/* first index that may be committed in this term: */ r.getLastIndex()+1)
```
armon: Could you add the comment as trailing instead of leading?

armon:
Minor nit, but LGTM

ongardie:
Updated based on @armon's comment.

slackpad:
Looks good!

PR #103, opened 2016-03-21
Increase test timeout in Makefile

abligh:
Increase the test timeout in Makefile to avoid some travis test false positives.


abligh:
There is some irony to the fact this failed Travis!

armon:
Thanks, this should hopefully help robustness of tests!

PR #105, opened 2016-03-21
Dont use transports that have been shutdown

abligh:
Using a transport that has been shutdown is not safe if that shutdown does something.

This change permits `raft_test.go` to be be reused / cannibalised / extended to run with multiple types of loopback peer, and removes the misleading impression it is OK to reuse a transport that might be closed.

* Make `InmemTransport` optionally take a specified address. This is to provide compatibility with the other transports. It's necessary on other `Loopback` transports as sometimes we Close() a transport, but then wish to recreate a new peer. As the transport has been `Close()`d we can't reuse the same transport with the new peer, so we need to create a fresh transport with a specified address (being the address of the old disconnected peer). This is in any case a better test.

* `raft_test.go`: do not reuse transports which have been shutdown (see `testRaft_SnapshotRestore` etc.)


slackpad:
This looks good as well - thanks!

PR #106, opened 2016-03-21
Add transport_test.go

abligh:
Add `transport_test.go` - runs net_transport_test on loopback transports


slackpad:
Merged in #112.

PR #109, opened 2016-03-23
Change calculation of heartbeat timer reset

abligh:
According to the raft spec, we should time out between 1 and 2 intervals (here called `r.conf.HeartbeatTimeout`) after the last contact. Before this change, the heartbeat timeout was set to between 1 and 2 intervals after `Now()`. `Now()` may be substantially after the last contact as it itself
is driven by a heartbeat timeout.

There is a possibility that this value (1 and 2 intervals since last contact) is before `Now()`. So what we do is take a randomized interval from the later of the last interval and `Now() - r.conf.HeartbeatTimeout`.

I have measured the effect of this change using 1,000 iterations of TestRaft_TripleNode. In essence there is no measurable difference in the total time to complete an election. The attached graph shows the % of samples of time to complete the election process which were completed below the value on the y axis.

screen shot 2016-03-23 at 12 16 22

What it does improve however is the total time for a given election cycle (i.e. with the same term). The reason why this does not show in the end statistics is as follows:
- Elections where more than one node becomes a candidate during the election period are comparatively rare
- In those elections (only), we are concerned about the time taken for a follower to go into candidate state, and the fact this may sometimes be too long. Much of the time another follower will take the position.

However, this does appear to fix one cause of approximately 1 in 1,000 tests failing (once a number of other bugs in `raft_test.go` have been fixed). I think it's at worst harmless and at best a fix.

Note that a very substantial performance improvement to election convergence can be made (but has not been made) by using the same algorithm on the first setting of the heartbeat timer (when `runFollower` is started). This removes about 50ms off the time taken for an election. In essence we'd go to candidate state a random interval between 0 and 50ms after start, rather than 50ms and 100ms after start. This change should _not_ be made as such because a new follower joining a cluster would be quite likely to start an election. There is, however, probably an optimisation to be made where if a cluster is new (rather than an existing cluster being joined) we wait less time.

Note that though this is failing the integration tests, this is due to another bug in the integration tests (which I am fixing elsewhere).

Signed-off-by: Alex Bligh alex@alex.org.uk

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -612,6 +612,8 @@ func (r *Raft) runFollower() {
 	didWarn := false
 	r.logger.Printf("[INFO] raft: %v entering Follower state", r)
 	metrics.IncrCounter([]string{"raft", "state", "follower"}, 1)
+	// ignore when lastContact is as that may be substantially in the past, which
+	// means new joiners would start elections too early
 	heartbeatTimer := randomTimeout(r.conf.HeartbeatTimeout)
 	for {
 		select {
@@ -633,11 +635,17 @@ func (r *Raft) runFollower() {
 
 		case <-heartbeatTimer:
 			// Restart the heartbeat timer
-			heartbeatTimer = randomTimeout(r.conf.HeartbeatTimeout)
-
-			// Check if we have had a successful contact
+			// According to the raft spec, we should time out between 1 and 2 intervals
+			// (here called r.conf.HeartbeatTimeout) after the last contact. There
+			// is a possibility that this value is before Now(). So what we do is
+			// take a randomized interval from the later of the last interval
+			// and Now() - r.conf.HeartbeatTimeout.
+			now := time.Now()
 			lastContact := r.LastContact()
-			if time.Now().Sub(lastContact) < r.conf.HeartbeatTimeout {
+			startPoint := latest(lastContact, now.Add(-r.conf.HeartbeatTimeout))
+			duration := startPoint.Add(randomDuration(r.conf.HeartbeatTimeout)).Sub(now)
+			heartbeatTimer = time.After(duration)
+			if now.Sub(lastContact) < r.conf.HeartbeatTimeout {
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


abligh:
Actually, re the optimization I did not do, from the dissertation at 3.4:

> While waiting for votes, a candidate may receive an AppendEntries RPC from another server claiming to be leader. If the leader’s term (included in its RPC) is at least as large as the candidate’s current term, then the candidate recognizes the leader as legitimate and returns to follower state. If the term in the RPC is smaller than the candidate’s current term, then the candidate rejects the RPC and continues in candidate state.

Given we seem to check this OK, I'm wondering why we don't simply send a `RequestVote` out *immediately* we become follower. No one will vote for it if it is a new joiner or we already have a leader, and it will fall back to follower state.

slackpad:
Looking through this change and capturing my understanding of the current code here:

```
 C                              C                 <- case 1
 C                                            C   <- case 2
|xxxxxxxxxxxxxxx|               xxxxxxxxxxxxxxx|
^               ^               ^              ^
timer fires     timer fires     now+2*timeout  timer fires
@now            @now+timeout                   @now+3*timeout
```

The region with the `x`-es denotes any place in the next interval where last contact will be and we won't time out and change state. The `C` denotes the last contact.

In this example, we have a shortest timeout followed by the longest, which should let us get at the worst case spacing between contacts. For case 1 if the second contact was any further to the left we would have waited 2X the timeout to detect it. For case 2, we've got the longest span between contacts, 3X the timeout, and that's our upper bound.

I'm thinking we should always be able to add the randomly-staggered timeout to the last contact time and get a time in the future, because otherwise we will time out and change state anyway. Working through that a bit before I comment further :-)

Would be interesting to hear any thoughts from @ongardie as well.

s-christoff:
Hey, @abligh  - we appreciate this PR. We would love to hear more about this issue, if you are still interested in working on it just reply back and we will keep it open. If not, we will close it in 30 days.

Issue #107, opened 2016-03-22
False leader stepdown


abligh:
I am seeing around 1 in 1,000 test failures (log below in new exciting log format) where the leader is stepping down having not seen setLastContact() done for more than the leader lease time. I'm trying to work out why.

Heartbeats are sent every `commitTimeout` (set to 5ms on this testbed). Something is stopping these being seen (possibly linked to the double election).

I think this is a real bug (and not a test symptom). I'm making good progress elsewhere eliminating these.

Incidentally I am puzzled as to why raft enforces the leader lease timeout being smaller than or equal to the heartbeat timeout. This means (I think) the leader has to step down itself prior to another node declaring itself a candidate (as that will happen in the interval `heartbeatTimeout` ... 2 x `heartbeatTimeout`). What is the reason for this restriction? In this test they are both 50ms.

```
--- FAIL: TestRaft_TripleNode (0.52s)
        raft_test.go:94: cluster: 20:40:11.941538 [WARN] Fully Connecting
        raft_test.go:94: c076a5cd-a0d4-28dc-5e55-8f1e1fce1b3a: 20:40:11.942432 [INFO] raft: Node at c076a5cd-a0d4-28dc-5e55-8f1e1fce1b3a [Follower] entering Follower state
        raft_test.go:94: b41be96f-5378-c20f-843c-361c3046e379: 20:40:11.942521 [INFO] raft: Node at b41be96f-5378-c20f-843c-361c3046e379 [Follower] entering Follower state
        raft_test.go:94: 478d1966-28e7-a8bb-8973-fd6498d0fce2: 20:40:11.942574 [INFO] raft: Node at 478d1966-28e7-a8bb-8973-fd6498d0fce2 [Follower] entering Follower state
        raft_test.go:94: b41be96f-5378-c20f-843c-361c3046e379: 20:40:12.014779 [WARN] raft: Heartbeat timeout reached, starting election
        raft_test.go:94: b41be96f-5378-c20f-843c-361c3046e379: 20:40:12.014831 [INFO] raft: Node at b41be96f-5378-c20f-843c-361c3046e379 [Candidate] entering Candidate state
        raft_test.go:94: b41be96f-5378-c20f-843c-361c3046e379: 20:40:12.015274 [DEBUG] raft: Votes needed: 2
[...]

PR #111, opened 2016-03-24
Make raft test resilient

abligh:
Summary
=======

This is a largish commit that makes Raft testing far more resilient, and far less likely to be upset by timing changes.

Prior to this commit (as can be seen from Travis) I was seeing around 50% of `go test` fail in `TestRaft_`. After this commit, the number drops to 3% (for details of what these are, see below). The simplest test of cluster communication:

    go test -v -run 'TestRaft_TripleNode'

would fail about 3% of the time before these changes, and now fails about 0.5% of the time.

The remaining failures as far as I can tell are due to #107 and produce telltale lines like this in the log:

        raft_test.go:94: bd65f325-6d19-1aaa-38c2-571b01434800: 13:13:00.321830 [INFO] raft: aborting pipeline replication to peer 301a319c-9f36-f78a-9864-3cd2fd5c32c1
        raft_test.go:94: bd65f325-6d19-1aaa-38c2-571b01434800: 13:13:00.321841 [INFO] raft: aborting pipeline replication to peer 0d22fa12-ca51-1940-b099-e5d42259892b

Somehow the pipeline seems to jam. This would appear to be a real problem, rather than a testing problem.

The commit also substantially improves reporting of test failures, ensuring a microsecond clock is on every log line, and the log lines are not disordered. This makes debugging problems an enormous amount easier. Whilst timeouts have been increased (necessarily) in places, I have tried to reduce the number of 'dumb waits', so acting to reduce total test time when the tests succeed. Net, there is a small increase in total test time.

Detail
=====

* Add `Observation`s - emitted when something happens

* `Makefile`: Change test timeout in `Makefile` to 30s

* `raft_test.go`: Change default commit timeout to 5ms

* `raft_test.go`: Centralise all references to time in a single place.

* `raft_test.go`: Make logger work consistently and output time in microseconds (very useful for debugging). Convert all logging to use the cluster logger.

* `raft_test.go`: provide `c.Failf` and `c.FailNowf` functions that consistently produces the output, in log format, with timestamps. Convert use of `panic()` and `t.Fatalf()` to `c.FailNowf()`

* `raft_test.go`: rewrite `GetInState()` so it is now reliable, i.e. by waiting for the state to remain stable for a given period of time.

* `raft_test.go`: provide `WaitEventChan()` and `WaitEvent()` which wait for 'something to happen' or a timeout.

* `raft_test.go`: provide `WaitForReplication()` which waits until the FSM has a supplied number of logs on each node.

* `raft_test.go`: rewrite `Leaders()` and `Followers()` to be much more simple now `GetInState()` is reliable.

* `raft_test.go`: rewrite `EnsureLeader()` now `Leaders()` is reliable.


slackpad:
Hi @abligh could you rebase this given the earlier merges today and perhaps pop off the change from https://github.com/hashicorp/raft/pull/109, which I'd like to consider separately. Thanks!

abligh:
@slackpad rebased. I've included `Add transport_test.go` (i.e. #106) within this because it would require a lot of rewriting not to, and #106 is pretty self-contained (unlike this one) and uncontroversial I hope. #109 is excluded.

raft.go
```diff
@@ -147,6 +150,10 @@ type Raft struct {
 	// verifyCh is used to async send verify futures to the main thread
 	// to verify we are still the leader
 	verifyCh chan *verifyFuture
+
+	// list of observers and the mutex that protects them
+	obsmutex  sync.RWMutex
```
slackpad: Let's call this `observerLock` to fit with other mutex naming in here.

observer.go
```diff
@@ -0,0 +1,93 @@
...
+	Raft *Raft
+	Data interface{}
+}
+
+type LeaderObservation struct {
+	leader string
+}
+
+var (
+	ErrNoSuchObservationType = errors.New("No such observation type")
```
slackpad: This didn't end up being used and can be deleted.

abligh:
@slackpad I've fixed both your issues and pushed.

slackpad:
Hey @abligh thanks for all your hard work on this one! I took your PR and added some changes on top over in #112. Will merge and close this out. Appreciate your efforts here - the observer is a really nice feature for making the tests a lot better.

PR #113, opened 2016-04-02
Stop truncating non-conflicting entries in AppendEntries

ongardie:
## Copy of commit message

The AppendEntries request handler used to temporarily truncate entries
from the store that had indexes overlapping with the request, then it'd
go back and add all the entries in the request. This opened a window
of vulnerability during which if the server crashed, it could have
deleted some log entries that it shouldn't have.

For this to happen, the follower would have to receive an AppendEntries
request with entries that it already had. In principle, that could be
due to a duplicate request (depends on transport behavior) or to a
leader with a nextIndex set too low (probably isn't supposed to happen
with the back-up-one-entry algorithm).

The new code truncates only conflicting entries from the log store, then
appends only the new entries.

This also fixes a related issue where the last log index and term are
set to the last entry in the request, even if the last entry in the
request is not the last entry in the log (suppose the request contained
an earlier subsequence of the existing log).

## Discussion items

 - Is this introducing new problems of accessing the store non-atomically (#84)? Are RPC handlers allowed to run concurrently?

 - I think the best way to unit test this `appendEntries` function would be with a server that's in an interesting state (say, with a snapshot and a few log entries), and injecting hand-crafted AppendEntries requests into it to see how it behaves. I don't want a leader running or the servers to start elections. I didn't look through all of them, but I don't think the current tests are written in that style. How should I do that?
  - I probably further need to inject a thing into the middle of this function to test this particular change, that only the minimal set of entries got truncated. Or if I could get at stats on the log store, that would work too.

/cc @superfell @abligh

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1425,9 +1425,7 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 
 			// Update the lastLog
 			last := newEntries[n-1]
-			if last.Index > lastLogIdx {
-				r.setLastLog(last.Index, last.Term)
-			}
+			r.setLastLog(last.Index, last.Term)
 		}
 
 		metrics.MeasureSince([]string{"raft", "rpc", "appendEntries", "storeLogs"}, start)
--- a/raft.go
+++ b/raft.go
@@ -1388,29 +1388,48 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 	}
 
 	// Process any new entries
-	if n := len(a.Entries); n > 0 {
+	if len(a.Entries) > 0 {
 		start := time.Now()
-		first := a.Entries[0]
-		last := a.Entries[n-1]
 
-		// Delete any conflicting entries
+		// Delete any conflicting entries, skip any duplicates
 		lastLogIdx, _ := r.getLastLog()
-		if first.Index <= lastLogIdx {
-			r.logger.Printf("[WARN] raft: Clearing log suffix from %d to %d", first.Index, lastLogIdx)
-			if err := r.logs.DeleteRange(first.Index, lastLogIdx); err != nil {
-				r.logger.Printf("[ERR] raft: Failed to clear log suffix: %v", err)
+		var newEntries []*Log
+		for i, entry := range a.Entries {
+			if entry.Index > lastLogIdx {
+				newEntries = a.Entries[i:]
+				break
+			}
+			var storeEntry Log
+			if err := r.logs.GetLog(entry.Index, &storeEntry); err != nil {
+				r.logger.Printf("[WARN] raft: Failed to get log entry %d: %v",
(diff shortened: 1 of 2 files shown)


abligh:
The principle of this commit looks right to me.

On the matter of testing, I agree this would be easier without a cluster. The tests in `transport_test.go`, transmit handcrafted `AppendEntries` but as you say don't actually process them. The `raft_test.go` stuff all has a cluster. There's a gap there I guess. It would be nice if we had tests which simply applied `AppendEntries` RPC calls to a log *in vitro*.

For the way similar things are tested at the moment, I think the closest is probably `TestRaft_BehindFollower`. Perhaps you could adapt that? I realise this means having a cluster which you don't really want, but I suppose it brings the test closer to reality. When debugging, you now have microsecond times which makes it easier to see what's going on, and you can alter `func inmemConfig` to change the default timeouts there to be 100 times larger, and it will slow things down for you while you develop.

raft.go
```diff
@@ -1388,29 +1388,48 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
...
+		if n := len(newEntries); n > 0 {
+			// Append the new entries
+			if err := r.logs.StoreLogs(newEntries); err != nil {
+				r.logger.Printf("[ERR] raft: Failed to append to logs: %v", err)
+				return
+			}
+
+			// Update the lastLog
+			last := newEntries[n-1]
+			if last.Index > lastLogIdx {
```
slackpad: In the case where we truncated above, is `lastLogIdx` now stale? Should we re-fetch it before this check?

slackpad:
> Is this introducing new problems of accessing the store non-atomically (#84)? Are RPC handlers allowed to run concurrently?

I don't think so - RPCs are fetched from the `rpcCh` as part of the follower/candidate/leader loops so they should be handled one at a time, and not concurrently with other store operations.

> I think the best way to unit test this `appendEntries` function would be with a server that's in an interesting state (say, with a snapshot and a few log entries), and injecting hand-crafted AppendEntries requests into it to see how it behaves.

Following along @abligh's thoughts, you might be able to test this by running a setup similar to `TestRaft_BehindFollower` to get all the state setup correctly and then disconnect to isolate one of the followers and then call `appendEntries()` directly on it to run it through various cases. The FSM's logs are available to inspect afterwards to make sure the correct contents get ultimately committed.

raft.go
ongardie: Good catch. I think I can just drop the if condition and call r.setLastLog.

ongardie:
Added some more commits, including the list of tests that should be written but that are not easy to write today. Not sure what we want to do about this PR.

I don't like how this library tries to recover from storage failures. Trying to handle these failures without testing them seems like a really bad idea, and a Raft server without a working disk isn't very useful. Would anyone object to crashing on storage failures?

/cc @superfell

PR #117, opened 2016-04-11
Discussion: reworking membership changes

ongardie:
This is meant for discussion and is not meant to be merged. See the text in `membership.md`.


ongardie:
mentioning #84 ("Cleanup Meta Ticket")

membership.md
```diff
@@ -0,0 +1,83 @@
+Simon (@superfell) and I (@ongardie) talked through reworking this library's cluster membership changes last Friday. We don't see a way to split this into independent patches, so we're taking the next best approach: submitting the plan here for review, then working on an enormous PR. Your feedback would be appreciated. (@superfell is out this week, however, so don't expect him to respond quickly.)
+
+These are the main goals:
+ - Bringing things in line with the description in my PhD dissertation;
+ - Catching up new servers prior to granting them a vote, as well as allowing permanent non-voting members; and
+ - Eliminating the `peers.json` file, to avoid issues of consistency between that and the log/snapshot.
+
+## Data-centric view
+
+We propose to re-define a *configuration* as a set of servers, where each server includes an address (as it does today) and a mode that is either:
```
slackpad: This seems like a good opportunity to give each server something like a GUID that's independent of its address, so that we can eventually support address changes without changing a server's identity.

membership.md
```diff
@@ -0,0 +1,83 @@
...
+ 2. its *latest configuration*: the latest configuration in the log/snapshot (may be committed or uncommitted), along with its index.
+
+When there's no membership change happening, these two will be the same. The latest configuration is almost always the one used, except:
+ - When followers truncate the suffix of their logs, they may need to fall back to the committed configuration.
+ - When snapshotting, the committed configuration is written, to correspond with the committed log prefix that is being snapshotted.
+
+
+## Application API
+
+We propose the following operations for clients to manipulate the cluster configuration:
```
slackpad: We could leave `AddPeer()` for now and have it call `AddVoter()` and print a deprecation warning to the logs.

membership.md
```diff
@@ -0,0 +1,83 @@
...
+the possible intent of the user. For example, if I want to make sure a server doesn't have a vote, but the server isn't part of the configuration at all, it probably shouldn't be added as a nonvoting server.
+
+Each of these application-level operations will be interpreted by the leader and, if it has an effect, will cause the leader to write a new configuration entry to its log. Which particular application-level operation caused the log entry to be written need not be part of the log entry.
+
+## Code implications
+
+This is a non-exhaustive list, but we came up with a few things:
+- Remove the PeerStore: the `peers.json` file introduces the possibility of getting out of sync with the log and snapshot, and it's hard to maintain this atomically as the log changes. It's not clear whether it's meant to track the committed or latest configuration, either.
+- Servers will have to search their snapshot and log to find the committed configuration and the latest configuration on startup.
+- Bootstrap will no longer use `peers.json` but should initialize the log or snapshot with an application-provided configuration entry.
```
slackpad: I think we will need to think through some recovery scenarios here since it's fairly common to have to edit the `peers.json` to get things going in the face of a permanently-failed server coupled with a quorum loss. I could see adding an API to set a fixed configuration, or possibly even running a command on each server to force all of the configuration to nonvoter, and then running `AddVoter()` or similar on one of the servers to get it to take leadership, and then promoting others one-by-one to recover.

membership.md
```diff
@@ -0,0 +1,83 @@
...
+- Snapshots should store the index of their configuration along with the configuration itself. In my experience with LogCabin, the original log index of the configuration is very useful to include in debug log messages.
+- As noted in hashicorp/raft#84, configuration change requests should come in via a separate channel, and one may not proceed until the last has been committed.
+- As to deciding when a log is sufficiently caught up, implementing a sophisticated algorithm *is* something that can be done in a separate PR. An easy and decent placeholder is: once the staging server has reached 95% of the leader's commit index, promote it.
```
slackpad: This seems like a good thing to make pluggable with a simple interface and default implementation. The application can provide a fancier implementation without having to implement that inside Raft.

membership.md
```diff
@@ -0,0 +1,83 @@
...
+
+While these operations aren't quite symmetric, we think they're a good set to capture
+the possible intent of the user. For example, if I want to make sure a server doesn't have a vote, but the server isn't part of the configuration at all, it probably shouldn't be added as a nonvoting server.
+
+Each of these application-level operations will be interpreted by the leader and, if it has an effect, will cause the leader to write a new configuration entry to its log. Which particular application-level operation caused the log entry to be written need not be part of the log entry.
+
+## Code implications
+
+This is a non-exhaustive list, but we came up with a few things:
+- Remove the PeerStore: the `peers.json` file introduces the possibility of getting out of sync with the log and snapshot, and it's hard to maintain this atomically as the log changes. It's not clear whether it's meant to track the committed or latest configuration, either.
```
slackpad: I like removing the peer store a lot. It seems like we could retain the peer store for one version as a way to transition to the new scheme. If we start up and don't find the new config record in the snapshot or log, we will try the peer store and use that as the committed configuration. Once we've committed a new-style configuration from the log we can zero out any contents of the peer store.

membership.md
```diff
@@ -0,0 +1,83 @@
...
+These are the main goals:
+ - Bringing things in line with the description in my PhD dissertation;
+ - Catching up new servers prior to granting them a vote, as well as allowing permanent non-voting members; and
+ - Eliminating the `peers.json` file, to avoid issues of consistency between that and the log/snapshot.
+
+## Data-centric view
+
+We propose to re-define a *configuration* as a set of servers, where each server includes an address (as it does today) and a mode that is either:
+ - *Voter*: a server whose vote is counted in elections and whose match index is used in advancing the leader's commit index.
+ - *Nonvoter*: a server that receives log entries but is not considered for elections or commitment purposes.
```
slackpad: Do we prioritize replication to voters first?

membership.md
```diff
@@ -0,0 +1,83 @@
...
+- Servers will have to search their snapshot and log to find the committed configuration and the latest configuration on startup.
+- Bootstrap will no longer use `peers.json` but should initialize the log or snapshot with an application-provided configuration entry.
+- Snapshots should store the index of their configuration along with the configuration itself. In my experience with LogCabin, the original log index of the configuration is very useful to include in debug log messages.
+- As noted in hashicorp/raft#84, configuration change requests should come in via a separate channel, and one may not proceed until the last has been committed.
+- As to deciding when a log is sufficiently caught up, implementing a sophisticated algorithm *is* something that can be done in a separate PR. An easy and decent placeholder is: once the staging server has reached 95% of the leader's commit index, promote it.
+
+## Feedback
+
+Again, we're looking for feedback here before we start working on this. Here are some questions to think about:
```
slackpad: > + - Does this seem like where we want things to go?

After a first look, I think so - there's a lot of utility with the extra states, and I think they provide nice ways to make operations easier. I'd much rather spin up my three new servers as non-voters and watch them to be healthy for a while before I put them into voting then to have to yolo like we do today. There are good applications for read scaling, too, with systems like Consul which allows stale reads.

> + - Is there a good way to break this up?

It seems like much of this could be done outside of `raft.go`. Ideally, we could have the peer state machine and "eligible to vote" algorithm in completely separate classes and testable outside of an integrated configuration. Similar for the log / snapshot configuration searcher (there will be quite a few cases to test there). 

> + - What do we need to worry about in terms of backwards compatibility?

Put some notes up above. I think we can fairly easily support `AddPeer()` during this transition, and it might be useful to hang on to the `PeerStore` to make migration easier. If something is going to be a huge effort to support we could probably break an interface, but it would be good to avoid it if we can, especially during this transition period.

We might need to be creative to support a zero-downtime rolling upgrade if we have a new type of config message. This would be ideal if we can figure out a way to do it.

> + - What implication will this have on current tests?

If we mirrored the committed configuration into the peer store (if one is supplied) we could probably get the existing `raft_test.go` to pass with the new code. This might give us a good baseline before we change all the tests to use whatever the new "who are your peers" interface is, then we could rip out the mirroring.

slackpad:
@ongardie added some initial thoughts as inline comments. I've got some cycles this week and can work on this with you guys as well if we can figure out some ways to break things up.

membership.md
ongardie: Want to spec out how that would work? Specifically:
 - Who would generate the GUID and when?
 - Where would it be stored?
 - How would it be reflected in the old AddPeer/RemovePeer or upcoming AddVoter/AddNonvoter/DemoteVoter/RemovePeer API?

membership.md
ongardie: We should definitely think through that. I'm not familiar with how peers.json is used today in recovery scenarios. Can you enumerate the use cases (user stories maybe?)?

membership.md
ongardie: A pluggable implementation makes sense if different applications would want different things, but you can write up a pretty damn good policy in the library that I think everyone would want to have (once adding the server will cause less than an election timeout of delay, and you've kept going for a bit after that and are still in good shape, then go ahead).

membership.md
ongardie: Something like that should work. I'm a fan of reading in the peers.json file, appending to the log if needed, then deleting the peers.json file.

What takes precedence today: peers.json or snapshot/log?

membership.md
ongardie: I wasn't thinking we would. It'd take some bookkeeping across the replication routines to do some sort of application-level rate limiting, unless the transports provide QoS.

membership.md
ongardie: > We might need to be creative to support a zero-downtime rolling upgrade if we have a new type of config message. This would be ideal if we can figure out a way to do it.

Does the leader have any way to tell what version other servers are running (or, equivalently, what their capabilities are)?

ongardie:
@slackpad thanks for the comments. I replied to a few above.

> I've got some cycles this week and can work on this with you guys as well if we can figure out some ways to break things up.

As I mentioned, @superfell is out this week. I was going to wait for him to return (he hasn't read this doc, and he might be sitting on some useful changes). We may have to do pair programming or something extreme like that, because I really don't see how to split up the work (pun intended).

If you've got space cycles this week, I think we'd be better off applying them to more parallelizable tasks. For example, how about attacking some more of #84 (AFAIK, multi-row fetching or follower replication are open)?

membership.md
slackpad: Fair enough - I was thinking about a possible use case where an operator might want to make the promote decision but that's already kind of there with your API so I'm good without this being pluggable.

membership.md
slackpad: This has a little detail on how it's used: https://www.consul.io/docs/guides/outage.html. I think the typical use cases are these:

1. Loss of quorum in a multiple server cluster where you can't bring up new servers with the same IPs as the lost servers. You can stop the remaining servers, edit the peers.json to remove the permanently failed servers, and then restart them to try to regain quorum, and then add new servers into the mix.

2. Sync mis-match between raft peer state and cluster state. If there's a leadership transition at the time the event comes in to remove a server from the Consul cluster, there have been cases where there's a failed server in the Raft peers that's no longer in the Consul cluster, so there's currently no mechanism to remove the server from the Raft peers. This can likely be handled with some better operator commands, or a more robust syncing between these two sources of state, but the current fix for this is to stop all the servers and remove the orphaned server from the peers.json file.

The peers.json file is kind of a good escape hatch for situations where the cluster gets into a state where it can't make changes on its own, so a human operator needs to figure out what to do. Unfortunately, the way it is currently set up, it doesn't always take precedence (things in the Raft log can override what the operator wants), so I think it would be much better if we could have the cluster get into a safe state (or have a way for an operator to put the cluster into a safe state), and let the operator promote a specific server to be leader to get quorum of 1, and then promote servers back into voting status once things are stable.

It seems super clean if the cluster is in an outage state to pick the node you want to be a leader, tell it to force all other nodes into nonvoter status, have it elect itself, and then have it convert them all to voting (via staging and the stability check).

membership.md
slackpad: I'll spec this out - I think I've got a plan that'll work for this.

membership.md
```diff
@@ -0,0 +1,83 @@
...
+ - When followers truncate the suffix of their logs, they may need to fall back to the committed configuration.
+ - When snapshotting, the committed configuration is written, to correspond with the committed log prefix that is being snapshotted.
+
+
+## Application API
+
+We propose the following operations for clients to manipulate the cluster configuration:
+ - AddVoter: server becomes staging unless voter,
+ - AddNonvoter: server becomes nonvoter unless staging or voter,
+ - DemoteVoter: server becomes nonvoter unless absent,
```
slackpad: To have more symmetry with `AddVoter` how about just calling `AddNonvoter` in order to demote? Either that or we should have a `PromoteVoter` API vs. adding it again.

fd:
Is there a time line for these enhancements?

slackpad:
@fd work is underway now, so hopefully very soon!

ongardie-sfdc:
Added a really, really big diff (687db02) with the bulk of the changes. @superfell gets credit for some of it (especially fixing tests), but I squashed his stuff into the one commit.

@slackpad, @sean-: please review. We should aim to come up with a list of outstanding problems. Some of them are going to be easy, others less so.

slackpad:
@ongardie thanks for this! I'm returning from some travel on Monday so I should be able to make a deep dive here on Tuesday.

raft.go
```diff
@@ -198,13 +267,8 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
...
 	}
 
-	// Construct the list of peers that excludes us
 	localAddr := trans.LocalAddr()
-	peers, err := peerStore.Peers()
-	if err != nil {
-		return nil, fmt.Errorf("failed to get list of peers: %v", err)
-	}
-	peers = ExcludePeer(peers, localAddr)
+	localGUID := localAddr // TODO: where should we get this from?
```
slackpad: My initial thought was that this would be provided as a new field in `conf`, and if that's empty then we'd use `localAddr` as a default and emit a deprecation warning that we'll stop defaulting it at some point in the future.

raft.go
```diff
@@ -359,54 +441,91 @@ func (r *Raft) VerifyLeader() Future {
...
-	peerFuture.init()
+// AddVoter will add the given server to the cluster as a staging server. If the
+// server is already in the cluster as a voter, this does nothing. This must be
+// run on the leader or it will fail. The leader will promote the staging server
+// to a voter once that server is ready. If nonzero, prevIndex is the index of
+// the only configuration upon which this change may be applied; if another
+// configuration entry has been added in the meantime, this request will fail.
+// If nonzero, timeout is how long this server should wait before the
+// configuration change log entry is appended.
+func (r *Raft) AddVoter(guid string, address string, prevIndex uint64, timeout time.Duration) IndexFuture {
```
slackpad: This would be a good time to create a new string type for the GUID so we can help enforce some compiler checking.

raft.go
```diff
@@ -488,7 +607,7 @@ func (r *Raft) Stats() map[string]string {
 		"fsm_pending":         toString(uint64(len(r.fsmCommitCh))),
 		"last_snapshot_index": toString(lastSnapIndex),
 		"last_snapshot_term":  toString(lastSnapTerm),
-		"num_peers":           toString(uint64(len(r.peers))),
+		"num_servers":         toString(uint64(len(r.configurations.latest.Servers))),
```
sean-: Elsewhere we've taken to including the list of servers along with the number of peers.  e.g.  https://github.com/hashicorp/nomad/blob/c6b7bb66f309b7fc3ef325b158115aa046dcbf30/nomad/server.go#L905-L906

ongardie-sfdc: A list of servers doesn't have all the information in this new world. What is this function used for?

raft.go
ongardie-sfdc: Like `type GUID string`? or `type ServerGUID string` ?

And do you want the same for an address?

raft.go
ongardie-sfdc: Ok, that's what I was leaning towards.

raft.go
slackpad: Exactly - `type ServerGUID string` would work, or even `type ServerID string`. It's a good idea to do the same for `Address` so we can be super clear which is which, especially as we make the transition.

raft.go
slackpad: This ends up getting displayed to operators when they run `consul info` on a server, so the results here can be pretty free-form. This is used for debugging and bug reporting.

configuration.go
```diff
@@ -0,0 +1,156 @@
...
+
+import "fmt"
+
+// ServerSuffrage determines whether a Server in a Configuration gets a vote.
+type ServerSuffrage int
+
+const (
+	// Voter is a server whose vote is counted in elections and whose match index
+	// is used in advancing the leader's commit index.
+	Voter ServerSuffrage = 1
```
slackpad: Would be slightly more idiomatic Go to use `iota` on this first one and leave the others blank to auto-number them.

commitment.go
```diff
@@ -26,14 +26,15 @@ type commitment struct {
...
 // 'startIndex' is the first index created in this term (see
 // its description above).
-func newCommitment(commitCh chan struct{}, voters []string, startIndex uint64) *commitment {
-	matchIndexes := make(map[string]uint64, len(voters))
-	for _, voter := range voters {
-		matchIndexes[voter] = 0
+func newCommitment(commitCh chan struct{}, configuration Configuration, startIndex uint64) *commitment {
+	matchIndexes := make(map[string]uint64)
+	for _, server := range configuration.Servers {
+		if server.Suffrage == Voter {
```
slackpad: I wonder if it's worth adding a `CanVote()` member to `ServerSuffrage` in case we add other states later. Definitely not needed now, and probably a YAGNI, thinking out loud :-)

configuration_test.go
```diff
@@ -0,0 +1,71 @@
+package raft
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+)
+
```
slackpad: TODO - add a test for `hasVote()`.

configuration_test.go
```diff
@@ -0,0 +1,71 @@
...
+	configuration.Servers[1].GUID = "guid0"
+	if checkConfiguration(configuration) == nil {
+		t.Fatalf("duplicate GUID should be error")
+	}
+	configuration.Servers[1].GUID = "guid1"
+
+	configuration.Servers[1].Address = "addr0"
+	if checkConfiguration(configuration) == nil {
+		t.Fatalf("duplicate address should be error")
+	}
```
slackpad: For these kinds of tests it's nice to use `Strings.Contains()` on `err.Error()` to make sure it's failing for the correct reason.

raft.go
```diff
@@ -250,6 +314,24 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
...
+	snapshotIndex, _ := r.getLastSnapshot()
+	for index := snapshotIndex + 1; index <= lastLog.Index; index++ {
+		var entry Log
+		if err := r.logs.GetLog(index, &entry); err != nil {
+			r.logger.Printf("[ERR] raft: Failed to get log at %d: %v", index, err)
+			panic(err)
+		}
+		// TODO: support LogAddPeer, LogRemovePeer too
+		if entry.Type == LogConfiguration {
+			r.configurations.committed = r.configurations.latest
```
slackpad: Since we are reading out of the log, shouldn't these always flow to both slots in the `configurations` struct? This looks like committed will be one step behind.

raft.go
slackpad: A `%+v` on the configuration might be the best thing here.

raft.go
```diff
@@ -552,10 +671,9 @@ func (r *Raft) runFSM() {
 				continue
 			}
 
-			// Get our peers
-			peers, err := r.peerStore.Peers()
-			if err != nil {
-				req.respond(err)
+			if lastIndex < r.configurations.committedIndex {
+				req.respond(fmt.Errorf("cannot take snapshot now, wait until the configuration entry at %v has been applied (have applied %v)",
```
slackpad: This'll be a little weird in the logs but should be exceedingly rare. It will just bubble up and result in a log message and shouldn't break anything.

raft.go
```diff
@@ -621,7 +740,7 @@ func (r *Raft) run() {
 
 // runFollower runs the FSM for a follower.
 func (r *Raft) runFollower() {
-	didWarn := false
+	didWarn := true
```
slackpad: This looks like a mistake and disables the warnings completely.

raft.go
```diff
@@ -656,14 +774,19 @@ func (r *Raft) runFollower() {
...
+					r.logger.Printf("[WARN] raft: no known peers, aborting election")
+					didWarn = true
+				}
+			} else if r.configurations.latestIndex == r.configurations.committedIndex &&
+				!hasVote(r.configurations.latest, r.localGUID) {
+				if !didWarn {
+					r.logger.Printf("[WARN] raft: not part of stable configuration, aborting election")
 					didWarn = true
 				}
 			} else {
```
slackpad: It feels like there's a hole here where a non-voter could try to elect itself if there's a config change in process (the indexes don't match). Do we need a blanket `hasVote` here in the final `else` clause?

raft.go
```diff
@@ -901,8 +1047,25 @@ func (r *Raft) leaderLoop() {
...
 		case <-r.leaderState.commitCh:
 			// Process the newly committed entries
+			oldCommitIndex := r.getCommitIndex()
 			commitIndex := r.leaderState.commitment.getCommitIndex()
 			r.setCommitIndex(commitIndex)
+			if r.configurations.latestIndex > oldCommitIndex &&
+				r.configurations.latestIndex <= commitIndex {
+				r.configurations.committed = r.configurations.latest
+				r.configurations.committedIndex = r.configurations.latestIndex
+				haveVote := false
```
slackpad: With the helper you can get rid of the loop here and just do `stepDown = !hasVote(r.configurations.committed, r.localGUID)`

raft.go
```diff
@@ -937,8 +1110,8 @@ func (r *Raft) leaderLoop() {
 				v.respond(nil)
 			}
 
-		case p := <-r.peerCh:
-			p.respond(ErrLeader)
+		case future := <-r.configurationChangeChIfStable():
```
slackpad: This'll need another event to kick the loop if the channel is `nil`. Any chance this can get stuck?

raft.go
```diff
@@ -1176,25 +1413,24 @@ func (r *Raft) processLogs(index uint64, future *logFuture) {
...
-			r.processLog(&future.log, future, false)
+			r.processLog(&future.log, future)
 
 		} else {
 			l := new(Log)
 			if err := r.logs.GetLog(idx, l); err != nil {
 				r.logger.Printf("[ERR] raft: Failed to get log at %d: %v", idx, err)
 				panic(err)
 			}
-			r.processLog(l, nil, false)
```
slackpad: Yay - RIP `precommit`!

raft.go
```diff
@@ -1650,51 +1869,43 @@ func (r *Raft) electSelf() <-chan *voteResult {
...
 				resp.Granted = false
 			}
-
-			// If we are not a peer, we could have been removed but failed
-			// to receive the log message. OR it could mean an improperly configured
-			// cluster. Either way, we should warn
-			if err == nil {
-				peerSet := decodePeers(resp.Peers, r.trans)
-				if !PeerContained(peerSet, r.localAddr) {
-					r.logger.Printf("[WARN] raft: Remote peer %v does not have local node %v as a peer",
```
slackpad: This was a pretty useful check to look for bad configuration state. Any reason you didn't want a similar check with the new scheme?

configuration.go
ongardie-sfdc: The problem with that is, since these are written into the log, we should be careful to never renumber or reuse numbers. It's the same reason why protocol buffers makes you assign numbers manually. So there's my rationale. I'll change to iota if you still think that's better.

raft.go
ongardie-sfdc: Don't think so. Not everything we find in the log will be committed. We do know a new configuration entry can't be created until the prior one was committed => the penultimate configuration in the log is committed. The last one may not be.

Note that latest is used almost everywhere. If latest is in fact committed, this server will never be asked to fall back to the previous one, and it should all work out just fine.

raft.go
ongardie-sfdc: Thanks. I had removed that, then tried to add it back but got it wrong.

raft.go
ongardie-sfdc: For a brief period, you might need that. See the top half of page 40 in my dissertation.

raft.go
ongardie-sfdc: Good point. hasVote() came after this code was written.

raft.go
ongardie-sfdc: To answer that, here are the variables in the condition for configurationChangeChIfStable(). The question becomes: can these change outside of leaderLoop() while this server is still leader?
 - r.configurations.latestIndex
 - r.configurations.committedIndex
 - r.getCommitIndex()
 - r.leaderState.commitment.startIndex (const)

I don't think so, right?

raft.go
ongardie-sfdc: I don't like the idea of warning under perfectly normal situations. Can you be specific about what scenarios you're trying to catch with a message here? Maybe there's another way to do that.

configuration.go
slackpad: For sure - we use `iota` in log.go for the actual log message types so we are already using it for critical numbering / add only at the end kinds of things.

ongardie-sfdc: ok, will switch to iota

ongardie-sfdc: with a // warning

raft.go
slackpad: Oh interesting - thank you for the reference for this one - that's pretty subtle.

raft.go
slackpad: True - if there are normal cases that can lead to this then a warning isn't appropriate. I'm mostly concerned about knowing if, in the steady state outside of configuration changes, there's ever a discrepancy between the configuration among all members in the cluster. Maybe this would be better done with a running checksum over the applied logs or something like that. This is probably something to think about separate from this PR.

raft.go
slackpad: Looking through the code I agree this looks fine. We should be getting the commit channel advancing us as the config change is accepted by the peers and that will cause us to loop again. These items won't change outside leader loop.

ongardie-sfdc: Ok, just adding a comment here.

ongardie-sfdc:
I believe I've addressed all the actionable code comments above in this latest push (commits 15e3819 to 33b1b0e). I tagged the specific comments that I believe I addressed with the 🎉 reaction (try to avoid using that to indicate celebration).

slackpad:
Went ahead and merged this so we can submit smaller-scoped changes to close this out. Make a list of outstanding items on #84.

ongardie-sfdc:
@slackpad, I wasn't expecting you to merge that! I think there's still some important stuff left to do. For example, we should audit whether this is going to work for rolling upgrades.

ongardie-sfdc:
I think not. For example, new code will create LogConfiguration entries. What will old code do with them?

ongardie-sfdc:
Oh, I see the list on #84 now, fine.

slackpad:
@ongardie I just merged it to the integration branch, not master!

PR #118, opened 2016-04-13
Adds a check of the `stopCh` to break out of the `replicateTo()` loop if needed.

slackpad:
This addresses the "Follower Replication" item on #84. There wasn't really a way to exercise this clause without artificial hooks, so I added a new test case that just pushes things into the new code as a regression.

I also made a sweep through here to look for other loops where we need to check `stopCh` and everything else looks good. The pipeline case already has handling of it because it doesn't internally loop like `replicateTo()` does.


armon:
LGTM!

PR #120, opened 2016-04-20
Adds test to check that config changes can't happen w/o quorum.

slackpad:
This is currently failing so I'll target this at the branch so we can verify it passes once the peer changes are complete.

Mentioning #84.


slackpad:
Merged this into #127 instead, since it has the config management changes that make it pass.

PR #121, opened 2016-05-05
Log the current leader when entering the follower state.

sean-:
When entering the candidate state, log which Leader went AWOL.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -620,7 +620,7 @@ func (r *Raft) run() {
 // runFollower runs the FSM for a follower.
 func (r *Raft) runFollower() {
 	didWarn := false
-	r.logger.Printf(`[INFO] raft: %v entering Follower state (Leader: %q)`, r, r.Leader())
+	r.logger.Printf("[INFO] raft: %v entering Follower state (Leader: %q)", r, r.Leader())
 	metrics.IncrCounter([]string{"raft", "state", "follower"}, 1)
 	heartbeatTimer := randomTimeout(r.conf.HeartbeatTimeout)
 	for {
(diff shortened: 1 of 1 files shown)


raft.go
```diff
@@ -620,7 +620,7 @@ func (r *Raft) run() {
 // runFollower runs the FSM for a follower.
 func (r *Raft) runFollower() {
 	didWarn := false
-	r.logger.Printf("[INFO] raft: %v entering Follower state", r)
+	r.logger.Printf(`[INFO] raft: %v entering Follower state (Leader: "%s")`, r, r.Leader())
```
mitchellh: Fun fact (you don't need to change this at all, but in case you didn't know), you can use the `%q` syntax to get a quoted string. This will also call `String()` on it if it is a `Stringer` impl.

sean-: ! I haven't committed all of the fmt verbs to memory yet.  Super useful, thanks.

slackpad:
LGTM

raft.go
slackpad: TIL - I was just Googling `%q` - that's a great tip :-)

raft.go
```diff
@@ -620,7 +620,7 @@ func (r *Raft) run() {
 // runFollower runs the FSM for a follower.
 func (r *Raft) runFollower() {
 	didWarn := false
-	r.logger.Printf("[INFO] raft: %v entering Follower state", r)
+	r.logger.Printf(`[INFO] raft: %v entering Follower state (Leader: %q)`, r, r.Leader())
```
slackpad: With `%q` you can probably delete the backquotes, too.

PR #124, opened 2016-06-14
buffer applyCh with up to conf.MaxAppendEntries

stapelberg:
This change improves throughput in busy Raft clusters.
By buffering messages, individual RPCs contain more Raft messages.
In my tests, this improves throughput from about 4.5 kqps to about 5.5 kqps.

As-is:
n1-standard-8-c8deaa9d333f69fb56c8935036e7ca5c-no-buffer

With my change:
n1-standard-8-f2fed4ebe05df23eb322c805b503a144

(Both tests were performed with 3 n1-standard-8 nodes on Google Compute Engine in the europe-west1-d region.)

### DIFF
--- a/api.go
+++ b/api.go
@@ -478,7 +478,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 	// Create Raft struct.
 	r := &Raft{
 		protocolVersion:       protocolVersion,
-		applyCh:               make(chan *logFuture),
+		applyCh:               make(chan *logFuture, conf.MaxAppendEntries),
 		conf:                  *conf,
 		fsm:                   fsm,
 		fsmMutateCh:           make(chan interface{}, 128),


stapelberg:
The travis build failed because it ran into a timeout. I think restarting it will help.

ongardie:
With the disclaimer that I might be completely wrong about Go's scheduler, this PR makes sense to me. That loop within leaderLoop tries to drain up to conf.MaxAppendEntries on the channel because it's more efficient to send a large batch of AppendEntries. With a buffered channel, it's easy to see that the leaderLoop will drain many entries from the channel. With an unbuffered channel, there could be multiple goroutines blocked to send on the channel, but I don't think go makes any guarantee that a non-blocking receive in a loop will consume all of them. I think this depends on getting lucky with the scheduler.

There's some related discussion that I found helpful in this thread: https://groups.google.com/d/msg/golang-nuts/UnzE5vgyzqw/eZdhrhoWJQUJ

ongardie-sfdc:
And I'm second-guessing myself on my previous comment.

cc @superfell

s-christoff:
Hey @stapelberg  - thank you for this. Is there any way you could rebase with master so CircleCI can take a stab at this?

superfell:
One side effect of this change would be related to how timeout works in ApplyLog. The timeout is based on the item being put on the channel, so timeouts won't start happening now until the channel buffer is full.

stapelberg:
Rebased!

i0rek:
@superfell thanks for helping out! I don't understand which timeouts you are referring to. Could you try to explain that again? Thanks a lot!

i0rek:
@superfell All good now, I figured out which timeout you mean:

https://github.com/hashicorp/raft/blob/db5ceea63eeda4e4022acfc1a333cc835ab957ac/api.go#L656-L684

But I still don't understand how the semantics are changing. Because right now we are starting the timeout before we write to the channel which means we are waiting for the leader to read it and that time is counted. 
Which is the same after this change were we can write to the channel, but are still waiting similarly for the leader to read it.

Thanks!

superfell:
Its timing how long it takes to write to the channel. If you change the channel to a buffered channel, the write to the channel can succeed (if there's space), but it hasn't been processed by anything, and when it'll get processed is still some indeterminate time in the future. With the unbuffered channel as it currently is, Apply can't write it to the channel until the leader loop is ready to read it from the channel. So what the timeout is timing against is different, and gets more different the larger the channel buffer is.

superfell:
ApplyLog doesn't wait for the leaderLoop to read the item from the channel, it only waits until it put the future on the channel. It only waits until the leaderLoop reads from the channel currently because its an unbuffered channel.

stapelberg:
> I know this is an _ancient_ PR so I'm not necessarily expecting you to pick this up again (although you are welcome)

I don’t have the time or motivation right now to pick this up again, so if anyone else wants to take it over, feel free to :)

alecjen:
@banks I found that the proposed solution will not ever buffer applyCh for Apply operations running on a single routine. Because Apply is waiting for an ack from the done channel, it is effectively blocked until the leaderLoop completes the current dispatchLogs operation.

You may have already known this, but I think we can consider at least exposing the option to buffer the applyCh, keeping in mind that we are re-working the timeout assumptions in this case.

ncabatoff:
Closing as this was implemented in https://github.com/hashicorp/raft/pull/445.

PR #125, opened 2016-06-23
small metric fix

semihalev:


slackpad:
Hi @semihalev thanks for the PR! We are reworking peer changes in the issue-84-integration branch. Would you mind retargeting this change against that? The information about peers will be in a slightly different form.

PR #127, opened 2016-06-28
Fixes tests and plumbs more ServerAddress vs. string parameters.

slackpad:
This cleans up a couple of todo items and plumbs `ServerAddress` into more places for extra clarity. I decided that since the fix is trivial (just a cast to support old code) that it was worth breaking the interface to make the code super clear.

/cc @ongardie 

Mentioning #84.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -353,10 +353,9 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 // Leader is used to return the current leader of the cluster.
 // It may return empty string if there is no current leader
 // or the leader is unknown.
-func (r *Raft) Leader() string {
-	// TODO: change return type to ServerAddress?
+func (r *Raft) Leader() ServerAddress {
 	r.leaderLock.RLock()
-	leader := string(r.leader)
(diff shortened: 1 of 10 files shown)


raft.go
```diff
@@ -452,15 +447,15 @@ func (r *Raft) VerifyLeader() Future {
...
+func (r *Raft) AddPeer(peer ServerAddress) Future {
+	return r.AddVoter(ServerID(peer), peer, 0, 0)
 }
 
 // RemovePeer (deprecated) is used to remove a peer from the cluster. If the
 // current leader is being removed, it will cause a new election
 // to occur. This must be run on the leader or it will fail.
 // Use RemoveServer instead.
-func (r *Raft) RemovePeer(peer string) Future {
+func (r *Raft) RemovePeer(peer ServerAddress) Future {
```
slackpad: I originally wanted to keep this as `string` but this one was confusing (it's an address that we assume is the same as the ID). The trivial work to fix this is probably worth the clarity that the type provides.

ongardie-sfdc:
lgtm, thanks for fixing those tests.

PR #129, opened 2016-06-29
doc fixes, goroutine WaitGroup, export Leader field

rogpeppe:
Some minor doc fixes to make comments more Go idiomatic
and alter/add some things that seemed unclear to me
when writing a proof-of-concept FSM.

Add a comment in deferError to clarify how things
work with a nil error - at first glance it looks like
a bug (also add some explicit unit tests of deferFuture).

Track the running goroutines with a WaitGroup so we
don't need to actively poll the goroutine count.

Export the Leader field from LeaderObservation, making the
observation value more useful.

### DIFF
--- a/config.go
+++ b/config.go
@@ -10,15 +10,17 @@ import (
 // Config provides any necessary configuration to
 // the Raft server
 type Config struct {
-	// Time in follower state without a leader before we attempt an election.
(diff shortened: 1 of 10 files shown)


slackpad:
This LGTM - thanks!

PR #130, opened 2016-06-30
Misc cleanups

sean-:
Been sitting on this for a while, but want to push it out of my tree.


ongardie-sfdc:
Did a very quick scan through this, looks fine to me. It might cause some conflicts with issue-84-integration, but that's better to resolve sooner than later, I think.

slackpad:
Yep - I'll go ahead and integrate this and the other cleanup PR now and merge into the branch since we are between changes.

PR #132, opened 2016-07-06
Split apart and test appendConfigurationEntry

ongardie-sfdc:
This fixes a bug where AddNonvoter would previously demote a server from
Staging to Nonvoter (the intended semantics are no-op in that case).

This is box 14 on James's post- hashicorp/raft#117 checklist on hashicorp/raft#84.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -956,9 +956,7 @@ func (r *Raft) runLeader() {
 	}()
 
 	// Start a replication routine for each peer
-	for _, server := range r.configurations.latest.Servers {
-		r.startReplication(server)
-	}
+	r.startStopReplication()
 
 	// Dispatch a no-op log entry first. This gets this leader up to the latest
 	// possible commit index, even in the absence of client commands. This used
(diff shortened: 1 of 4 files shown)


raft.go
```diff
@@ -977,41 +975,49 @@ func (r *Raft) runLeader() {
...
-func (r *Raft) startReplication(server Server) {
-	if server.ID == r.localID {
-		return
-	}
+// startStopReplication will set up state and start asynchronous replication to
+// new peers, and stop replication to removed peers. Before removing a peer,
+// it'll instruct the replication routines to try to replicate to the current
+// index.
+func (r *Raft) startStopReplication() {
+	inConfig := make(map[ServerID]bool, len(r.leaderState.replState))
```
slackpad: This should probably be `len(r.configurations.latest.Servers)`. This'll be 0 at this time.

slackpad: Nevermind - not for the stop case.

raft.go
```diff
@@ -977,41 +975,49 @@ func (r *Raft) runLeader() {
...
-func (r *Raft) stopReplication(server Server, index uint64) {
-	if server.ID == r.localID {
-		return
+	// Start replication goroutines that need starting
+	for _, server := range r.configurations.latest.Servers {
+		if server.ID == r.localID {
+			continue
+		}
+		inConfig[server.ID] = true
+		r.logger.Printf("[INFO] raft: Added peer %v, starting replication", server)
```
slackpad: This would be good to put down after the `!present` check so it only logs if we are starting replication. Also, might be good to log the ID here since we log that when removing a server below.

raft.go
```diff
@@ -977,41 +975,49 @@ func (r *Raft) runLeader() {
...
+		if _, present := r.leaderState.replState[server.ID]; !present {
```
slackpad: I was initially worried that this wouldn't create a new replication task if there was one in effect that was draining (if you added and removed a peer quickly), because the old code didn't have the `!present` check, but we delete from `replState` whenever we stop replication so I think we are good.

slackpad: For idiomatic go-ness we should rename `present` to `ok`.

raft.go
```diff
@@ -1232,23 +1238,22 @@ func (r *Raft) quorumSize() int {
...
-func (r *Raft) appendConfigurationEntry(future *configurationChangeFuture) {
-	configuration := cloneConfiguration(r.configurations.latest) // same as committed
-
-	if future.prevIndex > 0 && future.prevIndex != r.configurations.latestIndex {
+// nextConfiguration generates a new Configuration from the current one and a
+// configuration change request. It's split from appendConfigurationEntry so
+// that it can be unit tested easily. This function reports errors directly to
+// the future; it returns an empty configuration if and only if
+// future.responded is true.
+func nextConfiguration(current Configuration, currentIndex uint64, future *configurationChangeFuture) Configuration {
```
slackpad: I <3 non-member functions for things like this!

raft_test.go
```diff
@@ -983,6 +984,180 @@ func TestRaft_ApplyConcurrent_Timeout(t *testing.T) {
...
+		prevIndex:     1,
+	}
+	future.init()
+	next := nextConfiguration(Configuration{}, 2, future)
+	if len(next.Servers) > 0 || !future.responded ||
+		!strings.Contains(future.Error().Error(), "changed") {
+		t.Fatalf("nextConfiguration should have failed due to intervening configuration change")
+	}
+
+	// Current prevIndex.
```
slackpad: How are these cases getting past the `checkConfiguration` call with no voters (just one staging server)?

slackpad: Ah I forgot we bypassed the promotion logic. These should probably be changed to `AddVoter` since these will start to fail later.

slackpad:
@ongardie just a few minor comments, otherwise LGTM!

raft.go
ongardie-sfdc: Even then, you're still right: there'll be one entry in the map per len(r.configurations.latest.Servers), perhaps except for the local server.

raft.go
ongardie-sfdc: Good point. Logging the entire `server` will contain the ID, right? Cool?

raft.go
ongardie-sfdc: Not for this PR, but are we ok with two replication routines for the same peer? Might be ok, might be confusing.

raft.go
slackpad: Yeah it can't hurt to use that as the size for sure.

raft.go
slackpad: Yeah this is a superset so it will have the ID. I was worried more about consistency but we probably need to make a sweep over the logging related to IDs and can do that separately.

raft.go
slackpad: Agree - we should think that through - I'll add it to the list.

raft.go
ongardie-sfdc: I guess. The signature is uglier than I hoped. Maybe the future should have a nested struct called `req`, and this should take a `req` and return a `(Configuration, error)`?

The other ugly thing is that we can't log in this function. But overall it's better than pulling in all the test setup of `raft_test.go`.

raft.go
ongardie-sfdc: Ah, got it. I'll switch this one to ID for now, but agreed--I haven't been watching the log messages at all.

raft.go
slackpad: Yeah agree the future at this level is a bit ugly. I'd be in favor of a nested struct in the future as the argument to this with `(Configuration, error)` as the return so it's more natural to call this.

raft_test.go
ongardie-sfdc: Ain't got AddVoter. Switched to basing off `singleServer` instead of empty `Configuration{}`.

raft.go
ongardie-sfdc: Cool, will do.

ongardie-sfdc: That cleaned things up nicely.

ongardie-sfdc:
Pushed a few more, I think that addressed all the comments.

slackpad:
Looks good - thanks!

PR #134, opened 2016-07-08
Fixes races with Raft configurations member, adds GetConfiguration accessor.

slackpad:
This adds `GetConfiguration()` which is item 7 from https://github.com/hashicorp/raft/issues/84#issuecomment-228928110. This also resolves item 12.

While adding that I realized that we should move the "no snapshots during config changes" outside the FSM snapshot thread to fix racy access to the configurations and also because that's not really a concern of the FSM. That thread just exists to let the FSM do its thing, but the main snapshot thread is the one that should know about the configuration.

~~To add an external `GetConfiguration()` API I added an RW mutex which I'm not super thrilled about, but it lets us correctly manipulate this state from the outside, and among the different Raft threads.~~


ongardie:
Haven't looked through this patch yet, but I was sort of thinking GetConfiguration would be treated like  the other application calls that get shoved into a channel as some future type for leaderLoop to process without additional locking. Will read through and think more on this later.

slackpad:
We could definitely do it with channels, I think. I'll take a crack at that and see if it simplifies things.

slackpad:
@ongardie this is ready for a look. Using a channel was definitely more Go-like and kept a lot of lock cruft out of the main thread where we can just work the the configuration directly.

configuration_test.go
```diff
@@ -27,14 +27,35 @@ var sampleConfiguration Configuration = Configuration{
...
+		}
+		cloned.Servers[1].ID = "scribble"
+		if sampleConfiguration.Servers[1].ID == "scribble" {
+			t.Fatalf("cloned configuration shouldn't alias Servers")
+		}
 	}
-	cloned.Servers[1].ID = "scribble"
-	if sampleConfiguration.Servers[1].ID == "scribble" {
-		t.Fatalf("cloned configuration shouldn't alias Servers")
+
```
ongardie-sfdc: Split here into two separate test functions?

raft.go
```diff
@@ -297,6 +301,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		stable:                stable,
 		trans:                 trans,
 		verifyCh:              make(chan *verifyFuture, 64),
+		configurationsCh:      make(chan *configurationsFuture, 64),
```
ongardie-sfdc: At first glance, that's a big buffer on configurationsCh.

raft.go
```diff
@@ -602,17 +632,24 @@ func (r *Raft) Stats() map[string]string {
...
+		"last_log_index":      toString(lastLogIndex),
+		"last_log_term":       toString(lastLogTerm),
+		"commit_index":        toString(r.getCommitIndex()),
+		"applied_index":       toString(r.getLastApplied()),
+		"fsm_pending":         toString(uint64(len(r.fsmCommitCh))),
+		"last_snapshot_index": toString(lastSnapIndex),
+		"last_snapshot_term":  toString(lastSnapTerm),
+	}
+
+	configuration, _, err := r.GetConfiguration()
```
ongardie-sfdc: oops, didn't realize this was an external call before. I'm starting to think we should move all these application-level API calls into a separate file, perhaps called api.go.

raft.go
slackpad: Yeah I was taking the other size - I'll lower it to 8 which should be enough for anything.

raft.go
slackpad: It's probably more natural to keep all the public stuff in raft.go but maybe split all the rest into raft_internal.go. I'll give that a shake to see how it looks.

ongardie-sfdc:
I'm still trying to wrap my head around how snapshots are taken...

raft.go
```diff
@@ -1988,64 +1949,85 @@ func (r *Raft) shouldSnapshot() bool {
...
 
-	// Create a new snapshot
+	// We don't support snapshots while there's a config change outstanding
+	// since the snapshot doesn't have a means to represent this state. This
+	// is a little weird because we need the FSM to apply an index that's
+	// past the configuration change, even though the FSM itself doesn't see
+	// the configuration changes. It should be ok in practice with normal
+	// application traffic flowing through the FSM. If there's none of that
+	// then it's not crucial that we snapshot, since there's not much going
+	// on Raft-wise.
```
ongardie-sfdc: Why would little/no normal traffic cause the state machine to fall behind on applying entries?

slackpad:
There's the `runSnapshots()` goroutine that calls `takeSnapshot()` at the right intervals. I think the confusing thing is that `takeSnapshot()` does an async ping to `runFSM()` (another goroutine) and asks it to take a snapshot of the FSM, then waits on the results of that via a future. That future has the snapshot of the FSM, and now I use another async / future to pull the configuration and do the check.

We used to have the config stuff on the FSM side, but that was one level deeper than it needed to be, and the FSM doesn't have anything to do with the configuration changes.

I suspect that this division of things might be a bit of a leftover from an earlier version of things, but the one nice thing about this architecture is that the FSM is either taking a snapshot, restoring a snapshot, or applying changes - all the serialization happens by virtue of the single `runFSM()` loop and it reading from channels.

raft.go
```diff
@@ -1988,64 +1949,85 @@ func (r *Raft) shouldSnapshot() bool {
...
 		return err
 	}
-	defer req.snapshot.Release()
+	defer snapReq.snapshot.Release()
 
-	// Log that we are starting the snapshot
-	r.logger.Printf("[INFO] raft: Starting snapshot up to %d", req.index)
+	// Make a request for the configurations and extract the committed info.
+	// We have to use the future here to safely get this information since
+	// it is owned by the main thread.
```
ongardie-sfdc: Good, safer. Who knew we had all these races? :/

raft.go
slackpad: It wouldn't cause the state machine to fall behind. Imagine that the last commit at index X was a config change. The FSM won't see this log, so when you try to snapshot the FSM it will be at X-1 so this will hit the check above and not write the snapshot. Once X+1 comes along and the FSM ingests it then we will be happy to snapshot again with the config change at index X.

ongardie-sfdc:
@slackpad, thanks, after reading through, that makes sense to me now.

raft.go
slackpad: This depends on the *FSM* having seen some index past the config change in order to take a snapshot, which relies on some application-related logs to come through and hit the FSM.

raft.go
slackpad: I did it by inspection but ran `go test -race` and it saw some of these, too :-)

raft.go
ongardie-sfdc: Would the users of the library care?

raft.go
ongardie-sfdc: Ah. In LogCabin, the state machine is fed no-op entries to increment its "last applied" index.

raft.go
slackpad: Probably not :-) I just would look first in raft.go to see how to use the library. Calling it api.go makes me think of some kind of API client layer, but that's just me.

raft.go
slackpad: Interesting - that would prevent this wrinkle.

ongardie-sfdc: It seems like processLog() could just write to fsmCommitCh for any type of log entry, and the FSM already filters out uninteresting ones. We'd just have to remove the switch statement out of processLog().

raft.go
ongardie-sfdc: Meh, your call. If someone's looking in raft.go to see how to use the library today, I think Armon owes them a beer.

raft.go
slackpad: I added item 21 to #84 to track this since it's probably better done in a separate, small PR.

ongardie-sfdc:
Well, lgtm. We can split those files in a later PR too if that's easier.

slackpad:
It was easy enough I rolled in the split as the last change. Thanks for taking a look!

ongardie-sfdc:
Also, thanks for so eagerly porting over to use channels. Didn't expect that to happen so fast last night :)

slackpad:
The locks did make me feel gross, and like 95% of them were almost always useless b/c it's all in a single thread - channels were dramatically better in this case!

Issue #108, opened 2016-03-22
Make leader leases actual leases


jellevandenhooff:
The current version of the library supports leases, but these leases are not quite leases in the traditional sense. Traditional leases come with guarantees that no one else will be able to obtain a conflicting lease while the lease is valid, yet the current leases do not come with such guarantees. This was originally revealed by Jepsen testing, summarized in a [consul consistency gist](https://gist.github.com/armon/11059431), and resulted in Consul's [funny consensus modes](https://www.consul.io/docs/internals/consensus.html) where "default" is not quite consistent and "consistent" introduces extra latency.

This issue proposes to have leaders wait a period of LeaderLeaseTimeout at the beginning of their term to make leases correct. This delay would make all reads behave as consistent reads, at the cost of slightly longer elections. With the default leader lease timeout of 500ms in both Consul and Nomad, I think this is a reasonable trade-off to make.

For background, the [extended Raft paper](https://ramcloud.stanford.edu/raft.pdf) mentions leases at the end of section 8:

> Second, a leader must check whether it has been deposed before processing a read-only request (its information may be stale if a more recent leader has been elected). Raft handles this by having the leader exchange heartbeat messages with a majority of the cluster before responding to read-only requests. Alternatively, the leader could rely on the heartbeat mechanism to provide a form of lease [9], but this would rely on timing for safety (it assumes bounded clock skew).

Consul's "consistent" reads work with the prescribed method of exchanging heartbeats to verify leadership. Consul's "default" reads attempt to implement leases as prescribed end of the paragraph, but the paragraph does not quite provide enough guidance to do so safely:

The library currently uses the function checkLeaderLease in raft.go to verify if a leader is still the leader. That function checks if the leader has contacted at least a quorum of nodes in LeaderLeaseTimeout. While this provides a good heuristic for the leader still being the leader, it does not provide any guarantees. Specifically, consider the following scenario, with three nodes A, B, C. Node A starts out at as the leader. Node A and C are partitioned and can not communicate with each other, but both can reach node B:

T1: A sends a heartbeat to A, B, and C. Because C is partitioned, it does not receive the heartbeat. A and B reply successfully. Now A believes its lease is valid.
T2, before the lease has expired: C, being partitioned, has not heard from A for while, and decides to hold an election for itself. C requests a vote from A, B, and C with a new term. Because A is partitioned, it does not receive the vote request. Both B and C vote for C, as it is the only node currently up for election. Now C believes it is the leader for new term.
[...]

PR #136, opened 2016-07-12
Rework goroutines and synchronization

ongardie-sfdc:
Today, the division of work and the synchronization between goroutines gets to be hard to follow in places. I think we can do better, to make the library more maintainable and eliminate potential race conditions from accidentally shared state. Ideally, it'll become more unit testable too.

This commit includes a diagram and description of where I think we should go. I'm open to feedback on it. Some of it's probably underspecified, with details to be determined as we implement more; questions are fair game too.

I held back on subdividing the main Raft module into a nonblocking goroutine and blocking helpers, but it's something we could consider. I haven't studied the code enough to know whether that'd be feasible or advantageous.

The transition from here to there is going to take significant effort. Here are a few of the major differences:
 - Peer is structured completely differently from replication.go today.
 - Peer handles all communication including RequestVote, not just AppendEntries/InstallSnapshot as replication.go does today.
 - Fewer locks and shared state. commitment.go and raftstate.go remove locking/atomics, possibly merge into raft.go. Other goroutines don't get a handle to the Raft module's state.
 - Snapshots are created through a different flow.

I started on the replication.go/peer.go changes, but it was before I had a good idea of where things were heading. I'll be happy to pick that up again later.

/cc @superfell @cstlee @bmizerany @kr @slackpad @sean- hashicorp/raft#84


slackpad:
@ongardie-sfdc thanks for putting this together! I'm still wrapping my head around a few things (and traveling) so I'll have some more thorough feedback in a bit, but here are a couple of initial impressions:

The split of responsibilities and channel-based interfaces will be huge for testability, especially for cases that we currently can't drive things into today; we will now be able to inject whatever we want to these various channels. I'm super excited about this aspect of the design.

The short-lived goroutines kind of put me off initially as a performance smell but I see how they lead to nice, clean interactions with Peer non-blocking loop and as a checkpoint for the term check in the middle, etc. I'm thinking we might be able to create some pools of goroutines that we reuse and have that also be part of our work rate limiting. I need to think on these changes in general and how they can affect performance, and how pipelining will look. This might be a bogus worry as well, need to think more on it :-)

This will definitely be a large effort and will take some time. In the near term I'd like to try to ship a release candidate version of Consul with the https://github.com/hashicorp/raft/issues/84 peer changes. This design will address several of our outstanding issues (especially the `AppendEntries` stale term checks) so I'm trying to weigh if it's practical to clean those up as simply as possible in the short term in the existing architecture and target this into a different, longer term integration branch to roll out in a later release of Consul.

kr:
> we might be able to create some pools of goroutines that we reuse

Definitely measure that approach before adopting it. Making a new goroutine is pretty cheap; can be cheaper than the alternative. For example:

https://github.com/golang/net/blob/f841c39d/http2/server.go#L610-L617
https://github.com/golang/net/blob/f841c39d/http2/server.go#L848-L851

This was a surprising result; this way turned out faster than sending frames to a long-running goroutine to be written.

> as a performance smell … need to think more on it

For performance, I'd strongly encourage an empirical approach over intuition or even careful thought. You'll be surprised how hard it is to predict where the performance problems are (and aren't!). 😄

ongardie:
I've now reworked the replication/peer side of things, but I need to get the existing tests passing again. And then there's adding actual unit tests now that that's possible. I'll post for a review one of these days. Pipelining fit in easily using the Transport's future return value, not the additional channel as before.

So regarding short-term shipping vs making this change, I think the biggest part is already mostly done. The rest will be more incremental, like removing dependencies on shared Raft/RaftState state from other parts of the library.

slackpad:
@ongardie if you've got a branch some place I'm definitely willing to take a look!

ongardie-sfdc:
Well, this patch is about ready for review, but the project might feel less 1337 now.

    18 files changed, 2064 insertions(+), 1337 deletions(-)

Despite that, I'll post it for review this afternoon.

commands.go
```diff
@@ -29,10 +29,6 @@ type AppendEntriesResponse struct {
 
 	// We may not succeed if we have a conflicting entry
 	Success bool
-
-	// There are scenarios where this request didn't succeed
-	// but there's no need to wait/back-off the next attempt.
-	NoRetryBackoff bool
```
ongardie-sfdc: Removed this because I don't think there are legitimate scenarios where the leader *should* back off upon receiving a reply. Backoff happens now only in response to transport errors.

config.go
```diff
@@ -18,11 +18,6 @@ type Config struct {
 	// a leader before we attempt an election.
 	ElectionTimeout time.Duration
 
-	// CommitTimeout controls the time without an Apply() operation
```
ongardie-sfdc: I think this was here before so that Follower's commitIndexes would be updated in the absence of new entries to replicate. I still have no clue why it was randomized.

Now, commitIndex should be sent as part of every heartbeat. Is it important to send out new commitIndexes before the next heartbeat? We could accelerate the next heartbeat to achieve that goal, and I could see needing a config option for it, but let's name and document it clearly.

inmem_transport.go
```diff
@@ -248,6 +255,7 @@ func (i *inmemPipeline) decodeResponses() {
 				case i.doneCh <- inp.future:
 				case <-i.shutdownCh:
 					return
+				default: // can't write to doneCh, probably because it filled up. meh.
```
ongardie-sfdc: This is a little gross. And there might be some case still where this transport leaks the decodeResponses() goroutine and prevents a clean shutdown, though I'm not seeing it lately.

Do we want to update the Transport pipeline interface to remove the doneCh entirely?

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"time"
+
+	"github.com/armon/go-metrics"
+)
+
+type peerOptions struct {
```
ongardie-sfdc: // Settings controlling Peer behavior, as passed to startPeer().

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+	// Helper goroutines send fully prepared RPC requests onto requestCh.
+	// The main Peer goroutine validates each request against the latest
+	// control information, then spawns another goroutine to send it.
+	requestCh chan peerRPC
+
+	// Helper goroutines send RPC replies onto replyCh for the main Peer goroutine
+	// to process.
+	replyCh chan peerRPC
+
+	pipelineSendDoneCh chan struct{}
```
ongardie-sfdc: // Ugly wart used to not overwhelm AppendEntries pipeline.

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+	// Set to true while an InstallSnapshot RPC is in progress. This prevents
+	// further InstallSnapshot and bulk AppendEntries RPCs from starting.
+	outstandingInstallSnapshotRPC bool
+
+	// Counts the number of non-heartbeat AppendEntries and InstallSnapshot RPCs
+	// that have been sent but have not completed (in either error or response).
+	// When not pipelining, such RPCs are only sent when this 0.
+	outstandingAppendEntriesRPCs uint64
+
+	outstandingPipelineSend bool
```
ongardie-sfdc: // Part 2 of ugly wart to not overwhelm AppendEntries pipeline.

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+func (rpc *appendEntriesRPC) sendRecv(shared *peerShared) error {
+	numEntries := uint64(len(rpc.req.Entries))
+	defer func() {
+		metrics.MeasureSince([]string{"raft", "replication", "appendEntries", "rpc", string(shared.peerID)}, rpc.start)
+		metrics.IncrCounter([]string{"raft", "replication", "appendEntries", "logs", string(shared.peerID)}, float32(numEntries))
+	}()
+	var desc string
+	switch {
+	case rpc.heartbeat:
+		desc = fmt.Sprintf("heartbeat (term %v, prev %v term %v)",
```
ongardie-sfdc: This gets to be verbose. Do we have some way of squelching log messages based on conf?

peer_test.go
```diff
@@ -0,0 +1,450 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+// TODO: more tests for AppendEntries
+
+///////////////////////// InstallSnapshot /////////////////////////
+
+// TODO: more tests for InstallSnapshot
```
ongardie-sfdc: I don't know of any blockers to writing these, but I think it makes sense to get a review on the basic approach first.

raft.go
```diff
@@ -174,44 +189,43 @@ func (r *Raft) runCandidate() {
 	r.logger.Printf("[INFO] raft: %v entering Candidate state in term %v",
 		r, r.getCurrentTerm()+1)
 	metrics.IncrCounter([]string{"raft", "state", "candidate"}, 1)
+	// defer metrics.MeasureSince([]string{"raft", "candidate", "electSelf"}, time.Now())
```
ongardie-sfdc: I don't actually know where these go, so it's hard for me to tell if I screwed anything up. Do we want this one, and what does it represent?

raft.go
```diff
@@ -174,44 +189,43 @@ func (r *Raft) runCandidate() {
...
-			}
-
-			// Check if we've become the leader
-			if grantedVotes >= votesNeeded {
-				r.logger.Printf("[INFO] raft: Election won. Tally: %d", grantedVotes)
-				r.setState(Leader)
-				r.setLeader(r.localAddr)
-				return
-			}
+			*/
```
ongardie-sfdc: I think we can remove this, enough logging in peer.go now.

raft.go
```diff
@@ -346,49 +353,78 @@ func (r *Raft) runLeader() {
...
-		// Replicate up to lastIdx and stop
-		r.logger.Printf("[INFO] raft: Removed peer %v, stopping replication after %v", serverID, lastIdx)
-		repl.stopCh <- lastIdx
-		close(repl.stopCh)
-		delete(r.leaderState.replState, serverID)
+		peer.controlCh <- control
+	}
+}
+
+func (r *Raft) shutdownPeers() {
```
ongardie-sfdc: // shutdownPeers() instructs all peers to exit immediately.

raft.go
```diff
@@ -520,100 +510,221 @@ func (r *Raft) leaderLoop() {
...
 }
 
-// checkLeaderLease is used to check if we can contact a quorum of nodes
-// within the last leader lease interval. If not, we need to step down,
-// as we may have lost connectivity. Returns the maximum duration without
-// contact. This must only be called from the main thread.
-func (r *Raft) checkLeaderLease() time.Duration {
-	// Track contacted nodes, we can always contact ourself
-	contacted := 1
+func (r *Raft) verified(verifiedCounter uint64) {
```
ongardie-sfdc: // Responds to any verify futures that have been satisfied. A majority of the voting servers in the cluster have acknowledged this server's leadership since the given count.

raft.go
```diff
@@ -520,100 +510,221 @@ func (r *Raft) leaderLoop() {
...
+// contact. This must only be called from the main thread.
+func (r *Raft) checkLeaderLease() {
+	servers := len(r.configurations.latest.Servers)
+	lastContacts := make([]uint64, 0, servers)
+	if hasVote(r.configurations.latest, r.localID) {
+		lastContacts = append(lastContacts, uint64(time.Now().UnixNano()))
+	}
+	for peerID, peer := range r.peers {
+		if hasVote(r.configurations.latest, peerID) {
+			lastContacts = append(lastContacts, uint64(peer.progress.lastContact.UnixNano()))
```
ongardie-sfdc: Note that peer.progress.lastContact is more conservative than before. It's the time the last completed RPC was started. That cool? This should help #108 "Make leader leases actual leases".

raft.go
```diff
@@ -812,15 +921,16 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
...
+		r.updateTerm(a.Term)
 		resp.Term = a.Term
 	}
-
 	// Save the current leader
+	r.stepDown()
 	r.setLeader(ServerAddress(r.trans.DecodePeer(a.Leader)))
+	defer r.setLastContact()
+	// TODO: moving setLastContact up here is the right thing to do, but it poses a compatibility problem.
+	// It may be better to send heartbeats of (index 0, term 0) to old servers or something.
```
ongardie-sfdc: To be more specific, the problem now is that a new Leader will heartbeat with prevLogIndex set to its last log index. If an old Follower (running old code) receives such a heartbeat but is behind in its log, it won't invoke setLastContact(). This may cause the old Follower to time out.

raft.go
```diff
@@ -346,49 +353,78 @@ func (r *Raft) runLeader() {
...
+			term:          r.getCurrentTerm(),
+			role:          role,
+			verifyCounter: verifyCounter,
+			lastIndex:     lastIndex,
+			lastTerm:      lastTerm,
+			commitIndex:   r.getCommitIndex(),
 		}
-		// Replicate up to lastIdx and stop
-		r.logger.Printf("[INFO] raft: Removed peer %v, stopping replication after %v", serverID, lastIdx)
-		repl.stopCh <- lastIdx
```
ongardie-sfdc: The new code makes no attempt to continue replicating entries to a removed server. With the membership changes the way they are now, that didn't actually cause any benefit AFAIK. Let's talk about what we need to happen here.

raft.go
```diff
@@ -520,100 +510,221 @@ func (r *Raft) leaderLoop() {
...
+		/*
+					// Log at least once at high value, then debug. Otherwise it gets very verbose.
+					if diff <= 3*r.conf.LeaderLeaseTimeout {
+						r.logger.Printf("[WARN] raft: Failed to contact %v in %v", peer, diff)
+					} else {
+						r.logger.Printf("[DEBUG] raft: Failed to contact %v in %v", peer, diff)
+					}
+				}
+			metrics.AddSample([]string{"raft", "leader", "lastContact"}, float32(diff/time.Millisecond))
+		*/
```
ongardie-sfdc: drop all this?

ongardie-sfdc:
Sorry for the many emails. Blame GitHub.

This is ready for your review now, @slackpad @sean- @superfell and any other generous folks.

peer.go
sean-: Not at present, but this has come up a few times and is something we want to
eventually address, but haven't decided on how we're going to do it yet.  Any
time-based governor would work as a throttle right now.

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+		default:
+			return
+		}
+	}
+}
+
+// blockingSelect reads/writes the Peer channels just once, blocking if needed.
+func (p *peerState) blockingSelect() {
+	// We need to send a heartbeat at lastHeartbeatSent + heartbeatInterval.
+	heartbeatTimer := time.After(p.shared.options.heartbeatInterval -
```
ongardie-sfdc: Need an if statement around this so that Followers and Candidates don't fire all the time.

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+	// pipeline (a separate path through the Transport), false otherwise.
+	allowPipeline bool
+
+	// If true, the peer needs a snapshot because this server has already
+	// discarded the log entries it would otherwise send in AppendEntries.
+	needsSnapshot bool
+
+	// Set to true while an InstallSnapshot RPC is in progress. This prevents
+	// further InstallSnapshot and bulk AppendEntries RPCs from starting.
+	outstandingInstallSnapshotRPC bool
```
ongardie-sfdc: These few "outstanding*" fields should probably be per term (cleared out when the term changes and only decremented if the term hasn't changed). Otherwise after a term switch, you might needlessly wait around for the old ones to finish/timeout.

sean-: Can we move the per-Term variables into its own `struct` that way we can more easily guarantee they are cleared out every term by re-initing a new `struct`?  From a code-maintenance and readability perspective this would be very useful.

commands.go
slackpad: Yeah this is an interesting one. The types of things where we didn't set this to true were were kind of a miscellaneous grab of application errors like not being able to write to the log. I'd probably rather see the back pressure done as a delayed response (such as if we did retries writing to the log, etc.) vs. an immediate response asking for a back off. Removing this and keeping it for comm errors seems like the way to go.

config.go
slackpad: The timing is a lot more aggressive than a heartbeat and given how Consul works with allowing stale reads this is probably pretty important for good performance there. Accelerating heartbeats seems equivalent. Still working through the diff so I'm not sure if that would lead to an unclean coupling of things.

inmem_transport.go
```diff
@@ -10,7 +10,7 @@ import (
 // NewInmemAddr returns a new in-memory addr with
 // a randomly generate UUID as the ID.
 func NewInmemAddr() ServerAddress {
-	return ServerAddress(generateUUID())
+	return ServerAddress(generateUUID()[:4])
```
slackpad: Good call!

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+	"log"
+	"os"
+	"time"
+
+	"github.com/armon/go-metrics"
+)
+
+type peerOptions struct {
+	// Where to print debug messages, or nil for stderr.
+	logger *log.Logger
```
slackpad: This might be better passed in at start time and saved in `peerShared`. We end up with two references as it stands now.

peer.go
slackpad: These variables are making me want a simple state machine here. That might clean up this state quite a bit.

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+		didSomething := p.issueWork()
+		// If we've got no more work to do, block until something changes.
+		if !didSomething {
+			p.blockingSelect()
+		}
+	}
+
+	p.shared.logger.Printf("[INFO] raft: Peer routine for %v exiting", p.shared.peerID)
+	close(p.shared.stopCh)
+	p.shared.pipeline.Close()
```
slackpad: This'll crash if there's no pipeline support.

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+	return nil
+}
+
+// selectLoop is a long-running routine that sends RPCs to a single remote
+// server.
+func (p *peerState) selectLoop() {
+	for {
+		// At this point, we may or may not have more work that we could do right away.
+		// Check for new information but don't block.
+		p.drainNonblocking()
```
slackpad: Are there correctness concerns with the drain or is this a performance optimization to avoid calling `issueWork()` too often? It seems like we could simplify things a bit by always doing a blocking select and then calling `issueWork()` each time. It seems like the one special case we might want would be something like this:

`
for {
  select {
    // blocking select here over everything
  }

  select {
    // do a non-blocking poll of just the control channel one time to make sure we have latest
  }

  p.issueWork()
}
```

That would ensure we are getting the latest control info before we respond to any other operation.

File: raft.go (hunk 14 above)
Comment (slackpad) [inline raft.go]: We could probably make this conditional on protocol version and only support it for v0.

File: inmem_transport.go (hunk 19 above)
Comment (ongardie-sfdc) [inline inmem_transport.go]: It's definitely the best part of this patch.

File: peer.go (hunk 22 above)
Comment (ongardie-sfdc) [inline peer.go]: In the general case, new information could cause multiple things to become issue-able. Like let's say the top blocking select got us an RPC reply allowing a snapshot to be sent. And let's say the control select got us a new verifyCounter value to send a heartbeat. We want to do both before blocking again.

File: peer.go (hunk 18 above)
Comment (ongardie-sfdc) [inline peer.go]: will try

File: peer.go [hunk 23]
```diff
@@ -0,0 +1,1087 @@
...
+//          |
+//          '-------spawn-----> sendRecv
+//                                 |
+//                                 '------------> Transport
+//                                                    |
+//                                 .------------------'
+//                                 |
+//       process <--replyCh--------'
+//
+type peerRPC interface {
```
slackpad: Random thought - should we split this along the lines of "runs in the main thread" vs. "runs in a goroutine"? It seems like you have to read everything carefully to know what's safe where.

peer.go
```diff
@@ -0,0 +1,1087 @@
...
+	needsSnapshot bool
+
+	// Set to true while an InstallSnapshot RPC is in progress. This prevents
+	// further InstallSnapshot and bulk AppendEntries RPCs from starting.
+	outstandingInstallSnapshotRPC bool
+
+	// Counts the number of non-heartbeat AppendEntries and InstallSnapshot RPCs
+	// that have been sent but have not completed (in either error or response).
+	// When not pipelining, such RPCs are only sent when this 0.
+	outstandingAppendEntriesRPCs uint64
```
ongardie-sfdc: cap this

peer.go
slackpad: If you loop again, though, it won't block - it should keep draining channels and calling `issueWork()` until there's nothing to do.

config.go
ongardie-sfdc: The plan is to accelerate heartbeats. We don't think that'll be too expensive, and we'll find out through benchmarking if we're wrong.

inmem_transport.go
ongardie-sfdc: Let's simplify the pipeline interface in a separate PR.

raft.go
ongardie-sfdc: Plan is to add a stopit flag to RequestVoteResponse. Maybe better to do in a separate PR.

raft.go
ongardie-sfdc: kept metric only

peer.go
ongardie-sfdc: @slackpad, what I was concerned about was, using your proposed loop structure:

1. Blocking select: get an RPC reply that enables a snapshot to be sent.
2. Poll control: get a new verifyCounter that enables a heartbeat to be sent.
3. issueWork: sends heartbeat.
4. Blocking select: blocks.

We should have sent that snapshot instead of blocking. Oops.

And in general, it'd be nice if a single receive can enable multiple actions to fire. This leads to a structure like:

```
for {
  if issueWork() {
    continue
  }
  blockingSelect()
}
```

But then the concern is that blockingSelect() has no priority, so this might result in suboptimal schedules. For example, it might receive an RPC reply rather than new control information with a new term number. If we acted immediately, we might send an RPC that we could have avoided by polling. Are these suboptimal schedules a big deal? Shrug.

So do we poll just the control channel but block on all the channels? Which ones are really important to justify polling on? The channels are these:
 1. recv control
 2. send progress
 3. recv prepared RPC request
 4. recv RPC reply
 5. recv pipeline send completed
 6. recv backoff timer expired
 7. recv heartbeat timer expired (you can't poll this the way the code is structured now)

I think that neglecting (1) and (4) and possibly (3) could result in sending suboptimal RPC requests.

Then it's a judgment call:
 - duplicate none (my for loop above): It's simplest. We'll embrace any performance hit until we learn better.
 - duplicate all (the original patch): It's mechanical. There's no tough decision to make about whether a channel goes in both the poll and the blocking select: they all do.
 - duplicate some: If we pick the right subset, we should get the performance benefits of duplicating all but with less code duplication.

slackpad: I like your alternate structure above. With that in place we could start at "duplicate none" and then optimize further by adding a pre-poll in `blockingSelect()` to look for some high priority cases before we block on everything. That would let us have the simpler outer structure and keep the priority parts all together inside a single function. Starting with "duplicate none" and tweaking the priority as we measure is probably a good way to go - we might not need any duplication at all.

raft.go
ongardie-sfdc: I think it's working out to be fine to just send 0 values in the new code too. If we change our minds later, we can condition it on a suitable protocol version then.

config.go
ongardie-sfdc: Accelerated AppendEntries requests, not heartbeats. Same effect, just easier.

peer.go
ongardie-sfdc: If I get what you're saying, it didn't seem advantageous to me when I tried. Feel free to give it a shot (now or later).

ongardie-sfdc:
Heads up @slackpad, 601dcb8 could use close review, and it's one you'd probably want to backport if you're working on a release right now.

peer.go
slackpad: I'm ok then - the interface has clear documentation and isn't super wide. Thanks for trying!

raft.go
ongardie-sfdc: just kidding, hold tight

ongardie-sfdc: there's a big comment in the latest code explaining why it uses the matchIndex instead now

ongardie-sfdc:
This one is ready for review again. I think it's ready to merge, other than:
 - I asked a couple of unanswered questions in the GitHub comments above.
 - Upon merging, it'll need to set the new RPCHeader and InstallSnapshotRequest.SnapshotVersion fields.
 - Any new review comments that may come in from y'all (please take my sanity into consideration).

slackpad:
@ongardie-sfdc thanks for the heads up on https://github.com/hashicorp/raft/commit/601dcb8bb1d13648cc94cefa909c34d874b68293. I'll take a detailed look at that and the rest of the changes here. Do you want to do a rebase first?

ongardie-sfdc:
@slackpad, do I have to?

slackpad:
We will have to figure out how to merge :-) I'd be ok doing the integration for you if you want some help with that.

ongardie-sfdc:
@slackpad, right, right. Let's finish up the review process pre-merge, then I can squash+rebase or merge in the latest issue-84-integration changes here. I think the only thing that'll change on this end is adding the RPCHeader and InstallSnapshotRequest.SnapshotVersion fields.

slackpad:
Go it - yeah this shouldn't have too many material conflicts. I'll try to alert you as I go through for a this review pass.

api.go
```diff
@@ -253,17 +259,19 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
...
-		fsm:           fsm,
-		fsmCommitCh:   make(chan commitTuple, 128),
-		fsmRestoreCh:  make(chan *restoreFuture),
-		fsmSnapshotCh: make(chan *reqSnapshotFuture),
-		leaderCh:      make(chan bool),
-		localID:       localID,
-		localAddr:     localAddr,
-		logger:        logger,
-		logs:          logs,
+		peerProgressCh: make(chan peerProgress, 0),
```
superfell: ,0 ? should leave it out for an unbuffered channel.

ongardie: Ok, it is meant to be unbuffered, I'll remove the `, 0`.

commands.go
rogpeppe: How important is backward compatibility? If providing backward compatibility, perhaps it would be better to retain this field but document that it's deprecated and ignored.

config.go
rogpeppe: Similar concern about backward compatibility.

inmem_transport.go
```diff
@@ -139,20 +139,27 @@ func (i *InmemTransport) makeRPC(target ServerAddress, args interface{}, r io.Re
...
 	// Send the RPC over
-	respCh := make(chan RPCResponse)
-	peer.consumerCh <- RPC{
+	respCh := make(chan RPCResponse, 1)
+	rpc := RPC{
 		Command:  args,
 		Reader:   r,
 		RespChan: respCh,
 	}
+	timer := time.After(timeout)
```
rogpeppe: Better to use time.NewTimer and defer timer.Stop, so that we don't leave excess timer objects when the timeout isn't triggered?

ongardie-sfdc: Since
- I didn't make this any worse in this PR (the `time.After` was there on line 155 before) and
- there are 13 other occurrences of `time.After` in the non-test code in the library,

I'm inclined not to address this as part of this PR.

@rogpeppe, if dangling timers are an issue, could you file this in a separate issue to audit the timer use in the entire library? And more importantly, get the documentation on https://golang.org/pkg/time/#After updated to describe the issue and its severity.

superfell:
I ran our test suite against this branch, need to pick through it some more, but it seemed to get stuck, when the test gets to the point where it stops Apply() new entries, and is waiting for all nodes to get to the final state, it seems like the followers aren't getting caught up, this is logged repeatedly

m_2:11:21:02.377443 [INFO] raft: Starting heartbeat RPC for m_0
m_2:11:21:02.377533 [INFO] raft: Starting heartbeat RPC for m_1
m_2:11:21:02.377599 [INFO] raft: Sending heartbeat (term 2) to m_0
m_2:11:21:02.377612 [INFO] raft: Sending heartbeat (term 2) to m_1
m_2:11:21:02.377682 [INFO] raft: AppendEntries to m_0 succeeded. nextIndex is now 41
m_2:11:21:02.377682 [INFO] raft: AppendEntries to m_1 succeeded. nextIndex is now 49
m_2:11:21:02.582696 [INFO] raft: Starting heartbeat RPC for m_0
m_2:11:21:02.582707 [INFO] raft: Starting heartbeat RPC for m_1
m_2:11:21:02.582840 [INFO] raft: Sending heartbeat (term 2) to m_0
m_2:11:21:02.582875 [INFO] raft: Sending heartbeat (term 2) to m_1
m_2:11:21:02.582918 [INFO] raft: AppendEntries to m_0 succeeded. nextIndex is now 41
m_2:11:21:02.582942 [INFO] raft: AppendEntries to m_1 succeeded. nextIndex is now 49

There's a lot of goroutines stuck waiting on the transport, e.g.

goroutine 521 [chan receive, 9 minutes]:
salesforce.com/raft/fuzzy.(*appendEntry).Error(0xc820e6ee10, 0x0, 0x0)
	/Users/simon.fell/code/fuzzyRaft/src/salesforce.com/raft/fuzzy/transport.go:274 +0x46
github.com/hashicorp/raft.(*appendEntriesRPC).sendRecv(0xc820e3ffc0, 0xc820af2340, 0x0, 0x0)
	/Users/simon.fell/code/fuzzyRaft/.godeps/src/github.com/hashicorp/raft/peer.go:1007 +0x44f
github.com/hashicorp/raft.sendHelper(0xc820af2340, 0x9c7840, 0xc820e3ffc0)
	/Users/simon.fell/code/fuzzyRaft/.godeps/src/github.com/hashicorp/raft/peer.go:513 +0x46
github.com/hashicorp/raft.(*peerState).send.func1()
	/Users/simon.fell/code/fuzzyRaft/.godeps/src/github.com/hashicorp/raft/peer.go:506 +0x42
github.com/hashicorp/raft.(*waitGroup).spawn.func1(0xc820472990, 0xc820e2bba0)
	/Users/simon.fell/code/fuzzyRaft/.godeps/src/github.com/hashicorp/raft/util.go:151 +0x4c
created by github.com/hashicorp/raft.(*waitGroup).spawn
	/Users/simon.fell/code/fuzzyRaft/.godeps/src/github.com/hashicorp/raft/util.go:152 +0x5d

ongardie:
@superfell repeated heartbeat messages in stderr are normal, but the stuck goroutines are not. Maybe there's something with the transport integration?

superfell:
will dig in more, why isn't m_0 making any progress though, it get to be at 49 same as m_1

ongardie:
@superfell it could be related to the goroutines wedged in the transport, since non-heartbeat AppendEntries are limited to 32 at a time when pipelining (and 1 when not)

superfell:
Am i reading this wrong, or is it sending the same AE request multiple times ?

m_1:14:35:09.284039 [INFO] raft: Sending pipelined AppendEntries (term 2, entries 22 through 22, prev 21 term 2, commit 19) to m_2
m_1:14:35:09.284049 [INFO] raft: Starting AppendEntries RPC for m_2
m_1:14:35:09.284119 [INFO] raft: Sending pipelined AppendEntries (term 2, entries 22 through 22, prev 21 term 2, commit 19) to m_2
m_1:14:35:09.284134 [INFO] raft: Starting AppendEntries RPC for m_2
m_1:14:35:09.284200 [INFO] raft: Sending pipelined AppendEntries (term 2, entries 22 through 22, prev 21 term 2, commit 19) to m_2
m_1:14:35:09.284208 [INFO] raft: Starting AppendEntries RPC for m_2
m_1:14:35:09.284282 [INFO] raft: Sending pipelined AppendEntries (term 2, entries 22 through 22, prev 21 term 2, commit 19) to m_2

ongardie:
Yeah @superfell, that looks like a bug. Or somehow the RPC could be failing without printing a message, then being retried quickly. But I think more likely a bug in updating nextIndex.

superfell:
I think its down to how fast issueWork gets triggered

superfell:
looks like issueWork only checks the number of issued AE requests before deciding to send another one, I don't see where it track of whats AE is inflight.

superfell:
i think you'll hit this if issueWork is trigger faster than the AE RPC builder can read log entries

ongardie:
@superfell I see. Since nextIndex isn't updated until `confirm()`, we need to block sending more AppendEntries before then. This happens with the `outstandingPipelineSend` bool for pipelined RPCs but not for non-pipelined RPCs. RequestVote and InstallSnapshot might have similar issues, too.

ongardie:
Still, shouldn't it eventually make progress, even if it sends every RPC 32 times? :)

superfell:
Indeed, that seems to be a different issue to the first issue.

superfell:
Ahh, so nothing appears to read from the pipeline Consumer channel anymore. so its gets as far as filling that up, then my pipeline transport blocks.

ongardie:
Yeah, with this change Peer is no longer using the transport's pipeline Consumer(). I was hoping to update the transport interface(s) in a separate PR, but maybe we need to break that API atomically. Alternatively, I guess we could spawn a goroutine in Peer to consume and discard the pipeline futures.

ongardie-sfdc:
I wasn't able to just spawn an extra goroutine to consume and discard pipeline futures because at least inmem_transport wasn't erroring out all futures on the shutdown path. I think it's better to break the API explicitly so people simplify and audit their transports. Did so in e5e58b7.

cc @superfell

ongardie-sfdc:
Fixed two distinct bugs causing redundant RPCs to be sent in 93436b5 (and updated the unit tests to better catch this type of thing).

ongardie:
One difference between peer.go and replicate.go is that Peer keeps around a single Pipeline and uses it forever, whereas replicate.go would close the pipeline and reopen it if anything slightly bad happened. Is that alright?

slackpad:
@ongardie (haven't had time to re-review but I will in the next few days)

> One difference between peer.go and replicate.go is that Peer keeps around a single Pipeline and uses it forever, whereas replicate.go would close the pipeline and reopen it if anything slightly bad happened. Is that alright?

Without looking at the code again yet, this sounds suspicious if there was ever a write failure or partial, non-atomic write and the pipeline TCP stream gets corrupted.

ongardie-sfdc:
With de57f4d, Peer still assumes the pipeline won't fail, and a new "reliable pipeline" wrapper handles errors under the hood.

ongardie-sfdc:
Hey I just discovered this thing where net_transport.go tries to detect a heartbeat based on it being all zeros (no longer true with this PR), then directly invokes a function that's probably not safe to run concurrently. I think we should disable it for now and revisit later.

ongardie-sfdc:
^@slackpad please add to issue-84 checklist

slackpad:
@ongardie-sfdc added as item 25

ongardie-sfdc:
hey @slackpad @sean-, any ETA on reviewing this?

slackpad:
Sorry for the delay @ongardie-sfdc - I will review this week!

api.go
```diff
@@ -372,6 +382,7 @@ func (r *Raft) restoreSnapshot() error {
 			r.configurations.latest = configuration
 			r.configurations.latestIndex = snapshot.Index
 		}
+		r.updatePeers()
```
slackpad: This is going to get called on line 332 in `NewRaft()` is there a good reason to call it here? We might change the configuration later when we look through the log entries.

commands.go
slackpad: @rogpeppe we've got a number of non-compatible changes in this version of the library already, though they are generally pretty easy to deal with (it took about 2 hours to get Consul building and booting again). We intend to keep this in a branch and once it's done we will message the community and give a little bit of time for people to adapt to the changes before we take it to master. Whenever we can make something safer (like ID and addresses as separate types) or simpler we are breaking things. We've also got a story for how folks can interoperate with newer versions of the library during the transition - https://github.com/hashicorp/raft/blob/library-v2-stage-one/config.go#L10-L115.

config.go
```diff
@@ -18,11 +18,6 @@ type Config struct {
...
-	// before we heartbeat to ensure a timely commit. Due to random
-	// staggering, may be delayed as much as 2x this value.
-	CommitTimeout time.Duration
```
slackpad: Note to self - I think this is still useful as a tune-able. Need to look at how the new implementation works.

future.go
```diff
@@ -153,14 +150,8 @@ type restoreFuture struct {
 	ID string
 }
 
-// verifyFuture is used to verify the current node is still
```
slackpad: Did you mean to delete the comments here?

slackpad:
Made it about 1/3 through but I'll be out tomorrow, so this'll spill over into next week. Things are looking really good so far. Will update once I pick up again (may have a little time on Saturday).

config.go
ongardie-sfdc: Note to James: the new commit index is now just sent at the first opportunity with no artificial delay. I'm not sure why you'd ever need to delay it (it should be low overhead).

future.go
ongardie-sfdc: That was an accident. I think I had deleted the struct entirely and then added it back but forgot the comment.

peer.go
```diff
@@ -0,0 +1,1277 @@
...
+	// during control.term. When not pipelining, this is capped at 1.
+	// When pipelining, this is capped at peerOptions.maxPipelineWindow.
+	outstandingAppendEntriesRPCs uint64
+
+	// allowPipeline is true when AppendEntries RPCs should be sent over a
+	// pipeline (a separate path through the Transport), false otherwise.
+	// If allowPipeline is true, pipelineUnsupported must be false.
+	allowPipeline bool
+
+	// Set to true when we're an AppendEntries request is being sent on the
```
slackpad: Comment doesn't make sense.

peer.go
```diff
@@ -0,0 +1,1277 @@
...
+	return p
+}
+
+func (p *peerState) checkInvariants() error {
+	if p.shared.peerID != p.progress.peerID {
+		return errors.New("Progress must have peerID")
+	}
+	if p.progress.lastContact.After(p.progress.lastReply) ||
+		p.progress.lastContact == p.progress.lastReply && !p.progress.lastContact.IsZero() {
+		return errors.New("lastContact must be after lastReply (or both 0)")
```
slackpad: I think this error message is backwards.

raft.go
slackpad: This'll have the time to complete an election from the candidate's perspective (one election cycle). Probably useful to just uncomment.

raft.go
slackpad: I think that's fine. We may want to revisit the default config (and the Consul config and docs) based on this.

transport.go
```diff
@@ -89,19 +99,26 @@ type WithPeers interface {
 	DisconnectAll()                          // Disconnect all peers, possibly to reconnect them later
 }
 
-// AppendPipeline is used for pipelining AppendEntries requests. It is used
+// AppendPipeline2 is used for pipelining AppendEntries requests. It is used
```
slackpad: The "2" seems gross but I can't come up with anything better right now. I'll revisit this after you merge :-)

slackpad:
@ongardie-sfdc I've made a pass through everything and I think the basic structure of this is sound. Pointed out a couple tiny things in the comments.

Let's go ahead and rebase or merge with the latest branch contents and get this into issue-84-integration. Over there I'm going to make another super detailed pass through the peer logic and the unit tests, as well as try to get some benchmarks going that we can pick to the stage one branch to get a before/after.

slackpad:
And sorry it took so long to get back to this!

ongardie-sfdc:
Thanks @slackpad, that sounds good to me. I'll be on vacation next week and most of the following, but I'm going to fix up the little nits and then try to rebase or merge before I go.

ongardie-sfdc:
I think 8a71deb is the last of my non-merge-related changes.

ongardie-sfdc:
Merged issue-84-integration into ongardie/modules. My confidence that I got that merge perfectly right isn't terribly high, but it's probably good enough to call done. Unit tests and a quick run seem ok.

slackpad:
Cool I'll scan through it again for merge weirdness and bring it over to
the branch.

ongardie:
woot

Issue #137, opened 2016-07-13
Return a struct from `func (*Raft) Stats`


WillSewell:
At the moment it just returns a map of strings. Currently it is unclear what keys the map will contain without looking at the source.

It would be nice if this was a struct so that the caller would know the format of what was being returned by looking at the type.

If you think this is worth doing, then I am happy to implement it.

rogpeppe:
I like the idea, but that would break backward compatibility. which is probably not a good idea for a relatively minor improvement. There should at least be a doc comment that explains what the expected keys are.

otoolep:
Initially I too was surprised that it just strings, as I thought it would be more explicit. That said, backwards compatibility is important, and I use this call in [rqlite](https://github.com/rqlite/rqlite/blob/master/store/store.go#L408). It might break if it changed.

WillSewell:
Fair enough. I agree that backwards compatibility is more important than a bit of type safety.

I could add an _additional_ function that returns the stats as a struct. Otherwise I'm happy to document the keys.

slackpad:
Hi @WillSewell I think this was always intended more for humans to dump and look at vs. using the information programatically. Documenting the keys is probably the best way to go for this one :-)

WillSewell:
That's the impression I got.

Our use case is using it as the response for an endpoint which returns the current status of raft. A client uses uses the endpoint for health checks, so it's nice to know up front what the keys are.

PR #138, opened 2016-07-16
Document map fields returned by Stats()

WillSewell:
Follows from the discussion: https://github.com/hashicorp/raft/issues/137

Let me know if this is too little/too much.

I would have liked to have presented the fields as a list, but this would not format as a list in the HTML output unless it was preformatted (which I didn't think it should be).


slackpad:
Thanks!

PR #139, opened 2016-07-18
Adds in-place upgrade and manual recovery support.

slackpad:
This adds several important capabilities to help in upgrading to the new Raft protocol version:

1. We can migrate an existing peers.json file, which is sometimes the source of truth for the old version of the library before this support was moved to be fully in snapshots + raft log as the official source.
1. If we are using protocol version 0 where we don't support server IDs, operators can continue to use peers.json as an interface to manually recover from a loss of quorum.
1. We left ourselves open for a more full-featured recovery manager by giving a new RecoverCluster interface access to a complete Configuration object to consume. This will allow us to manually pick which server is a voter for manual elections (set 1 to a voter and the rest to nonvoters, the 1 voter will elect itself), as well as basically any other configuration we want to set.

This also gives a path for introducing Raft servers running the new version of the library into a cluster running the old code. Things would work like this:

```
// These are the versions of the protocol (which includes RPC messages as
// well as Raft-specific log entries) that this server can _understand_. Use
// the ProtocolVersion member of the Config object to control the version of
// the protocol to use when _speaking_ to other servers. This is not currently
// written into snapshots so they are unversioned. Note that depending on the
// protocol version being spoken, some otherwise understood RPC messages may be
// refused. See isVersionCompatible for details of this logic.
//
// There are notes about the upgrade path in the description of the versions
// below. If you are starting a fresh cluster then there's no reason not to
// jump right to the latest protocol version. If you need to interoperate with
// older, version 0 Raft servers you'll need to drive the cluster through the
// different versions in order.
//
// The version details are complicated, but here's a summary of what's required
// to get from an version 0 cluster to version 3:
//
// 1. In version N of your app that starts using the new Raft library with
//    versioning, set ProtocolVersion to 1.
// 2. Make version N+1 of your app require version N as a prerequisite (all
//    servers must be upgraded). For version N+1 of your app set ProtocolVersion
//    to 2.
// 3. Similarly, make version N+2 of your app require version N+1 as a
//    prerequisite. For version N+2 of your app, set ProtocolVersion to 3.
//
// During this upgrade, older cluster members will still have Server IDs equal
// to their network addresses. To upgrade an older member and give it an ID, it
// needs to leave the cluster and re-enter:
//
// 1. Remove the server from the cluster with RemoveServer, using its network
//    address as its ServerID.
// 2. Update the server's config to a better ID (restarting the server).
// 3. Add the server back to the cluster with AddVoter, using its new ID.
//
// You can do this during the rolling upgrade from N+1 to N+2 of your app, or
// as a rolling change at any time after the upgrade.
//
// Version History
//
// 0: Original Raft library before versioning was added. Servers running this
//    version of the Raft library use AddPeerDeprecated/RemovePeerDeprecated
//    for all configuration changes, and have no support for LogConfiguration.
// 1: First versioned protocol, used to interoperate with old servers, and begin
//    the migration path to newer versions of the protocol. Under this version
//    all configuration changes are propagated using the now-deprecated
//    RemovePeerDeprecated Raft log entry. This means that server IDs are always
//    set to be the same as the server addresses (since the old log entry type
//    cannot transmit an ID), and only AddPeer/RemovePeer APIs are supported.
//    Servers running this version of the protocol can understand the new
//    LogConfiguration Raft log entry but will never generate one so they can
//    remain compatible with version 0 Raft servers in the cluster.
// 2: Transitional protocol used when migrating an existing cluster to the new
//    server ID system. Server IDs are still set to be the same as server
//    addresses, but all configuration changes are propagated using the new
//    LogConfiguration Raft log entry type, which can carry full ID information.
//    This version supports the old AddPeer/RemovePeer APIs as well as the new
//    ID-based AddVoter/RemoveServer APIs which should be used when adding
//    version 3 servers to the cluster later. This version sheds all
//    interoperability with version 0 servers, but can interoperate with newer
//    Raft servers running with protocol version 1 since they can understand the
//    new LogConfiguration Raft log entry, and this version can still understand
//    their RemovePeerDeprecated Raft log entries. We need this protocol version
//    as an intermediate step between 1 and 3 so that servers will propagate the
//    ID information that will come from newly-added (or -rolled) servers using
//    protocol version 3, but since they are still using their address-based IDs
//    from the previous step they will still be able to track commitments and
//    their own voting status properly. If we skipped this step, servers would
//    be started with their new IDs, but they wouldn't see themselves in the old
//    address-based configuration, so none of the servers would think they had a
//    vote.
// 3: Protocol adding full support for server IDs and new ID-based server APIs
//    (AddVoter, AddNonvoter, etc.), old AddPeer/RemovePeer APIs are no longer
//    supported. Version 2 servers should be swapped out by removing them from
//    the cluster one-by-one and re-adding them with updated configuration for
//    this protocol version, along with their server ID. The remove/add cycle
//    is required to populate their server ID. Note that removing must be done
//    by ID, which will be the old server's address.
```

```
// These are versions of snapshots that this server can _understand_. Currently,
// it is always assumed that this server generates the latest version, though
// this may be changed in the future to include a configurable version.                                                                                                              //
// Version History
//
// 0: Original Raft library before versioning was added. The peers portion of
//    these snapshots is encoded in the legacy format which requires decodePeers
//    to parse. This version of snapshots should only be produced by the
//    unversioned Raft library.
// 1: New format which adds support for a full configuration structure and its
//    associated log index, with support for server IDs and non-voting server
//    modes. To ease upgrades, this also includes the legacy peers structure but
//    that will never be used by servers that understand version 1 snapshots.
//    Since the original Raft library didn't enforce any versioning, we must
//    include the legacy peers structure for this version, but we can deprecate
//    it in the next snapshot version.
```

This isn't super great, but will give us a path to keep things compatible with existing clusters as we roll out the changes. We can make some higher-level tooling in Consul to help orchestrate this.


slackpad:
@ongardie this should address TODO item 2 from #84. If you have a few PTAL - thanks!

api.go
```diff
@@ -223,47 +231,53 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
...
 
+	// Make sure we have a valid server address and ID.
+	protocolVersion := conf.ProtocolVersion
 	localAddr := ServerAddress(trans.LocalAddr())
 	localID := conf.LocalID
-	if localID == "" {
-		logger.Printf("[WARN] raft: No server ID given, using network address: %v. This default will be removed in the future. Set server ID explicitly in config.",
+	if protocolVersion < 1 || localID == "" {
+		// During the transition to the new ID system, keep this as an
+		// INFO level message. Once the new scheme has been out for a
```
sean-: Can you comment this as `COMPAT` and a `TODO` so it's easily grepable that there is a pending action to enable a depreciate `WARN` message?

config.go
```diff
@@ -7,9 +7,35 @@ import (
...
+//
+// Version History
+//
+// 0: Unversioned original protocol spoken until Q2 2016.
+// 1: Added server IDs and a new peer change mechanism via a new LogConfiguration
+//    log entry type. All servers must be running >= 1 in order to support new
+//    staging and nonvoter modes for servers.
+const (
+	ProtocolVersionMin uint8 = 0
+	ProtocolVersionMax       = 1
```
sean-: Is there a reason this isn't also `uint8` or explicitly a sized integer (e.g. `uint32`)?

config.go
```diff
@@ -96,6 +123,9 @@ func DefaultConfig() *Config {
 
 // ValidateConfig is used to validate a sane configuration
 func ValidateConfig(config *Config) error {
+	if config.ProtocolVersion > ProtocolVersionMax {
+		return fmt.Errorf("Protocol version cannot be > %d", ProtocolVersionMax)
```
sean-: Can you include the current protocol version that didn't match from the config in the error message?

configuration.go
```diff
@@ -274,17 +274,39 @@ func nextConfiguration(current Configuration, currentIndex uint64, change config
 	return configuration, nil
 }
 
+// encodePeers is used to serialize a Configuration into the old peers format.
+// This is here for backwards compatibility when operating with a mix of old
+// servers and should be removed eventually.
```
sean-: Add a little context for when this can be removed:

> removed when support for speaking with protocol version 0 clients is no longer supported.

raft.go
```diff
@@ -927,6 +940,23 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 	return
 }
 
+// checkAndProcessConfigurationLog takes a log entry and updates the latest
+// configuration if the entry results in a new configuration. This must only be
+// called from the main thread.
```
sean-: I haven't looked too deeply, but `checkAndProcessConfigurationLog()` is called from two places.  Does it make sense to put a guard in place w/ an atomic compare and swap op to make sure this isn't ever run concurrently?  If so, would that relax the restriction on this being called from only the main thread?

config.go
slackpad: That's because I pasted it from elsewhere :-) Changing to an int since there's no reason to be specific.

sean-: Is `ProtocolVersionMax` ever serialized as a specific type?  Does it make sense to just declare this as a `uint32` so that the type has a fixed width?

slackpad: We don't serialize this at all, so I won't bake in any assumptions right now.

raft.go
slackpad: I don't think we want to relax this restriction with any locks since this should only apply in the main thread, ever. I'll update the comment about getting called from `NewRaft()` that's before any threads are running.

slackpad: Given what it does I might make it just a plain old function that takes the state in though - will be easier to test.

slackpad: Ah it needs the transport to support the old methods, not worth de-membering. When we kill the support for the old protocol we can nuke this fn and just inline the 4 lines that will matter at that point.

ongardie-sfdc:
sorry @slackpad, was too busy generating code, will review

slackpad:
@ongardie-sfdc thanks! If you have a chance please take a look at #140 as well - it's based on this branch so relative to this PR.

config.go
```diff
@@ -96,6 +123,11 @@ func DefaultConfig() *Config {
 
 // ValidateConfig is used to validate a sane configuration
 func ValidateConfig(config *Config) error {
+	if config.ProtocolVersion < ProtocolVersionMin ||
+		config.ProtocolVersion > ProtocolVersionMax {
+		return fmt.Errorf("Protocol version %d must be >= %d and <= %d",
+			config.ProtocolVersion, ProtocolVersionMin, ProtocolVersionMax)
+	}
 	if config.HeartbeatTimeout < 5*time.Millisecond {
```
ongardie-sfdc: Add a new validation check: when ProtocolVersion > 0, LocalID should be nonempty.

api.go
```diff
@@ -287,30 +304,24 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
...
 			panic(err)
 		}
-		// TODO: support LogAddPeer, LogRemovePeer too
-		if entry.Type == LogConfiguration {
-			r.configurations.committed = r.configurations.latest
-			r.configurations.committedIndex = r.configurations.latestIndex
-			r.configurations.latest = decodeConfiguration(entry.Data)
-			r.configurations.latestIndex = entry.Index
-		}
+		r.checkAndProcessConfigurationLog(&entry)
```
ongardie-sfdc: Please stop naming things "log" when you mean "log entry". It's confusing.

api.go
```diff
@@ -486,15 +497,24 @@ func (r *Raft) GetConfiguration() (Configuration, uint64, error) {
 // AddPeer (deprecated) is used to add a new peer into the cluster. This must be
 // run on the leader or it will fail. Use AddVoter/AddNonvoter instead.
 func (r *Raft) AddPeer(peer ServerAddress) Future {
-	return r.AddVoter(ServerID(peer), peer, 0, 0)
+	return r.requestConfigChange(configurationChangeRequest{
```
ongardie-sfdc: reject AddPeer and RemovePeer on protocolVersion >= 1?

config.go
```diff
@@ -7,9 +7,35 @@ import (
 	"time"
 )
 
+// These are the versions of the protocol (which includes wire protocol as
```
ongardie-sfdc: by wire protocol, you mean the message fields sent over the network, but not their encoding?

Also, are snapshots included in this protocol version or not?

config.go
```diff
@@ -7,9 +7,35 @@ import (
...
 type Config struct {
+	// ProtocolVersion allows a Raft server to inter-operate with older
+	// Raft servers running an older version of the code. This is used to
+	// version the wire protocol as well as Raft-specific log entries that
+	// the server uses when _speaking_ to other servers. There is currently
+	// no auto-negotiation of versions so all servers must be manually
+	// configured with compatible versions. See ProtocolVersionMin and
+	// ProtocolVersionMax for the versions of the protocol that this server
+	// can _understand_.
+	ProtocolVersion int
```
ongardie-sfdc: When I set this to 1, will the server start speaking protocol 1, or will it stay on 0 until 1 is safe to use?

api.go
ongardie-sfdc: While you're at it, it's not really checking much, so I think you can drop the "checkAnd" too.

ongardie-sfdc:
This sets some good groundwork.

I think having the config dictate which version the server speaks may be too simplistic. Here's an example: Bring up server A with version n + 1  while servers B and C still have version n. Server A can send B a LogConfiguration entry, then crash. B can become leader and have to replicate that LogConfiguration. But B isn't allowed to speak n + 1 yet. That's awkward.

slackpad:
@ongardie-sfdc agree the manual part isn't ideal but we don't have a good mechanism to negotiate. I was thinking I'd ship the next release of Consul (N) w/protocol version set to 0 so the upgrade works with the old servers, but all new servers will be able to understand 1. The release after that (N+1) would switch to protocol version 1, and we'd make N a prerequisite for N+1.

I suppose we could try to add a version to some existing messages and then upshift if we see that, but given that we can add incompatible things to the log I don't think that will work if an older server suddenly shows up later.

ongardie-sfdc:
> ... I don't think that will work if an older server suddenly shows up later.

If an older server suddenly shows up later with old code, the best thing for it to do is crash. If we're clever, there might be some way to make that happen, though there aren't many panics around.

slackpad:
I just meant if we are in protocol 0 mode we should work with old servers, even if we bootstrap a new cluster. In the >= 1 world, old servers will print errors for all peer changes and will never be able to accept them. Will have to do a little thinking/testing to see if that's enough to keep them out of a cluster. With our current message serialization library we could add the protocol version to all our RPCs and reject anything from 0 (which it will default to on new servers) if we are > 0 - this might be a clean way to stop shenanigans after the upgrade is complete.

api.go
slackpad: That's a good idea, and will force an update of all the unit test code as well.

sean-:
Automatically add them to the blacklist and ignore them.

config.go
slackpad: I'll clarify this comment it's just Raft-specific log entries and RPC messages. It doesn't currently apply to snapshots, though we could eventually use it there if we needed to.

config.go
slackpad: The way it's set up now it'll start speaking 1 right away - I'll do some more thinking about this since I ran into a snag when making the change for https://github.com/hashicorp/raft/pull/139#discussion_r71607693.

slackpad:
Merged #140 into here and I'll continue the cleanup work on this PR - these two features were closely related.

ongardie-sfdc:
@slackpad, that's cool, but I hope you're not trying to compete with #136 on the diff size.

slackpad:
Realized that we will need a more dynamic "upshift" to protocol 1  (or some other solution) to handle this situation:

1. Cluster is all running P0.
2. Roll to new nodes that understand P1 but are speaking P0. All good, and they set their `localID` to their address for compatibility.
3. Roll a node so it starts speaking P1.

At this last step we get into trouble because this node _won't_ set its `localID` to its address, so it won't think it has a vote if it becomes a leader. Not sure the cleanest way to handle this - I'll see what I can come up with. I'm thinking maybe the `localID` isn't optional and I tack in a new-style configuration onto the end of `LogRemovePeerDeprecated` that the P1-understanding servers can use to update their configuration properly (and that P0-understanding servers would ignore). Will update.

ongardie-sfdc:
@slackpad FYI, I'm not able to keep up on this one this week. I should be able to start thinking through these issues on Monday.

slackpad:
@ongardie-sfdc no problem - I'll need some time tomorrow to finalize this. In parallel I've got a branch of Consul with a basic integration of this branch and it's forming clusters just fine (that's helping me make sure the upgrade stuff is going to actually work) :-)

slackpad:
Figured out a solution over the weekend and working on cleaning it up and checking it in. Should have a review-ready state here in a few hours.

slackpad:
@ongardie-sfdc I think this one is ready for review. I ended up adding an intermediate protocol version to help with the transition. This is not as nice as I would like, but I could not come up with a simpler scheme to transition from no server IDs to full server IDs.

api.go
```diff
@@ -146,73 +154,152 @@ type Raft struct {
...
+	if err := logs.StoreLog(entry); err != nil {
+		return fmt.Errorf("failed to append configuration entry to log: %v", err)
+	}
+
+	return nil
+}
+
+// IsClusterBootstrapped returns true if the cluster has been bootstrapped
+// before.
+func IsClusterBootstrapped(logs LogStore, stable StableStore, snaps SnapshotStore) (bool, error) {
```
ongardie-sfdc: This is a misleading name (maybe I started it). You can't tell if the cluster has been bootstrapped. At best you can tell if this server has any state.

config.go
```diff
@@ -7,9 +7,74 @@ import (
...
+//
+// Version History
+//
+// 0: Unversioned, original protocol, used to interoperate with unversioned
+//    Raft servers. Under this version all configuration changes are propagated
+//    using the deprecated RemovePeerDeprecated Raft log entry. This means that
+//    server IDs are always set to be the same as the server addresses (since
+//    the old log entry type cannot transmit an ID), and only AddPeer/RemovePeer
+//    APIs are supported. This version can understand the new LogConfiguration
+//    Raft log entry but it will never generate one.
```
ongardie-sfdc: How could this version possibly understand the new LogConfiguration entries? They're not part of `master` today.

slackpad: New servers running in version 0 mode, I'll update the comment.

api.go
```diff
@@ -146,73 +154,152 @@ type Raft struct {
...
+		entry.Data = encodeConfiguration(configuration)
+	}
+	if err := logs.StoreLog(entry); err != nil {
+		return fmt.Errorf("failed to append configuration entry to log: %v", err)
+	}
+
+	return nil
+}
+
+// RecoverCluster is used to manually force a new configuration in order to
```
slackpad: Use snapshots for this - this isn't safe b/c we can create conflicting log entries.

slackpad:
@ongardie-sfdc switched recovery over to the new snapshot-based approach. If you can give that a quick eyeball and I'll add some additional tests to close this out. Thanks!

slackpad:
Ok I'm going to stop touching this one - should be good for review!

api.go
```diff
@@ -146,73 +154,228 @@ type Raft struct {
...
+
+	// Compact the logs so that we don't get bad interference from any
+	// configuration change log entries that might be there.
+	firstLogIndex, err := logs.FirstIndex()
+	if err != nil {
+		return fmt.Errorf("failed to get first log index: %v", err)
+	}
+	if err := logs.DeleteRange(firstLogIndex, lastLogIndex); err != nil {
+		return fmt.Errorf("log compaction failed: %v", err)
+	}
```
ongardie-sfdc: Don't we want all the stores cleared out except for this snapshot?

api.go
```diff
@@ -146,73 +154,228 @@ type Raft struct {
...
+
+	// Sanity check the Raft peer configuration.
 	if err := checkConfiguration(configuration); err != nil {
 		return err
 	}
 
-	// Make sure we don't have a current term
+	// Refuse to recover if there's no existing state; this isn't safe to
+	// do because we'd have to synthesize a log entry.
```
ongardie-sfdc: I don't understand why this check is needed.

api.go
slackpad: We aren't messing with the term, so it seemed safe to leave the stable store alone. We do clear the logs, and the older snapshots seem good to retain in case of emergencies.

api.go
slackpad: Ah yeah - it would probably work in practice but I was trying to be super careful here. If there's nothing in the logs then you'd have to produce a snapshot with all the indexes at zero which didn't seem like a good idea since there's no normal way to get a cluster into that state. You should just call `BootstrapCluster()` instead.

api.go
```diff
@@ -223,47 +386,53 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
...
+	protocolVersion := conf.ProtocolVersion
 	localAddr := ServerAddress(trans.LocalAddr())
 	localID := conf.LocalID
-	if localID == "" {
-		logger.Printf("[WARN] raft: No server ID given, using network address: %v. This default will be removed in the future. Set server ID explicitly in config.",
+
+	// TODO (slackpad) - When we deprecate protocol version 1, remove this
+	// along with the AddPeer() and RemovePeer() APIs.
+	if protocolVersion < 2 {
+		logger.Printf("[INFO] raft: Configured ProtocolVersion < 2, using network address as LocalID: %v",
```
ongardie-sfdc: If I'm at protocolVersion 1 and have set a localID != localAddr, what's supposed to happen?

slackpad: In `protocolVersion` 1 you have to have your `localID == localAddr` in order for you to recognize your own suffrage properly, so for that case we force them to be equal no matter what you set them to.

api.go
ongardie-sfdc: Ok, so I'm with you that recovering an empty server is 99.9% likely to mean that the user is doing something dumb, and we should probably disallow it as a policy. OTOH, I think it'd be correct as a mechanism. How about:
- s/this isn't safe to do because we'd have to synthesize a log entry/this is dumb/
- s/cannot recover cluster with no initial state/Refusing to recover a server with no initial state. Seriously, you should ask for help./

api.go
```diff
@@ -146,73 +154,228 @@ type Raft struct {
...
+		break
+	}
+	if len(snapshots) > 0 && (snapshotIndex == 0 || snapshotTerm == 0) {
+		return fmt.Errorf("failed to restore any of the available snapshots")
+	}
+
+	// The snapshot information is the best known end point for the data
+	// until we play back the Raft log.
+	finalIndex := snapshotIndex
+	finalTerm := snapshotTerm
```
ongardie-sfdc: nit: Probably better to call these lastIndex, lastTerm. Maybe it's just me, but final seems too const.

api.go
```diff
@@ -146,73 +154,228 @@ type Raft struct {
...
+
+	// Apply any Raft log entries past the snapshot.
+	lastLogIndex, err := logs.LastIndex()
+	if err != nil {
+		return fmt.Errorf("failed to find last log: %v", err)
+	}
+	var lastLog Log
+	if lastLogIndex > 0 {
```
ongardie-sfdc: I don't get why we need this if-block to read the last log entry. Isn't it about to get read in the for loop?

api.go
```diff
@@ -146,73 +154,228 @@ type Raft struct {
...
+			return fmt.Errorf("failed to get last log at index %d: %v", lastLogIndex, err)
+		}
+	}
+	for index := snapshotIndex + 1; index <= lastLog.Index; index++ {
+		var entry Log
+		if err := logs.GetLog(index, &entry); err != nil {
+			return fmt.Errorf("failed to get log at index %d: %v", index, err)
+		}
+		if entry.Type == LogCommand {
+			_ = fsm.Apply(&entry)
```
ongardie-sfdc: Out of curiosity, why the `_ =`? Does it have meaning to the compiler or some linter?

api.go
```diff
@@ -146,73 +154,228 @@ type Raft struct {
...
+		return fmt.Errorf("failed to finalize snapshot: %v", err)
+	}
+
+	// Compact the logs so that we don't get bad interference from any
+	// configuration change log entries that might be there.
+	firstLogIndex, err := logs.FirstIndex()
+	if err != nil {
+		return fmt.Errorf("failed to get first log index: %v", err)
+	}
+	if err := logs.DeleteRange(firstLogIndex, lastLogIndex); err != nil {
```
ongardie-sfdc: Is logs.DeleteRange(0, infinity) not cool?

api.go
```diff
@@ -146,73 +154,228 @@ type Raft struct {
...
+		return fmt.Errorf("log compaction failed: %v", err)
+	}
+
+	return nil
+}
+
+// HasExistingState returns true if the cluster has any existing state (logs,
```
ongardie-sfdc: s/cluster/server/

api.go
ongardie-sfdc: Should a difference cause an (fatal) error or (nonfatal) warning?

api.go
```diff
@@ -506,6 +686,10 @@ func (r *Raft) RemovePeer(peer ServerAddress) Future {
 // If nonzero, timeout is how long this server should wait before the
 // configuration change log entry is appended.
 func (r *Raft) AddVoter(id ServerID, address ServerAddress, prevIndex uint64, timeout time.Duration) IndexFuture {
+	if r.protocolVersion < 1 {
+		return errorFuture{ErrUnsupportedProtocol}
+	}
+
```
ongardie-sfdc: In protocolVersion == 2, do we need id == address here?

commands.go
```diff
@@ -1,8 +1,24 @@
 package raft
 
+// VersionInfo is a common sub-structure used to pass along
+// protocol version information. For older Raft implementations before
+// versioning was added this will default to protocol version 0.
+type VersionInfo struct {
```
ongardie-sfdc: If we added cluster IDs (we should someday), would we want them in this struct? I wonder if this should be named something to the effect of "Common".

api.go
slackpad: Ok - changed the error to include the phrase "this is probably an operator error" and added wording to the comment that this would work, but we are choosing not to do this because it's probably the wrong thing to do.

api.go
slackpad: The way it's set up now, they can start setting the `LocalID` right away, and then they get an INFO-level message when things start up saying "we are using the network address as the LocalID". I figured this was safest so we don't require them to do something special in the config to make things match.

inmem_store.go
```diff
@@ -82,6 +82,10 @@ func (i *InmemStore) DeleteRange(min, max uint64) error {
 		delete(i.logs, j)
 	}
 	i.lowIndex = max + 1
```
ongardie-sfdc: Let's assume that DeleteRange will delete a prefix of the log, a suffix of the log, neither, or the entire thing. I'm not a unit test, but I don't think this updates lowIndex and highIndex correctly when deleting a suffix, either before or after your change.

api.go
slackpad: I took that from your `NewRaft()` log scanning code :-) It should be the same as the last index we retrieved, but this actually gets the last log and sees what index is set on it. We could probably eliminate this in both places.

api.go
slackpad: Nah - I just wanted the code to read "I am explicitly ignoring the return here and I thought about it".

log.go
```diff
@@ -26,7 +26,8 @@ const (
 	LogBarrier
 
 	// LogConfiguration establishes a membership change configuration. It is
-	// created when a server is added, removed, promoted, etc.
+	// created when a server is added, removed, promoted, etc. Only used
+	// when protocol version 1 or greater is in use.
```
ongardie-sfdc: GitHub disallows me from placing this comment where it should go on lines 14 and 18. Those should state "Only used when protocol version whatever is in use".

api.go
slackpad: Ah nm - you need the term. I'll get rid of it recover though.

api.go
slackpad: I didn't want to make any assumptions about the log store, this should be equivalent though.

raft.go
```diff
@@ -753,6 +798,19 @@ func (r *Raft) processLog(l *Log, future *logFuture) {
...
 func (r *Raft) processRPC(rpc RPC) {
+	if v, ok := rpc.Command.(WithVersionInfo); ok {
+		if !r.isVersionCompatible(v.GetVersionInfo()) {
+			r.logger.Printf("[ERR] raft: Ignoring unsupported %#v for command: %#v", v, rpc.Command)
+			rpc.Respond(nil, ErrUnsupportedProtocol)
+			return
+		}
+
+	} else {
+		r.logger.Printf("[ERR] raft: Ignoring un-versioned command: %#v", rpc.Command)
```
ongardie-sfdc: How does this work? Don't we need to listen to unversioned commands until we've reached some version ourselves?

ongardie-sfdc: And isn't that the same as ErrUnsupportedProtocol?

raft.go
```diff
@@ -753,6 +798,19 @@ func (r *Raft) processLog(l *Log, future *logFuture) {
 // processRPC is called to handle an incoming RPC request. This must only be
 // called from the main thread.
 func (r *Raft) processRPC(rpc RPC) {
+	if v, ok := rpc.Command.(WithVersionInfo); ok {
+		if !r.isVersionCompatible(v.GetVersionInfo()) {
+			r.logger.Printf("[ERR] raft: Ignoring unsupported %#v for command: %#v", v, rpc.Command)
+			rpc.Respond(nil, ErrUnsupportedProtocol)
```
ongardie-sfdc: Does this send back what version this server is running? If not, I wonder if we might regret that later.

api.go
slackpad: This only works for protocol >0, which unfortunately needs to support both ways, so there's nothing we can enforce here. You need to be able to remove protocol version 1 servers by address and add the new protocol version 2 servers by ID, so there's a small burden on the app during that time to call this in the right way.

commands.go
slackpad: That's a good idea and this is exactly where that would go. How about `RPCHeader`?

recovery.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Configuration is the override configuration.
+	Configuration Configuration
+}
+
+// NewPeersJSONRecovery takes the given base directory and parses the peers.json
+// file at that location, returning a recovery manager to apply this
+// configuration at startup. If there's no recovery defined then this will return
+// nil, which is a valid thing to pass to NewRaft().
+func NewPeersJSONRecovery(base string) (*PeersJSONRecovery, error) {
```
ongardie-sfdc: A "recovery manager" doesn't seem to really be a thing, and I don't think it needs to be. How about simply:

    func ReadPeersJSON(path) (Configuration, error)

The caller would have the burden of `path := filepath.Join(base, "peers.json")` and `os.Remove(path)`, which I think would be entirely tolerable.

commands.go
ongardie-sfdc: Cool, sure.

raft.go
slackpad: Unversioned commands come through as version 0 because of the way MsgPack decoding works - the new receivers get zero-valued version info. I updated the comment to reflect this.

recovery.go
slackpad: Good catch - I originally had this horrible thing that was passed in to `NewRaft()` that waited for things to replicate, etc. This is a leftover from that so I'll rename this and remove the `Disarm()` shiz.

api.go
ongardie-sfdc: You know I didn't write it in NewRaft because the comment reads "Get the log." I would have written "Get the last log entry", if anything. The log vs entry conflation makes me cringe. Every time.

slackpad: Doh - I fixed that in `NewRaft()`.

ongardie-sfdc: Haha, thanks, just a pet peeve of mine.

raft.go
slackpad: This currently doesn't. I think it would probably be more trouble than it's worth to add at this point, because the error object itself is sent back.

config.go
```diff
@@ -7,9 +7,89 @@ import (
 	"time"
 )
 
+// These are the versions of the protocol (which includes RPC messages as
+// well as Raft-specific log entries) that this server can _understand_. Use
+// the ProtocolVersion member of the Config object to control the version of
+// the protocol to use when _speaking_ to other servers. This is not currently
+// written into snapshots so they are unversioned. Note that depending on the
```
ongardie-sfdc: Any reason not to write some version number into snapshots?

config.go
```diff
@@ -7,9 +7,89 @@ import (
...
+// to get from an unversioned cluster to version 2:
+//
+// 1. In version N of your app that starts using the new Raft library with
+//    versioning, set ProtocolVersion to 0.
+// 2. Make version N+1 of your app require version N as a prerequisite (all
+//    servers must be upgraded). For version N+1 of your app set ProtocolVersion
+//    to 1.
+// 3. Similarly, make version N+2 of your app require version N+1 as a
+//    prerequisite. For version N+2 of your app, set ProtocolVersion to 2. Also,
+//    for this upgrade users need to remove each server from the cluster by
```
ongardie-sfdc: That's not "for this upgrade". That's any time they want to change the server's ID, including upon this special event that unpegs the ID from the address.

config.go
```diff
@@ -7,9 +7,89 @@ import (
...
+//    address before upgrading that server, and then add it back by ID after the
+//    upgrade.
+//
+// Version History
+//
+// 0: Unversioned, original protocol, used to interoperate with unversioned
```
ongardie-sfdc: For cleanliness, I think:
- v0 should refer to the original code running the original protocol before LogConfiguration existed.
- v1 should generate RemovePeerDeprecated but understand LogConfiguration.
- v2 should generate LogConfiguration with id == address but understand LogConfiguration with id != address.
- v3 should generate LogConfiguration with id != address.

raft.go
slackpad: If we get a message without a version at all it's kind of a bug in the code, so I wanted that to be distinguishable from a normal "talking to an old, unsupported thing" error.

config.go
slackpad: Since the snapshots are somewhat divorced from Raft log entry types I figured it would be overkill. You can think of them as being at "snapshot format version 0" :-)

config.go
slackpad: The part I was trying to emphasize here was the "remove by address" part for the old servers.

config.go
slackpad: I do agree this is cleaner. The only argument against this is that I don't want to actually implement protocol 0 in the new code. I suppose we could make this change and set the min protocol version to 1, which would not be too bad of a change and would make what 0 means super clear.

ongardie-sfdc: The min version is supposed to be what this server understands, and has to understand version 0 to do a rolling upgrade from 0 to 1. Perhaps the minimum version a server understands ought to be different from the minimum version it may run.

slackpad: Ah yeah you are correct. We can check them separately.

config.go
ongardie-sfdc: Something like this? It's a little verbose, sorry.

s/

> Also, for this upgrade users need to remove each server from the cluster by address before upgrading that server, and then add it back by ID after the upgrade.

/

> After this upgrade, the cluster members will still have Server IDs equal to their network addresses. If you want to change a server's ID, it needs to leave the cluster and re-enter:
1. Remove the server from the cluster with RemoveVoter, using its network address as its ServerID.
2. Update the server's config to a better ID (restarting the server).
3. Add the server back to the cluster with AddVoter, using its new ID.
You can do this during the rolling upgrade from N+1 to N+2 of your app, or you can do it as a rolling change at any time after the upgrade.

/

config.go
ongardie-sfdc: One thing we need to guarantee is that if you don't understand something in your log or snapshot, you crash. It's not cool to skip it like we're doing now.

This could be a server was running new code, wrote out a log entry/snapshot with important new information, then somehow got downgraded.
 * For log entries, we can probably agree to do this using new log entry types. But we're missing the check in NewRaft and Recover right now.
 * For snapshots, we don't have a way to know.

Otherwise, this could be that we've deleted the code that reads in old log entries/snapshot formats. Someday we need to remove that cruft. How can we guarantee that no log entry or snapshot is so old? I think the solution has to be of the form: write out a new snapshot in a newer format before you upgrade to code that stops knowing how to read older log entry/snapshot formats. How do we arrange that?

BTW, I think all this might lead us to having multiple versions covering different things, at least internally. You can't read in snapshots with newer versions than what you support, but the snapshot format won't change on every upgrade. So I think it needs its own class of version number.

api.go
ongardie-sfdc: I'm not sure that ignoring what they wrote in the config is the safest thing to do. It seems safer to force them to set their config's ID == to their local address.

api.go
ongardie-sfdc: I thought that while this server is in what used to be called version 1, it must never generate a LogConfiguration with ID != addr. Is that correct?

recovery.go
ongardie-sfdc: phew

ongardie-sfdc:
Ok, sorry for delaying a few days. I think I'm done with this pass of reviewing. It's a hairy issue, but thanks for taking it on, and you're making great progress. Have a good weekend.

slackpad:
@ongardie-sfdc thanks for taking a look. I'm finishing up the last bit of changes now and will push up shortly. There's a follow on change in https://github.com/hashicorp/raft/pull/143 that I'll rebase off of this.

Appreciate your detailed scrub of this one - have a good weekend as well!

api.go
slackpad: That's a fair point - it's better to be super explicit. I'll change this.

api.go
slackpad: > I thought that while this server is in what used to be called version 1, it must never generate a LogConfiguration with ID != addr. Is that correct?

That's ok to do when running 1 (now 2) because everybody in the cluster understands `LogConfiguration`. This is how you will be able to add version 2 (now 3) to the cluster by ID. The only requirement is that servers running 1 (now 2) need to have their ID set to their address so that you can work with the old version 0 (now 1) servers while rolling them out of the cluster.

config.go
slackpad: This should probably be done as a separate PR, but we could add a snapshot version as well as a function similar to `RecoverCluster()` that doesn't take a new configuration but rolls up the logs and takes a snapshot, writing that out in the latest version. That tool should allow for us to deprecate things by saying "you must run this on all your servers prior to running X (it could maybe be in `NewRaft()` as something we do automatically for folks).

config.go
slackpad: This ended up being nice! I like that we don't have to make a distinction between "old unversioned servers" and "new servers running in protocol 0 mode" anymore.

config.go
ongardie: > a function similar to RecoverCluster() that doesn't take a new configuration but rolls up the logs and takes a snapshot, writing that out in the latest version. 

That implicitly commits everything, so we can't do that safely.

> it could maybe be in NewRaft() as something we do automatically for folks

Maybe after an upgrade once your commit index has reached ???, you force a snapshot. Filling in the ??? is tricky, hmm.

slackpad: > That implicitly commits everything, so we can't do that safely.

Is that true even if we call it at startup during `NewRaft()`?

slackpad: Re-reviewed your Raft paper and I see why this is dangerous. After an outage we have to do the best we can because we don't have the right peers to determine if the last few logs were committed or not, so the best we can do is assume they were. We wouldn't want to do this process any other time.

I don't think it's going to be safe to ingest the peers.json initially like I'm doing now, though. It would be much better to just delete it.

slackpad: Fixed this over in Consul by adding some code to blow away the peers.json file on the first boot - https://github.com/hashicorp/consul/pull/2222/commits/82084c2a6c8a8e5a3dced3bcb71db4f0d4dba278.

slackpad: Er re-reading it again I think it's safe _at startup_ time because whoever becomes leader won't ever change their log and will enforce their log on the followers, and in this case it will have to happen with an install snapshot after a recovery.

I still like the change on the Consul side, though, because there really is no reason to run recovery on first startup. Going to sleep now!

slackpad: Sorry for my late night thrashing here - I edited the last few comments based on my current understanding.

slackpad: Tomorrow I'll take a shot at versioning snapshots and making things die for log entries that aren't understood - that'll make things more robust.

Still not sure what a great deprecation story is for when it's safe to delete code for handling old versions because of the uncertainty of snapshot timing. We could do something simple like run a cleanup goroutine off a snapshot request 15 minutes after a to-be-deprecated log entry gets committed (or if we see one in the log at startup).

This would only cause up to 4 snapshots per hour (or whatever) even if there were tons of these deprecated entries flying by, and would eliminate the deprecated entries within 15 minutes of finishing an upgrade once there were no servers producing the old entries any longer. The ongoing cost would be very low as long as the cleanup routine doesn't need to scan the logs; it can just select on a one-item buffered channel that gets pinged in a non-blocking fashion whenever we encounter something that needs cleanup.

slackpad: Adding the snapshot versioning uncovered an issue where we have dropped support for old peers in the snapshot (we can read it but not produce it) - we will need to write both in order to upgrade so that old servers can read new snapshots during the migration. I'll unpack this at the same time as snapshot versioning.

slackpad:
Ok I think I've got all the issues addressed. Unless there are any huge mistakes I'd like to merge this and fix things in separate PRs at this point. I was able to form a cluster with a legacy Consul 0.6.4 binary and a build using this and #143!

config.go
```diff
@@ -7,9 +7,121 @@ import (
...
+// 1: New format which adds support for a full configuration structure and its
+//    associated log index, with support for server IDs and non-voting server
+//    modes. To ease upgrades, this also includes the legacy peers structure but
+//    that will never be used by servers that understand version 1 snapshots.
+//    Since the original Raft library didn't enforce any versioning, we must
+//    include the legacy peers structure for this version, but we can deprecate
+//    it in the next snapshot version.
+const (
+	SnapshotVersionMin = 0
+	SnapshotVersionMax = 1
```
ongardie-sfdc: What controls which SnapshotVersion the server generates?

Right now it's set to SnapshotVersionMax, which would be problematic if it were non-backwards-compatible. For example, a new server could generate a new snapshot, send it to an old server, and that server would panic.

It seems like this should either be a function of the ProtocolVersion, or it should be specified separately in the config.

raft.go
```diff
@@ -20,6 +20,34 @@ var (
...
+// responses.
+func (r *Raft) getRPCHeader() RPCHeader {
+	return RPCHeader{
+		ProtocolVersion: r.conf.ProtocolVersion,
+	}
+}
+
+// dispositionRPC houses logic about whether this instance of Raft can process
+// an RPC message with the given header.
+func (r *Raft) dispositionRPC(header RPCHeader) bool {
```
ongardie-sfdc: What does the name "dispositionRPC" mean? How about `func (r* Raft) checkRPCHeader(header RPCHeader) error`, which we could later extend for cluster ID?

ongardie-sfdc:
@slackpad left you two more comments (1. SnapshotVersion, 2. dispositionRPC), then lgtm to merge. The issue of when is it safe to remove support for old log entry types and snapshot versions can be a separate PR.

config.go
slackpad: I'll add a function for now that takes the protocol version and returns the snapshot version to use so we don't hard code the max constant all over the place.

PR #143, opened 2016-07-28
Changes configuration API to a normal future and adds live bootstrap.

slackpad:
@ongardie there were two loose ends I needed to clean up. This is based off of the branch for #139 but I decided to make a separate PR for easier review.

1. It was bugging me that `GetConfiguration()` wasn't a normal `Future` since it can block, so I worked that to look like the other APIs.

2. In "bootstrap expect" mode, Consul creates a Raft object at startup and then bootstraps the cluster once the expected number of servers come online. I looked at ways to use the new `BootstrapCluster` to achieve this, but it was going to be super hard to do this correctly the way things are structured. I added a new live bootstrap capability that lets us to what we did before, but in a much safer way than the old method which just jammed some stuff in the peer store. This uses a new future/channel that the main thread consumes and makes an atomic decision to bootstrap or not based on whether there is any state present.

Added a few other small loose ends I ran into during integration.

### DIFF
--- a/api.go
+++ b/api.go
@@ -473,7 +473,8 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		}
 		r.processConfigurationLogEntry(&entry)
 	}
-	r.logger.Printf("[INFO] raft: Initial configurations: %+v", r.configurations)
+	r.logger.Printf("[INFO] raft: Initial configuration (index=%d): %+v",
+		r.configurations.latestIndex, r.configurations.latest)
 
 	// Setup a heartbeat fast-path to avoid head-of-line
 	// blocking where possible. It MUST be safe for this
--- a/api.go
+++ b/api.go
@@ -857,16 +857,18 @@ func (r *Raft) Stats() map[string]string {
 	lastLogIndex, lastLogTerm := r.getLastLog()
 	lastSnapIndex, lastSnapTerm := r.getLastSnapshot()
 	s := map[string]string{
-		"state":               r.getState().String(),
-		"term":                toString(r.getCurrentTerm()),
-		"last_log_index":      toString(lastLogIndex),
-		"last_log_term":       toString(lastLogTerm),
-		"commit_index":        toString(r.getCommitIndex()),
-		"applied_index":       toString(r.getLastApplied()),
-		"fsm_pending":         toString(uint64(len(r.fsmCommitCh))),
-		"last_snapshot_index": toString(lastSnapIndex),
-		"last_snapshot_term":  toString(lastSnapTerm),
-		"protocol_version":    toString(uint64(r.protocolVersion)),
+		"state":                r.getState().String(),
+		"term":                 toString(r.getCurrentTerm()),
+		"last_log_index":       toString(lastLogIndex),
+		"last_log_term":        toString(lastLogTerm),
+		"commit_index":         toString(r.getCommitIndex()),
+		"applied_index":        toString(r.getLastApplied()),
+		"fsm_pending":          toString(uint64(len(r.fsmCommitCh))),
+		"last_snapshot_index":  toString(lastSnapIndex),
+		"last_snapshot_term":   toString(lastSnapTerm),
+		"protocol_version":     toString(uint64(r.protocolVersion)),
+		"protocol_version_min": toString(uint64(ProtocolVersionMin)),
+		"protocol_version_max": toString(uint64(ProtocolVersionMax)),
 	}
 
 	future := r.GetConfiguration()
--- a/api.go
+++ b/api.go
@@ -144,6 +144,10 @@ type Raft struct {
 	// outside of the main thread.
 	configurationsCh chan *configurationsFuture
 
+	// bootstrapCh is used to attempt an initial bootstrap safely from
+	// outside of the main thread.
+	bootstrapCh chan *bootstrapFuture
+
 	// List of observers and the mutex that protects them. The observers list
 	// is indexed by an artificial ID which is used for deregistration.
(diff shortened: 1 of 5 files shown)


api.go
```diff
@@ -32,21 +32,17 @@ var (
 	// ErrEnqueueTimeout is returned when a command fails due to a timeout.
 	ErrEnqueueTimeout = errors.New("timed out enqueuing operation")
 
-	// ErrKnownPeer is returned when trying to add a peer to the configuration
```
slackpad: Realized we nuked these in a way that Consul needed. It wasn't a big deal to recover by checking the configuration beforehand, which is probably cleaner than looking for a specific error. Deleting these so people know this interface is gone.

slackpad: Only difference now is that if you call the peer change functions in a way that should be a no-op it'll make the changes anyway and run them through the Raft log. Consul would rely on this since it called these functions over and over again to reconcile things.

slackpad:
Rebased this off latest https://github.com/hashicorp/raft/pull/139 and force-pushed.

slackpad:
Some day I'd like to add a live version of `RecoverCluster()`, too. I think it would only work when Raft is in a new safe state but you could run a command on each server to kick it into a safe state and then recover it. This would let you fix things up without having to take the servers down which would be really nice.

ongardie:
> In "bootstrap expect" mode, Consul creates a Raft object at startup and then bootstraps the cluster once the expected number of servers come online.

Why does it create the Raft object ahead of time? Is that difficult to change?

ongardie:
> Some day I'd like to add a live version of RecoverCluster(), too.

So you don't even have to stand still, you can shoot yourself in the foot while you're running?

slackpad:
> So you don't even have to stand still, you can shoot yourself in the foot while you're running?

You'd still have to pause before you shoot :-) It's equivalent to what we do now in `RecoverCluster()` but you don't have to take you servers totally down.

> Why does it create the Raft object ahead of time? Is that difficult to change?

It used to be able to set the peers when it saw the right number of servers come up, so it always created the Raft object at startup. I went down the road of trying to change this but it got pretty ugly, and it would have needed some synchronization primitives around the Raft instance that would have invaded quite a few places. This was relatively simple on the Raft side, and it made integration into Consul much, much simpler. I'm not super thrilled with this interface, but it's not possible to use incorrectly so it doesn't seem like it will bite us. Also, you can be in bootstrap expect mode and then get added as a peer by someone else, do you do kind of want Raft to be running from the start, at least how Consul does things. We have a gossip layer that informs the Raft peers, so we are probably more dynamic than most.

api.go
```diff
@@ -140,6 +136,10 @@ type Raft struct {
 	// outside of the main thread.
 	configurationsCh chan *configurationsFuture
 
+	// bootstrapCh is used to attempt an initial bootstrap safely from
```
ongardie-sfdc: s/safely//

api.go
```diff
@@ -228,9 +228,15 @@ func RecoverCluster(conf *Config, fsm FSM, logs LogStore, stable StableStore,
...
 	}
 
-	// Sanity check the Raft peer configuration.
-	if err := checkConfiguration(configuration); err != nil {
-		return err
+	// Sanity check the Raft peer configuration. Note that it's ok to
+	// recover to an empty configuration if you want to keep the data
+	// but not allow a given server to participate any more (you'd have
+	// to recover again, or add it back into the existing quorum from
+	// another server to use this server).
```
ongardie-sfdc: Is this important to allow?

raft.go
```diff
@@ -197,6 +186,47 @@ func (r *Raft) runFollower() {
 	}
 }
 
+// liveBootstrap attempts to seed an initial configuration for the cluster. See
+// the Raft object's member BootstrapCluster for more details. This must only be
+// called on the main thread, and only makes sense in the follower state.
+func (r *Raft) liveBootstrap(configuration Configuration) error {
+	// Sanity check the Raft peer configuration.
```
ongardie-sfdc: A lot of this code is the same as non-live bootstrap. Is it better to call that here?

ongardie-sfdc:
after the above nits, lgtm

api.go
slackpad: Consul needed this - when you gracefully leave a cluster it used to blank out the peers. This isn't strictly necessary to do, but until we take the "go away" change later to this this to stop trying to elect, this server will be a nuisance in the Raft logs. This also seemed like a reasonable way to button up a server that was explicitly removed from a cluster with a final snapshot. I'd be open to removing this eventually, but it's good for Consul in the meantime, and seemed like a low impact on the interface.

ongardie-sfdc: I don't think that's safe to do. It'd be marking entries committed that may not have been, and the other servers may still think this one is part of the cluster.

In general, it's never safe to call RecoverCluster. When users invoke it manually, they should know that they're doing something risky and have signed the appropriate waivers. It's not cool to invoke automatically. It's not a hammer you can just pick up and use that way.

slackpad: @ongardie-sfdc agreed - I was just having misgivings about this as I was grabbing lunch and noticed your comment as I was jumping on to remove this :-)

I was thinking the user had signed up for this because they removed the server from the cluster, but if it comes back and somehow gets elected as leader it can cause damage. I'll remove this in a separate PR.

ongardie-sfdc: thanks

slackpad: Removed this in https://github.com/hashicorp/raft/pull/148.

ongardie-sfdc:
@slackpad, this one should be closed due to #147, right?

slackpad:
Correct - this made it in.

Issue #149, opened 2016-08-03
Are entry indices guaranteed to be contiguous and starting from 1?


barrucadu:
Looking at [`dispatchLogs`](https://github.com/hashicorp/raft/blob/37bf1111efb0c6ba96a9c0e92adb51c2f858996a/raft.go#L1138) it looks like log entries have contiguous indices and are numbered from 1. The benchmarks don't respect this: the benchmark for [`StoreLog`](https://github.com/hashicorp/raft/blob/37bf1111efb0c6ba96a9c0e92adb51c2f858996a/raft.go#L1138) uses 0-indexing, and for [`StoreLogs`](https://github.com/hashicorp/raft/blob/master/bench/bench.go#L76) uses 0-indexing and non-contiguous indices.

I noticed this because I'm writing a `LogStore` implementation which assumes indices are identical to database insertion order (using 1-indexing, and modulo rolling back recent entries), so [I added a check](https://github.com/barrucadu/logdb/commit/27555654d7ef5541c5796f20fbaeae298a0a8c76) to my `StoreLog` and `StoreLogs` methods to return an error if that's not the case.  I then had to [revert that check](https://github.com/barrucadu/logdb/commit/e3fa6abe30a1aaea18c0eda0a07ec809a239f3a3) to make the benchmarks run, which isn't ideal.

Before I start blaming the benchmarks, I'd first like to check that I'm not misunderstanding something, and these cases could arise in normal execution. And if those cases can't arise normally, would you be open to a patch to the benchmarks bringing the behaviour more in-line with the main hashicorp/raft library?

ongardie:
The indexing might start past 1 after log compaction, but there should never ever be an entry 0 in Raft.

barrucadu:
Starting after 1 after compaction is fine, as long as originally there was a 1 inserted.

ongardie:
If this is a new server added after the others have compacted and removed entry 1, then this server will never see entry 1.

barrucadu:
Ah, of course. So I do need to keep track of the offset between "insertion order IDs" and log indices. And to make the `DeleteRange` benchmark work, I also need to keep track of the discontinuities.

Thanks.

ongardie:
> And to make the DeleteRange benchmark work, I also need to keep track of the discontinuities.

Discontinuities? Raft doesn't allow gaps.

slackpad:
It looks like the benchmarks are doing bad things. We should make them more realistic.

[...]

PR #150, opened 2016-08-09
Make benchmarks only perform Raft-permitted actions

barrucadu:
See also #149.

I assume this never came to light before because both raft-boltdb and raft-mdb use a key-value store as the underlying database, so there is no inherent relation between Raft indices and anything in the database. It becomes an issue when implementing an actual log database, where database IDs correspond exactly with insertion order, so it's desirable for the Raft indices to be a simple function of database IDs.


s-christoff:
Howdy @barrucadu ,
Thank you so much for this contribution! We would be happy to merge this and our apologies it has taken us so long to review. Would you mind adding some comments to this to explain why these numbers are important? We looked at this for awhile before we could understand it's usefulness. If not, we can do it! 😄

PR #153, opened 2016-08-31
break groupcommit if there's no pending commit log

weiwei04:


ongardie:
Ha, nice. I didn't realize that break statement was only breaking out of the select before.

ongardie-sfdc:
> A "break" statement terminates execution of the innermost "for", "switch", or "select" statement within the same function.  -https://golang.org/ref/spec#Break_statements

@slackpad, can we get this into v2, at least?

banks:
Hey @weiwei04, sorry it's been so long but thanks for this contribution!

I've actually made a separate PR #341 just because there is a slightly cleaner (arguably) way to do this in Go but we really appreciate this!

PR #159, opened 2016-09-28
Adds ability to access user snapshots and restore them.

slackpad:
This leverages Raft's existing snapshot plumbing in a way that can be used to store snapshots externally and later restore them, such as for a disaster recovery scenario.

1. The future returned by `Snapshot()` has been extended with an `Open()` method that allows a snapshot to be saved externally.

2. A new `Restore()` method has been added to the the API to take an external snapshot and apply it to the state of the cluster.

The restore can only be run on the leader, and works by making a new local snapshot. This snapshot uses the latest configuration, not the one coming from the external snapshot (it is assumed that the cluster has been bootstrapped already and we don't want the old configuration). The index is also set to one past the last index, or the index in the snapshot, whichever is greater. This means that the post-restore index still makes sense with the contents of the snapshot because the index will be moved forward to accommodate its contents if needed. We add one to create a "hole" in the leader's log which will cause the replication threads to fault and send the snapshot.

The leader applies this snapshot to its FSM and updates state, and the replication threads take care of the sending the snapshot because of the hole in the leader's log.

We commit a no-op after the snapshot restore in order to provide a mechanism that tells the caller when the snapshot has been consumed by a quorum of servers, since they must have taken the snapshot in order to have committed an index that follows the snapshot.

Targeting this to library-v2-stage-one first and will forward-port it to issue-84-integration in a separate PR.

/cc @armon @ongardie


api.go
```diff
@@ -792,18 +798,78 @@ func (r *Raft) Shutdown() Future {
...
 	case <-r.shutdownCh:
 		return errorFuture{ErrRaftShutdown}
+	case r.userRestoreCh <- restore:
+		// If the restore is ingested then wait for it to complete.
+		if err := restore.Error(); err != nil {
+			return restore
+		}
 	}
 
+	// Apply a no-op log entry. Waiting for this allows us to wait until the
```
armon: I wonder if this should be done by the caller? e.g. the semantics are that Restore just applies to the leader, and the caller can use a `Barrier` or something to ensure it has been applied.

future.go
```diff
@@ -46,6 +48,15 @@ type ConfigurationFuture interface {
 	Configuration() Configuration
 }
 
+// SnapshotFuture is used for waiting on a user-triggered snapshot to complete.
+type SnapshotFuture interface {
+	Future
+
+	// Open is a function you can call to access the underlying snapshot and
+	// its metadata.
+	Open() (*SnapshotMeta, io.ReadCloser, error)
```
armon: Is it only safe to call this once we get a non-nil Err back? Good to just add to the comment.

future.go
```diff
@@ -150,9 +161,36 @@ func (s *shutdownFuture) Error() error {
...
+	opener func() (*SnapshotMeta, io.ReadCloser, error)
+}
+
+// Open is a function you can call to access the underlying snapshot and its
+// metadata.
+func (u *userSnapshotFuture) Open() (*SnapshotMeta, io.ReadCloser, error) {
+	if u.opener == nil {
+		return nil, nil, fmt.Errorf("no snapshot available")
+	} else {
+		return u.opener()
```
armon: Looks like it's probably unsafe to call `Open` more than once, might be good to nil out the Opener after

raft.go
```diff
@@ -680,6 +691,94 @@ func (r *Raft) quorumSize() int {
...
 
+// restoreUserSnapshot is used to manually consume an external snapshot, such
+// as if restoring from a backup. We will use the current Raft configuration,
+// not the one from the snapshot, so that we can restore into a new cluster. We
+// will also use the higher of the index of the snapshot, or the current index,
+// and then add 1 to that, so we force a new state with a hole in the Raft log,
+// so that the snapshot will be sent to followers and used for any new joiners.
+// This can only be run on the leader, and returns a future that can be used to
+// block until complete.
+func (r *Raft) restoreUserSnapshot(future *userRestoreFuture) {
```
armon: Seems cleaner to have this function return an error and have the caller respond to the future. It's odd to have everything inside the nested anonymous function call for this.

raft.go
```diff
@@ -680,6 +691,94 @@ func (r *Raft) quorumSize() int {
...
+		term := r.getCurrentTerm()
+		lastIndex := r.getLastIndex()
+		if future.meta.Index > lastIndex {
+			lastIndex = future.meta.Index
+		}
+		lastIndex++
+
+		// Dump the snapshot. Note that we use the latest configuration,
+		// not the one that came with the snapshot.
+		sink, err := r.snapshots.Create(version, lastIndex, term,
```
armon: Should we truncate the logs prior to the snapshot as well? Or just wait for that to be truncated later?

armon: Sorry, after the snapshot, not prior.

raft.go
```diff
@@ -680,6 +691,94 @@ func (r *Raft) quorumSize() int {
...
+			return fmt.Errorf("failed to close snapshot: %v", err)
+		}
+		r.logger.Printf("[INFO] raft: Copied %d bytes to local snapshot", n)
+
+		// Restore the snapshot into the FSM. If this fails we are in a
+		// bad state so we panic to take ourselves out.
+		fsm := &restoreFuture{ID: sink.ID()}
+		fsm.init()
+		select {
+		case r.fsmRestoreCh <- fsm:
```
armon: An interesting problem with the fsmRestoreCh is that we can't promise an ordering with `r.fsmCommitCh`. This could cause issues if we restore and then older commits are applied on top of it. I think we may need to fix the `runFSM` to use a single fixed FIFO queue so that we have reliable ordering.

api.go
slackpad: Interesting - I originally had this down in the `Restore()` itself and moved it up here since it is kind of an extra thing. I think from an API perspective it's pretty nice to have a future that tells you the restore is complete all the way through, especially given that this is slightly sketchy in making the leader take on the state and trigger for it to get sent to the followers.

raft.go
slackpad: Ah yeah that'll clean this up.

raft.go
slackpad: I thought about that and I think it would be fine to truncate but it seemed to add extra complexity without too much benefit. That will happen at the next snapshot interval.

raft.go
slackpad: Good catch - I'll take a look at that.

slackpad: Since the main thread is the producer and the FSM thread is the consumer we can safely poll until that commits channel is drained. This might be a good step now to capture the need to do this inside of one function, as the structure will change a bit when we integrate this with the changes in issue-84-integration.

slackpad: Hmm maybe the single FIFO is cleaner - I'll dig on this a bit.

slackpad: (previous comment was so unsafe and wrong i deleted it :-)

slackpad:
@armon ok responded to everything and pushed up changes - PTAL

slackpad:
Got out-of-band LGTM.

PR #175, opened 2016-10-07
Add new types for Index and Term

### LINKED ISSUE #171: [v2] Add new types for Index and Term
They're presently just uint64.

ongardie-sfdc:
Close #171.


ongardie-sfdc:
That wasn't nearly as bad as I expected it to be. I guess `:=` helps.

Issue #172, opened 2016-10-07
[v2] Merge state.go into raft.go?


ongardie:
I'll volunteer to start this off.

ongardie-sfdc:
I'll hold off on physically moving code still, but I'd like to end up with:
- raft.go consisting of the public-facing API exported by the library (mostly in api.go today)
- server.go consisting of the `raftServer` struct and its methods (mostly in raft.go today)
- No state.go (folded into `raftServer`)
- No other files accessing `Raft` or `raftServer` (fsm.go, observer.go, snapshot.go, state.go do this today)

A few questions:

The functions `BootstrapCluster()`, `RecoverCluster()`, and `HasExistingState()` are a little unique, in that they're part of the public API but are most similar to code in `raftServer`. Should we move them to a new file like "offline.go"?

What is observer.go for? If it's just for unit tests, can we hide it from the public API?

One ugly bit right now is that a server can be a Follower, Candidate, Leader, or Shutdown, and shutting down sets the state.go state of the server to Shutdown forcefully from api.go. Any objection to removing Shutdown from the RaftState enum? A server that shut down would just stay in the Follower state. This would affect and possibly break some applications that call Raft.State() and compare it to things w.r.t. Shutdown.

Should we remove State(), LastIndex(), LastContact() from the API? Clients can get this information from Stats() instead. Alternatively, we could rewrite them in terms of Stats(). And what about Leader()? It's not in Stats today but probably should be.

@slackpad @superfell

slackpad:
> The functions BootstrapCluster(), RecoverCluster(), and HasExistingState() are a little unique, in that they're part of the public API but are most similar to code in raftServer. Should we move them to a new file like "offline.go"?

I like offline.go for that - those are special and it's nice to make them clear they aren't for a running cluster.

> What is observer.go for? If it's just for unit tests, can we hide it from the public API?

This is only for unit tests so let's definitely hide it. We also don't want external users waiting on leader changes using this, as they are probably doing something racy and unreliable. Given how state has been reorganized we should also see if we can get rid of this - it's mostly used in unit tests to wait for things so settle.

> One ugly bit right now is that a server can be a Follower, Candidate, Leader, or Shutdown, and shutting down sets the state.go state of the server to Shutdown forcefully from api.go. Any objection to removing Shutdown from the RaftState enum?

This sounds ok - I don't know of any places that look at the state for shutdown in our stuff. What's the reason not to have this as a special state?

[...]

Issue #167, opened 2016-10-07
Make the stats into a real structure


slackpad:
Clean up how configuration comes out in `Stats()`, the `"%+v"` formatting is weird / ugly. Also add a test to make sure it shows up (fixed a bug where the logic was wrong).

Better yet, make a new stats API that pulls the structure in one go w/o text.

ongardie:
Another thing I wanted to do was to get at the term from my application (to ensure it hadn't changed across a period of time for reasons I can explain in gory detail). Should that be a new getTerm() or should this cover it?

slackpad:
I think you'd get the term in this stats struct.

ongardie-sfdc:
I think we should make getting this stats struct a proper query into the main raft.go goroutine over a channel. This would help a lot with #172 (merging state.go into raft.go).

slackpad:
Yeah a future, similar to what we did for the configuration getter would be good for this as well.

ongardie-sfdc:
btw I'm starting on this one.

slackpad:
@ongardie-sfdc this one can be closed I think, right?

ongardie:
@slackpad oh, I thought so but now I see the comment:

> Clean up how configuration comes out in Stats(), the "%+v" formatting is weird / ugly. Also add a test to make sure it shows up (fixed a bug where the logic was wrong).

Better check on that before we close it out.

ongardie-sfdc:
PR #191 finishes this up.

PR #176, opened 2016-10-07
Start to merge state.go and raft.go

ongardie-sfdc:
This is atop #175.

c176f78 is a good start to #172 that'd be nice to merge now to avoid conflicts later. But I think most of #172 will have to wait until #167.


ongardie-sfdc:
rebased since #175 was merged

ongardie-sfdc:
nvm, starting new PR

Issue #165, opened 2016-10-07
Rename "Configuration" to "Membership"


slackpad:
This'll stop conflating it with `Config`.

ongardie:
Also, add a note to type configurations that latest and committed may be the same, and often are.

superfell:
I'll take this one

PR #177, opened 2016-10-08
Configuration ->Membership rename

superfell:
This address issue #165 

The one place i didn't change is the field names in SnapshotMeta as that'll affect reading existing snapshots (generated by consul 0.7?)

While trawling through this, i also made a number of minor tweaks checkConfiguration() is now a member of the Membership type, i cleaned up some repeated code in the tests.


superfell:
#178 fixes the travis failures which are unrelated to this change.

ongardie:
> The one place i didn't change is the field names in SnapshotMeta as that'll affect reading existing snapshots (generated by consul 0.7?)

Can an annotation on the struct field be used to override the name for msgpack (like people do with json)?

superfell:
In the library this is only serialized to/from json, and we can annotate it to handle that. However snapshot store is externally pluggable, and its not clear to what degree that is used, and if it should be handled in a serialization format agnostic fashion.

superfell:
@slackpad does Consul use the file_snapshot store from here, or does it have its own?

slackpad:
@superfell Consul uses the file snapshot store directly from the Raft repo and I think the store is responsible for serialization of the metadata. If it's possible use annotations to read it back from either key name in the JSON then I think that would work.

superfell:
I added the rename of the SnapshotMeta fields, it uses a json tag to preserve the previous field names in the serialized version [file_snapshot serializes this to/from json]

slackpad:
Thanks guys and sorry for the delay on this one!

Issue #141, opened 2016-07-24
Take an interface on WithLogger rather than *log.Logger


connor4312:
It'd be nice if an interface was accepted rather than requiring a *log.Logger struct so that logging could be further customized and swapped out other other systems (for instance, we use Logrus with a Sentry transport for centralized logging).

I'd be happy to put in a PR for this if there's interest.

superfell:
This would be a good candidate to include for v2 @slackpad @ongardie

ongardie:
@slackpad @superfell close-able now?

slackpad:
I was going to leave this open until I switch over to logxi (see https://github.com/hashicorp/raft/pull/180#issuecomment-256726944). Working on that PR now.

connor4312:
Thanks everyone 😄  Close at-will

PR #180, opened 2016-10-11
Abstract Logging behind an interface

superfell:
Which’ll make it easier for consumers of the library to plug in their
logging framework of choice. Addresses issue #141

### DIFF
--- a/api.go
+++ b/api.go
@@ -3,7 +3,6 @@ package raft
 import (
 	"errors"
 	"fmt"
-	"log"
 	"os"
 	"strconv"
 	"sync"
@@ -98,7 +97,7 @@ type Raft struct {
 	localAddr ServerAddress
 
 	// Used for our logging
-	logger *log.Logger
+	logger Logger
 
 	// LogStore provides durable storage for logs
 	logs LogStore
@@ -388,14 +387,14 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 	}
 
 	// Ensure we have a LogOutput.
-	var logger *log.Logger
+	var logger Logger
 	if conf.Logger != nil {
 		logger = conf.Logger
 	} else {
 		if conf.LogOutput == nil {
 			conf.LogOutput = os.Stderr
 		}
-		logger = log.New(conf.LogOutput, "", log.LstdFlags)
+		logger = DefaultStdLogger(conf.LogOutput)
 	}
 
 	// Try to restore the current term.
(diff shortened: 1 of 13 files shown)


superfell:
In a number of places this preserves the original option to provide just a writer for logging, I'm not sure if that's still need. @slackpad do you need that path for Consul? or can we remove those paths and just have the logger interface versions ?

ongardie-sfdc:
Can we use this to remove the "raft: " prefix from every single log line?

superfell:
I was going to, but there are some log lines with different prefixes

ongardie-sfdc:
Ah. My grep returns "[ERR] snapshot:" and "[ERR] raft-net:". Not sure if that's particularly intentional or important to maintain.

superfell:
I can remove the prefixes if no one has a strong objection.

superfell:
This does have an issue that @ongardie-sfdc and I discussed which is if you supply a log.Logger with the Lshortfile or Llongfile flags turned on, its just going to show the relevant line in logger.go

superfell:
the other approach would to level methods that returns a Logger that a log.Logger directly implements, 

type Logger interface{
    Printf(format string, values ...interface{})
}

type Loggers interface{
   Debug() Logger
   Info() Logger
   Warn() Logger
   Error() Logger
}
or 
type Loggers interface{
   Log(Level) Logger
}

ongardie-sfdc:
How about https://github.com/ongardie/hashicorp-raft/commit/4c7d4284b4aa3fe77faf9efce9f8323c2b8e2be6#diff-40646a6e1108f49d452915ba3515c206 instead?

tl/dr:
````
type Logger func(calldepth int, level LogLevel, s string)
func NewStdLogger(l *log.Logger, levelToLog LogLevel) Logger {
	return func(calldepth int, level LogLevel, s string) {
		if level >= levelToLog {
			l.Output(calldepth+3, fmt.Sprintf("[%v] %s", level, s))
		}
	}
}
```

As a concrete type, we can hang various methods off of it, including Debug() Info() Warn() Error() and Panic().

This also adds a unit test to make sure we're getting the filename handling right.

ongardie-sfdc:
🌮

superfell:
Is that a taco ?

ongardie-sfdc:
That's a :taco, yes. I'm not allowed to use 🎉  but tacos are like a celebration.

ongardie-sfdc:
@superfell, I fixed the few trivial conflicts over on https://github.com/ongardie/hashicorp-raft/commits/logging_itf, please update to that

ongardie-sfdc:
@superfell I added one more commit for unit+integ tests over on https://github.com/ongardie/hashicorp-raft/commits/logging_itf

slackpad:
Sorry to come late to this one. I realized we have some precedent for logxi in our other tools, though that's not official - https://github.com/hashicorp/vault/blob/master/helper/logformat/vault.go. I'm going to go ahead and merge this to unblock other changes but then I'll make another pass through and convert to the logxi v1 interface. It's similar to this interface, with the exception of passing kv pairs in vs. a fmt string, so should be a pretty simple conversion.

PR #184, opened 2016-10-14
refactors for stats, raft.go and api.go

ongardie-sfdc:
This is stacked on top of #180. It's a WIP but should be safe to merge, and it might be good to merge before it's done as it'll conflict with things.

Completes #167 and starts on #172.


ongardie-sfdc:
(rebased)

PR #187, opened 2016-10-28
Converts logger interface to logxi.

### LINKED ISSUE #141: Take an interface on WithLogger rather than *log.Logger

slackpad:
There's some precedent in other HC products to use logxi v1 (even though that's not yet an official thing). This switches to that instead of making a custom log interface specific to Raft.

Closes #141.


slackpad:
@superfell and @ongardie-sfdc this switches to the logxi interface.

logger.go
```diff
@@ -1,6 +1,222 @@
...
+
+	switch level {
+	case log.LevelCritical:
+		writer.Write([]byte(" [CRIT ] "))
+	case log.LevelError:
+		writer.Write([]byte(" [ERROR] "))
+	case log.LevelWarn:
+		writer.Write([]byte(" [WARN ] "))
+	case log.LevelInfo:
+		writer.Write([]byte(" [INFO ] "))
```
ongardie-sfdc: I don't want to get into a flamewar, but I find this a bit jarring. Here's an example from the unit tests:
```
(server-5a6f) peer.go:607 2016/10/28 11:26:01.944915 [INFO ] Starting heartbeat RPC for peer: id=server-e71d address=e71d
(server-5a6f) peer.go:1042 2016/10/28 11:26:01.944985 [INFO ] Not updating next index for peer, already set: id=server-8f3b address=8f3b next_index=3
```

First, the beginning of "[INFO" doesn't line up. Then, even if it did, I think it'd look nicer as [INFO] followed by the necessary spacing after the close bracket.

And another nit, I'd prefer the date on the far left, not the filename. That makes it easier to sort by time if you're catting files together or something.

slackpad: Not flamewar material at all! I was curious what you'd think and agree about the `[ ]` stuff not lining up. Space inside of the brackets vs. outside we've internally talked about and settled on this, but I'll move the other stuff right so the brackets line up. The jarring parts of this only apply to unit tests, but that's what we spend a lot of time on, so worth getting nice to look at.

slackpad:
Latest update is easier to scan:

```
=== RUN   TestRaft_SingleNode
2016/10/28 15:05:47.639707 [DEBUG] (cluster) raft_test.go:377 Fully Connecting
2016/10/28 15:05:47.640145 [INFO ] (server-70de) api.go:532 Initial membership: index=1 servers="[{Voter server-70de 70de}]"
2016/10/28 15:05:47.640278 [INFO ] (server-70de) raft.go:156 Entering Follower state: leader=""
2016/10/28 15:05:47.711952 [WARN ] (server-70de) raft.go:215 Heartbeat timeout reached, starting election: last_leader="" term=1 last_contact="0001-01-01 00:00:00 +0000 UTC"
2016/10/28 15:05:47.712061 [INFO ] (server-70de) raft.go:259 Entering Candidate state: term=2
2016/10/28 15:05:47.712144 [INFO ] (server-70de) raft.go:695 Election won: tally=1
2016/10/28 15:05:47.712165 [INFO ] (server-70de) raft.go:327 Entering Leader state
--- PASS: TestRaft_SingleNode (0.07s)
```

ongardie-sfdc:
thanks

PR #189, opened 2016-10-29
Added an in-mem snapshot store

kyhavlov:
Added an in-memory snapshot store that keeps the most recent snapshot for opening, to support hashicorp/consul#2439.

### DIFF
--- a/inmem_snapshot.go
+++ b/inmem_snapshot.go
@@ -2,6 +2,7 @@ package raft
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 )
@@ -15,14 +16,14 @@ type InmemSnapshotStore struct {
 // InmemSnapshotSink implements SnapshotSink in memory
 type InmemSnapshotSink struct {
 	meta     SnapshotMeta
-	contents []byte
+	contents *bytes.Buffer
 }
 
 // NewInmemSnapshotStore creates a blank new InmemSnapshotStore
 func NewInmemSnapshotStore() *InmemSnapshotStore {
 	return &InmemSnapshotStore{
 		latest: &InmemSnapshotSink{
-			contents: []byte{},
+			contents: &bytes.Buffer{},
 		},
 	}
 }
@@ -32,11 +33,13 @@ func (m *InmemSnapshotStore) Create(index, term uint64, peers []byte) (SnapshotS
 	name := snapshotName(term, index)
 
 	sink := m.latest
-	sink.meta.ID = name
-	sink.meta.Index = index
-	sink.meta.Term = term
-	sink.meta.Peers = peers
-	sink.contents = []byte{}
+	sink.meta = SnapshotMeta{
+		ID:    name,
+		Index: index,
+		Term:  term,
+		Peers: peers,
+	}
+	sink.contents = &bytes.Buffer{}
 
 	return sink, nil
 }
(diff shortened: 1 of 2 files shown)


inmem_snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+type InmemSnapshotSink struct {
+	meta     SnapshotMeta
+	contents []byte
+}
+
+// NewInmemSnapshotStore creates a blank new InmemSnapshotStore
+func NewInmemSnapshotStore() *InmemSnapshotStore {
+	return &InmemSnapshotStore{
+		latest: &InmemSnapshotSink{
+			contents: []byte{},
```
slackpad: You could drop this - the zero value for a slice is safe.

inmem_snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// List returns the latest snapshot taken
+func (m *InmemSnapshotStore) List() ([]*SnapshotMeta, error) {
+	return []*SnapshotMeta{&m.latest.meta}, nil
+}
+
+// Open wraps an io.ReadCloser around the snapshot contents
+func (m *InmemSnapshotStore) Open(id string) (*SnapshotMeta, io.ReadCloser, error) {
+	return &m.latest.meta, ioutil.NopCloser(bytes.NewReader(m.latest.contents)), nil
```
slackpad: I'd bail with an error if the ID doesn't match what's in the meta, otherwise this might cause some weird errors that will be hard to diagnose (if someone holds onto an ID and then another snapshot takes place).

inmem_snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// Write appends the given bytes to the snapshot contents
+func (s *InmemSnapshotSink) Write(p []byte) (n int, err error) {
+	s.contents = append(s.contents, p...)
```
slackpad: If you use a `bytes.Buffer` for storage this could be a return of an `io.Copy()` call. That might be more efficient if the snapshot is kind of large as well (kind of a guess, not 100% sure) since it seems like it could do larger blocks for copying.

inmem_snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+	return len(p), nil
+}
+
+// Close updates the Size and is otherwise a no-op
+func (s *InmemSnapshotSink) Close() error {
+	s.meta.Size = int64(len(s.contents))
```
slackpad: I'd update this as the writes occur.

inmem_snapshot.go
kyhavlov: Is there a case where Open() could be called multiple times on the same snapshot, or is it fine to just return the bytes.Buffer directly for reading here?

inmem_snapshot.go
```diff
@@ -0,0 +1,79 @@
...
+	return []*SnapshotMeta{&m.latest.meta}, nil
+}
+
+// Open wraps an io.ReadCloser around the snapshot contents
+func (m *InmemSnapshotStore) Open(id string) (*SnapshotMeta, io.ReadCloser, error) {
+	if m.latest.meta.ID != id {
+		return nil, nil, fmt.Errorf("[ERR] snapshot: failed to open snapshot id: %s", id)
+	}
+
+	return &m.latest.meta, ioutil.NopCloser(m.latest.contents), nil
```
kyhavlov: Is there a case where Open() could be called multiple times on the same snapshot, or is it fine to just return the bytes.Buffer directly for reading here?

kyhavlov:
Updated to reflect merging into library-v2-stage-one

inmem_snapshot.go
slackpad: I think this should be ok since it's not well-defined that you can open multiple times.

PR #191, opened 2016-10-31
Bug fixes + issue 167

ongardie-sfdc:


logger.go
```diff
@@ -130,15 +130,25 @@ func (v *raftFormatter) formatDefault(writer io.Writer, level int, msg string, a
...
+	// This is a little jank, but it's only used during unit tests.
 	if v.withLine {
-		_, file, line, ok := runtime.Caller(4)
-		if !ok {
-			file = "???"
-			line = 0
+		file := "???"
+		line := 0
+		depth := 0
+		for {
```
slackpad: I thought about doing this and hoped we didn't have to, but didn't catch the `Error()` case. Should be fine for tests, I agree.

ongardie-sfdc:
This String() function broke the unit tests, working on fixing them.

PR #194, opened 2016-11-10
Use an RWMutex for inmem snapshot store

kyhavlov:
Fixes race errors identified in https://github.com/hashicorp/consul/issues/2496


slackpad:
LGTM

PR #203, opened 2017-03-16
Fix last_contact case for leader in Stats()

kyhavlov:


slackpad:
LGTM (good to check CI before merge - there could be a test that's affected).

PR #207, opened 2017-05-04
Provides a JSON format for recovery when using server IDs.

slackpad:
We need this for outage recovery when using server IDs, see https://github.com/hashicorp/consul/issues/3003.


slackpad:
out of band lgtm from @kyhavlov

PR #210, opened 2017-05-17
Allow streaming of snapshots without a size known in advance

JelteF:
Right now snapshotting is heavily bound to sending a single byte stream of which the size is known in advance, such as sending a file stored on disk. However, in some cases the size is not known. For instance when gzipping the data that is being sent, or as in my case when sending a directory that is being tarred and untarred on the fly. This allows specifying a size of `-1` to indicate that the size should not be checked by raft and that other mechanisms for integrity are used.

It also changes the net_transport to support this automatically for half closable connections, such as TCP. Closing of the write half of the connection by the sender will be used to indicate the end of the snapshot and a success response can still be sent back by the receiver.

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -87,6 +87,11 @@ type StreamLayer interface {
 	Dial(address string, timeout time.Duration) (net.Conn, error)
 }
 
+type HalfClosableConn interface {
+	net.Conn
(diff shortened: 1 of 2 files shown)


JelteF:
Is there any problem in merging this?

PR #214, opened 2017-06-01
Correct docs on value of state Stat

JelteF:
It's not true that "state" contains a numerical value. It contains the string representation of the RaftState.


mkeeler:
@JelteF If you rebase this to master I can go ahead and merge it.

s-christoff:
Hey @JelteF, thank you so much for your contribution! We went ahead and made this change in #331 since we really wanted to add it, but it's been awhile since we heard from you.

Issue #218, opened 2017-06-21
Suggestion: control over leader identity


shlomi-noach:
I find it convenient that I could have control over the identity of the leader in a stable group. For example, I may be running cross-DC nodes (in an attempt to fight fencing), and would _prefer_ the leader to normally run from a specific, local, "active" DC to where most operations are running.

Or, as another example, I might have a 3-node group, but only 2 are full blown provisioned machines, and the 3rd is only a raft mediator.

Ideally, I would like to be able to point to a specific node and ask "become the leader". I can see various ways to achieve this:

- ask a leader to step down (perhaps re-elections would pick it again, perhaps not)
- configure a/some nodes to never even attempt entering `Candidate` state (up to `quorum - 1` nodes are safe to run this)
- instruct a/some nodes to refrain from entering `Candidate` state for a bound period of time

I will try some pull-requests to support these ideas. I'm aware that my understanding of Raft is limited, and will work with what makes sense to me following existing use patterns.

Related: https://github.com/hashicorp/raft/issues/211

shlomi-noach:
I'm happy to receive any feedback on the idea in general and on the proposed PRs in particular. Thank you!

shlomi-noach:
#222 seems to be working very well on my environment.

superfell:
Have you seen the v2 branch which has support for non-voting members? it might be a better fit for the problem you're trying to solve.

shlomi-noach:
@superfell thank you, I was not aware of `v2`. Some notes and questions, if you please:

0. [AddNonVoter()](https://github.com/hashicorp/raft/blob/85a7a8b892171d6f115c49b7bd4e7f0202c6cb2f/raft.go#L314) seems to deprecate the idea behind #220 
0. If I run [DemoteVoter()](https://github.com/hashicorp/raft/blob/85a7a8b892171d6f115c49b7bd4e7f0202c6cb2f/raft.go#L347) with the leader's ID, 
  - does the leader step down? (a-la #219)
  - Or will it keep leading, and will not participate in next elections?
0. I do not see a `BecomeVoter()` or similar, the anti-operation of `DemoteVoter()` -- is it possible to turn a non-voter to be a voter?

My current implementation/use allows for something like: `https://myservice/api/raft-yield/1.2.3.4:10008`

[...]

PR #219, opened 2017-06-21
StepDown(): requesting reelections

shlomi-noach:

With this PR the `StepDown()` function is available, that forces a leader back into `Follower` state, and effectively calling for reelections.


shlomi-noach:
#222 extends this PR

shlomi-noach:
https://github.com/hashicorp/raft/pull/222 supersedes this PR and makes for a production-tested solution.

PR #220, opened 2017-06-22
NeverLeader config: node will never switch to Candidate 

shlomi-noach:

This PR introduces the `NeverLeader` (`bool`) config. A node where `NeverLeader` is `true` never changes state from `Follower` to `Candidate`. Hence, it will never become a leader.

Purpose: we may have a setup where some raft nodes are mostly acting as mediator (tie breakers) but are themselves located in a remote datacenter with some latencies; we may look for the leader to be located closer to the rest of our operation.

**Danger**: you may only apply this onto (quorum-1) nodes, or else your group may be incapable of electing a leader.

- In a 3-node group, up to 1 node can be set with `NeverLeader`
  - same for 4-node group
- In a 5-node group, up to 2 nodes can be set with `NeverLeader`
  - same for 6-node group
- In a 7-node group, up to 3 nodes can be set with `NeverLeader`
  - same for 8-node group

### DIFF
--- a/config.go
+++ b/config.go
@@ -72,6 +72,12 @@ type Config struct {
 	// never be used except for testing purposes, as it can cause a split-brain.
 	StartAsLeader bool
 
+	// NeverLeader instructs this Raft node to never become the leader. This
+	// configuration may be applied on up-to quorum-1 nodes.
+	// For example, on a 3-node cluster, quorum is 2 and NeverLeader may only be applied on 1 node.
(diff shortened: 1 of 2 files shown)


preetapan:
Closing this out because it duplicates what AddNonVoter/DemoteVoter/AddVoter already do. See https://github.com/hashicorp/raft/issues/218#issuecomment-332941998 for more context

PR #222, opened 2017-07-02
Yield()

shlomi-noach:

This combines https://github.com/hashicorp/raft/issues/219 and https://github.com/hashicorp/raft/pull/221

The `Yield()` function disables promotion for the running node, and in the event it is the leader, steps down.

This makes for a nicer flow than suggested by #220 or #221 thus closes them.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/armon/go-metrics"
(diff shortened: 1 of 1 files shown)


shlomi-noach:
I should note this seems to be working very well on my environment. I left it for the app to determine who to yield to, and for the app to advertise the need for yielding. I would love `yield` to be part of the gossip, but do not expect as much.

I would love your thoughts/review.

tylertreat:
FWIW, I would like to see this (particularly `StepDown`) as I have a use case for it. https://github.com/hashicorp/raft/issues/239

shlomi-noach:
@tylertreat while I patiently wait for a review/thoughts on merging this into `hashicorp/raft`, please note that I've forked into https://github.com/openark/raft, where this PR is merged.

shlomi-noach:
As per https://github.com/hashicorp/raft/issues/218#issuecomment-332941998 there is no intention to merge this PR. Closing.

PR #224, opened 2017-07-06
Delete temporary snapshot files. 

preetapan:
A minor but useful thing to do. When a snapshot save fails for any reason, we leave a tmp file hanging around. This patch deletes the temp file the next time we attempt a snapshot.

### DIFF
--- a/file_snapshot.go
+++ b/file_snapshot.go
@@ -240,11 +240,7 @@ func (f *FileSnapshotStore) getSnapshots() ([]*fileSnapshotMeta, error) {
 		// Ignore any temporary snapshots
 		dirName := snap.Name()
 		if strings.HasSuffix(dirName, tmpSuffix) {
-			f.logger.Printf("[WARN] snapshot: Found temporary snapshot: %v, deleting..", dirName)
-			path := filepath.Join(f.path, dirName)
-			if err := os.RemoveAll(path); err != nil {
-				f.logger.Printf("[ERR] snapshot: Failed to delete temporary snapshot %v: %v", path, err)
-			}
+			f.logger.Printf("[WARN] snapshot: Found temporary snapshot: %v", dirName)
 			continue
 		}
 
@@ -388,6 +384,11 @@ func (s *FileSnapshotSink) Close() error {
 	// Close the open handles
 	if err := s.finalize(); err != nil {
 		s.logger.Printf("[ERR] snapshot: Failed to finalize snapshot: %v", err)
+		s.logger.Printf("[INFO] snapshot: Deleting temporary files...")
+		if delErr := os.RemoveAll(s.dir); delErr != nil {
+			s.logger.Printf("[ERR] snapshot: Failed to delete temporary snapshot at path %v: %v", s.dir, err)
+			return delErr
+		}
 		return err
 	}
 
--- a/file_snapshot.go
+++ b/file_snapshot.go
@@ -240,7 +240,11 @@ func (f *FileSnapshotStore) getSnapshots() ([]*fileSnapshotMeta, error) {
 		// Ignore any temporary snapshots
 		dirName := snap.Name()
 		if strings.HasSuffix(dirName, tmpSuffix) {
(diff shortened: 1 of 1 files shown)


file_snapshot.go
```diff
@@ -240,7 +240,11 @@ func (f *FileSnapshotStore) getSnapshots() ([]*fileSnapshotMeta, error) {
 		// Ignore any temporary snapshots
 		dirName := snap.Name()
 		if strings.HasSuffix(dirName, tmpSuffix) {
-			f.logger.Printf("[WARN] snapshot: Found temporary snapshot: %v", dirName)
+			f.logger.Printf("[WARN] snapshot: Found temporary snapshot: %v, deleting..", dirName)
```
slackpad: Hmm - this is good to do but it feels wrong as a side effect got a `getSnapshots()` call, since I don't expect that to write anything. It seems like we should attempt this over in the path where the snapshot failed.

slackpad: You could add cleanup of temp snapshots after here, since this is the thread where snapshots get created, that way they will get periodically cleaned up:

https://github.com/hashicorp/raft/blob/library-v2-stage-one/snapshot.go#L72

slackpad: Having them in this thread will ensure that we don't step on a temp snapshot while it is being created, I think putting it into the read path could have been a dangerous from that POV as well.

preetapan: How about adding this logic to ReapSnapshots (https://github.com/hashicorp/raft/blob/8b2b46d7a9e2c2d95d2262afa983e048a2c45ca3/file_snapshot.go#L346-L346) which gets called from snapshot.go via sink.Close. I was primarily concerned with fixing the file system snapshots (other implementations of `SnapshotSink` are inmemory and one that discards it all)

slackpad: I think that would work and makes sense. I'd double check what context the `Close()` gets called on to ensure that it's done as part of taking a snapshot. I think it is, otherwise reap seems like it would be unsafe, too, but good to double check where that gets called.

file_snapshot.go
```diff
@@ -384,6 +384,11 @@ func (s *FileSnapshotSink) Close() error {
 	// Close the open handles
 	if err := s.finalize(); err != nil {
 		s.logger.Printf("[ERR] snapshot: Failed to finalize snapshot: %v", err)
+		s.logger.Printf("[INFO] snapshot: Deleting temporary files...")
```
preetapan: I ended up moving it here because finalize() is where it tries to save files to the temporary directory. If that fails, we can clean up after ourselves by deleting the temp directory right there.

slackpad: Won't this delete all snapshots?

slackpad: I'd get rid of this `INFO` line since we have an error if it fails, and it's pretty expected that it will clean up.

file_snapshot.go
```diff
@@ -384,6 +384,11 @@ func (s *FileSnapshotSink) Close() error {
...
+		if delErr := os.RemoveAll(s.dir); delErr != nil {
```
slackpad: This is definitely a temp dir, right? I looked at the code and I think this is fine - just want to double check it doesn't blow away snapshots.

file_snapshot.go
```diff
@@ -384,6 +384,11 @@ func (s *FileSnapshotSink) Close() error {
...
+			s.logger.Printf("[ERR] snapshot: Failed to delete temporary snapshot at path %v: %v", s.dir, err)
```
slackpad: Should be `delErr` in here.

slackpad:
Once you get this in, can you merge the same change to master and library-v2-stage-two?

file_snapshot.go
preetapan: Yes this is a temp dir - this is at the point where its about to save to a temp directory and it fails. So we can delete that temp directory before returning. 
Tested locally, unfortunately we don't use interfaces to wrap file system calls so that I could mock out an error in a unit test. That's a much larger refactor so I didn't do that right now.

preetapan:
Merged similar change to master and library-v2-stage-two.

PR #227, opened 2017-07-11
Add fuzztest testsuite

superfell:
We've been running this test suite internally for quite some while, i refactored it to make it part of the raft library. This was responsible for finding most of the issues that we've previously sent patches for.


superfell:
fyi @slackpad @ongardie

superfell:
ping @slackpad

slackpad:
This looks awesome - thanks @superfell and sorry it took me a little while to get to this.

PR #230, opened 2017-08-03
Make sure to call sync during snapshot saves

preetapan:


preetapan:
@slackpad to your question around ignoring temporary directories, we do that already. However, the directory had gotten renamed but was missing state.bin, which is treated as a legit error.

file_snapshot.go
```diff
@@ -404,6 +406,23 @@ func (s *FileSnapshotSink) Close() error {
...
+		s.logger.Printf("[ERR] snapshot: Failed to open snapshot parent directory %v, error: %v", s.parentDir, err)
+		return err
+	}
+
+	if err = parentFH.Sync(); err != nil {
+		s.logger.Printf("[ERR] snapshot: Failed syncing parent directory %v, error: %v", s.parentDir, err)
+		return err
+	}
+
+	if err = parentFH.Close(); err != nil {
```
slackpad: Should you defer this so we don't leak an fd if we get an error?

PR #238, opened 2017-08-15
Introduce AddressProvider interface that can be wired into the net tr…

preetapan:
…ansport layer. This allows overriding ip addresses of nodes when establishing connections to peers. All methods in transport interface now take a server ID to pass to the AddressProvider.

### DIFF
--- a/api.go
+++ b/api.go
@@ -492,8 +492,25 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		}
 		r.processConfigurationLogEntry(&entry)
 	}
+
 	r.logger.Printf("[INFO] raft: Initial configuration (index=%d): %+v",
 		r.configurations.latestIndex, r.configurations.latest.Servers)
+
+	mismatchedIP := false
+	// Fix IP address mismatches between processConfiguration log entries and this server's address
+	for index, server := range r.configurations.latest.Servers {
+		if server.ID == localID && server.Address != localAddr {
+			r.logger.Printf("[INFO] raft: Fixing IP address of server from %v to %v", server.Address, localAddr)
+			r.configurations.latest.Servers[index] = Server{server.Suffrage, server.ID, localAddr}
+			mismatchedIP = true
+			break
+		}
+	}
+
+	if mismatchedIP {
+		r.logger.Printf("[INFO] raft: Configuration after IP address fixes (index=%d): %+v",
+			r.configurations.latestIndex, r.configurations.latest.Servers)
+	}
 
 	// Setup a heartbeat fast-path to avoid head-of-line
 	// blocking where possible. It MUST be safe for this
--- a/api.go
+++ b/api.go
@@ -496,22 +496,6 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 	r.logger.Printf("[INFO] raft: Initial configuration (index=%d): %+v",
 		r.configurations.latestIndex, r.configurations.latest.Servers)
 
-	mismatchedIP := false
-	// Fix IP address mismatches between processConfiguration log entries and this server's address
-	for index, server := range r.configurations.latest.Servers {
-		if server.ID == localID && server.Address != localAddr {
-			r.logger.Printf("[INFO] raft: Fixing IP address of server from %v to %v", server.Address, localAddr)
-			r.configurations.latest.Servers[index] = Server{server.Suffrage, server.ID, localAddr}
-			mismatchedIP = true
-			break
-		}
-	}
-
-	if mismatchedIP {
-		r.logger.Printf("[INFO] raft: Configuration after IP address fixes (index=%d): %+v",
-			r.configurations.latestIndex, r.configurations.latest.Servers)
-	}
-
 	// Setup a heartbeat fast-path to avoid head-of-line
 	// blocking where possible. It MUST be safe for this
 	// to be called concurrently with a blocking RPC.
--- a/net_transport.go
+++ b/net_transport.go
@@ -68,6 +68,8 @@ type NetworkTransport struct {
 
 	maxPool int
 
+	serverAddressProvider ServerAddressProvider
+
 	shutdown     bool
 	shutdownCh   chan struct{}
 	shutdownLock sync.Mutex
@@ -76,6 +78,10 @@ type NetworkTransport struct {
 
 	timeout      time.Duration
 	TimeoutScale int
+}
+
+type ServerAddressProvider interface {
+	ServerAddr(id ServerID) ServerAddress
 }
 
 // StreamLayer is used with the NetworkTransport to provide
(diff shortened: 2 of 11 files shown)


api.go
```diff
@@ -492,9 +492,26 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		}
 		r.processConfigurationLogEntry(&entry)
 	}
+
 	r.logger.Printf("[INFO] raft: Initial configuration (index=%d): %+v",
 		r.configurations.latestIndex, r.configurations.latest.Servers)
 
+	mismatchedIP := false
```
slackpad: I don't think we want to do this since each server will only fix themselves, and it will result in each of them having a different configuration, which seems dangerous.

net_transport.go
```diff
@@ -155,6 +161,63 @@ func NewNetworkTransportWithLogger(
 	return trans
 }
 
+// NewNetworkTransportWithServerAddressProvider creates a new network transport with the given dialer
```
slackpad: This pattern won't scale super well and duplicates a lot. How about a `NewNetworkTransportWithConfig` with a new `NetworkTransportConfig` struct that has an optional provider, and then you can call that from the old `NewNetworkTransport` and `NewNetworkTransportWithLogger` methods as well?

api.go
preetapan: This was necessary for raft peers call to show the correct IP address for whoever becomes the leader. Otherwise, even though the network communication between all the peers work, it looks incorrect to the operator.  
I didn't notice configuration being incorrect when i tried doing `consul operator raft list-peers` using the http-addr of each server with this in place. Is there some other way to view configuration that's like a local only view that I am missing?

slackpad: Talked in person and we should be able to delete this. `consul operator raft list-peers` is the right thing to look at, and this will be incorrect until the reconcile loop fixes it up, but there's no need to fix the local node up like this. The local node will determine if it has a vote by ID - https://github.com/hashicorp/raft/blob/library-v2-stage-one/configuration.go#L145-L154.

net_transport.go
```diff
@@ -141,15 +157,31 @@ func NewNetworkTransportWithLogger(
 	if logger == nil {
 		logger = log.New(os.Stderr, "", log.LstdFlags)
 	}
+	config := &NetworkTransportConfig{Stream: stream, MaxPool: maxPool, Timeout: timeout, Logger: logger}
+	return NewNetworkTransportWithConfig(config)
+}
+
+// NewNetworkTransportWithServerAddressProvider creates a new network transport with the given config
+// struct that encapsulates dialer, listener and server address provider. The maxPool controls how many connections we will pool. The
```
slackpad: This documentation would make more sense moved up into the config structure.

net_transport.go
```diff
@@ -141,15 +157,31 @@ func NewNetworkTransportWithLogger(
 	if logger == nil {
```
slackpad: This logic is already in `NewNetworkTransportWithConfig()`.

net_transport.go
```diff
@@ -214,6 +246,17 @@ func (n *NetworkTransport) getPooledConn(target ServerAddress) *netConn {
 	return conn
 }
 
+// getConnFromAddressProvider returns a connection from the server address provider if available, or defaults to a connection using the target server address
+func (n *NetworkTransport) getConnFromAddressProvider(id ServerID, target ServerAddress) (*netConn, error) {
+	if n.serverAddressProvider != nil {
+		serverAddressOverride := n.serverAddressProvider.ServerAddr(id)
```
slackpad: I wonder if it's worthwhile to add an error return from the provider as well? Might make for a richer log message in this layer if something goes wrong in the lookup.

tcp_transport.go
```diff
@@ -42,15 +42,26 @@ func NewTCPTransportWithLogger(
 	timeout time.Duration,
 	logger *log.Logger,
 ) (*NetworkTransport, error) {
-	return newTCPTransport(bindAddr, advertise, maxPool, timeout, func(stream StreamLayer) *NetworkTransport {
+	return newTCPTransport(bindAddr, advertise, func(stream StreamLayer) *NetworkTransport {
 		return NewNetworkTransportWithLogger(stream, maxPool, timeout, logger)
 	})
 }
 
+// NewTCPTransportWithLogger returns a NetworkTransport that is built on top of
```
slackpad: This comment is stale.

net_transport_test.go
```diff
@@ -72,274 +80,295 @@ func TestNetworkTransport_Heartbeat_FastPath(t *testing.T) {
...
 
-	pipeline, err := trans2.AppendEntriesPipeline(trans1.LocalAddr())
-	if err != nil {
-		t.Fatalf("err: %v", err)
+		// Verify the response
+		if !reflect.DeepEqual(resp, out) {
+			t.Fatalf("command mismatch: %#v %#v", resp, out)
+		}
+
+		trans2.Close()
```
slackpad: Why not `defer` these in the right spots?

net_transport.go
slackpad: `EncodePeer` doesn't have an error return, so it would need to be added there as well.

preetapan: I thought about that, but we would still fall back to the address in the method arg, so it means logging it here as opposed to logging it on the consul side, and you have more info like the datacenter on the consul side to log it there. 

It does make for a cleaner interface than assuming special return values like "", so that could be the argument for adding it.

net_transport.go
```diff
@@ -80,6 +80,12 @@ type NetworkTransport struct {
 	TimeoutScale int
 }
 
+// NetworkTransportConfig encapsulates configuration for the network transport layer.
```
slackpad: One last nit - can you move the description of each thing above each member? That'll be easier to maintain and read.

net_transport.go
```diff
@@ -247,13 +247,21 @@ func (n *NetworkTransport) getPooledConn(target ServerAddress) *netConn {
...
+}
+
+func (n *NetworkTransport) getProviderAddressOrFallback(id ServerID, target ServerAddress) ServerAddress {
 	if n.serverAddressProvider != nil {
-		serverAddressOverride := n.serverAddressProvider.ServerAddr(id)
-		if serverAddressOverride != "" {
-			return n.getConn(serverAddressOverride)
+		serverAddressOverride, err := n.serverAddressProvider.ServerAddr(id)
+		if err != nil {
+			n.logger.Printf("[WARN] Unable to get address for server id %v, got error:%v", id, err)
```
slackpad: I'd make this just one log line since it's the same event (and make it a WARN for the one line).

net_transport.go
```diff
@@ -255,8 +259,7 @@ func (n *NetworkTransport) getProviderAddressOrFallback(id ServerID, target Serv
 	if n.serverAddressProvider != nil {
 		serverAddressOverride, err := n.serverAddressProvider.ServerAddr(id)
 		if err != nil {
-			n.logger.Printf("[WARN] Unable to get address for server id %v, got error:%v", id, err)
-			n.logger.Printf("[INFO] Using fallback address %v", target)
+			n.logger.Printf("[WARN] Unable to get address for server id %v, got error:%v. Using fallback address", id, err, target)
```
slackpad: The args don't match the format string for this one, and i'd put the error at the end since it could be anything, so:

`n.logger.Printf("[WARN] Unable to get address for server id %v, using fallback address %v: %v", id, target, err)`

PR #242, opened 2017-08-24
Disable parent directory fsync on windows

preetapan:


preetapan:
Decided to disable this parent fsync on windows, because its definitely an added protection that only applies to unix style filesystems. 

Haven't tested on a windows VM yet, I am going to set that up today.

preetapan:
Verified that this fixes the `file handle is invalid` error seen on windows

PR #247, opened 2017-09-18
Add support for snapshots that don't have a known size

JelteF:
Follow up on #210 that can be merged into `library-v2-stage-one`. 

Right now snapshotting is heavily bound to sending a single byte stream of
which the size is known in advance, such as sending a file stored on disk.
However, in some cases the size is not known. For instance when gzipping the
data that is being sent, or as in my case when sending a directory that is
being tarred and untarred on the fly. This allows specifying a `Size` of `-1` to
indicate that the size should not be checked by raft and that other mechanisms
for integrity are used.

It also changes the `NetworkTransport` to support this automatically for half
closable connections, such as TCP. Closing of the write half of the connection
by the sender will be used to indicate the end of the snapshot and a success
response can still be sent back by the receiver.

It also adds a `EstimatedSize` field to `SnapshotMeta` and
`InstallSnapshotRequest`. This is used when determining the timeout for the
transfer.

This closes #210, as that contains some problems fixed here.

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -111,6 +111,11 @@ type StreamLayer interface {
 	Dial(address ServerAddress, timeout time.Duration) (net.Conn, error)
 }
 
+type HalfClosableConn interface {
+	net.Conn
+	CloseWrite() error
+}
+
 type netConn struct {
 	target ServerAddress
 	conn   net.Conn
@@ -371,7 +376,12 @@ func (n *NetworkTransport) InstallSnapshot(id ServerID, target ServerAddress, ar
 
 	// Set a deadline, scaled by request size
 	if n.timeout > 0 {
-		timeout := n.timeout * time.Duration(args.Size/int64(n.TimeoutScale))
+		transferSize := args.EstimatedSize
+		if args.Size > 0 {
+			transferSize = args.Size
+		}
+
+		timeout := n.timeout * time.Duration(transferSize/int64(n.TimeoutScale))
 		if timeout < n.timeout {
 			timeout = n.timeout
 		}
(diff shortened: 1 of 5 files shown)


net_transport.go
```diff
@@ -371,7 +376,12 @@ func (n *NetworkTransport) InstallSnapshot(id ServerID, target ServerAddress, ar
 
 	// Set a deadline, scaled by request size
 	if n.timeout > 0 {
-		timeout := n.timeout * time.Duration(args.Size/int64(n.TimeoutScale))
+		transferSize := args.EstimatedSize
+		if args.Size > 0 {
+			transferSize = args.Size
+		}
+
```
preetapan: Would be good to check if both size and estimated size are set to -1 and fail fast if so

net_transport.go
```diff
@@ -393,6 +406,14 @@ func (n *NetworkTransport) InstallSnapshot(id ServerID, target ServerAddress, ar
 		return err
 	}
 
+	if halfClosableConn, ok := conn.conn.(HalfClosableConn); ok {
+		// Close the writing half of the connection to indicate all data is sent
+		err = halfClosableConn.CloseWrite()
```
preetapan: I'll note that we've had problems with Consul snapshots inside a docker container because it doesn't properly support half close https://github.com/docker/libnetwork/issues/1204. Unclear if newer versions of docker have fixed this. @slackpad any thoughts on whether this is an acceptable risk ? ie. merging this may increase the blast radius of potential issues with using snapshots inside Consul/Nomad when they are running in containers.

net_transport.go
```diff
@@ -111,6 +111,11 @@ type StreamLayer interface {
 	Dial(address ServerAddress, timeout time.Duration) (net.Conn, error)
 }
```
preetapan: Can you also add a test case to net_transport_test.go that sets the EstimatedSize param instead of size?

net_transport.go
slackpad: That's a good point - I'd make the use of half-close opt-in so people can consciously use it. For Consul and Nomad I think we'd leave as-is (the API-driven snapshot and restore operations use the half-close for their internal RPCs and we have seen folks hit issues there).

JelteF: I could add a check to only do this when `Size == -1`. Would that be sufficient?

preetapan: That seems to unnecessarily couple things. I see half close as a good thing to do, we just need to make it opt in, so having a boolean for it in the request, and defaulting to false would work so we can disable it.

langmartin:
Hey @JelteF, thanks for your contribution! We're closing this without merging now, given the age of the change and the open comments. If you'd like to follow up, especially clarifying the use case, we'd be happy to take a look at it again. Thanks!

PR #252, opened 2017-09-29
Master mergelibrary v2 stage one

preetapan:


preetapan:
I didn't want to push to master directly, made this PR instead. I'll note in https://github.com/hashicorp/raft/pull/252/commits/0af042659b6b1067c00d87253d7ca3cf31b7896d what I had to fix manually. It was quite minor

fuzzy/transport.go
```diff
@@ -0,0 +1,322 @@
...
+func (e *appendEntry) Respond(err error) {
+	e.err = err
+	close(e.ready)
+	e.consumer <- e
+}
+
+type pipeline struct {
+	t        *transport
+	target   raft.ServerAddress
+	id       raft.ServerID
```
preetapan: Added this id here and modified all interface methods in the fuzzy test to send the id along

PR #255, opened 2017-10-23
Travis: Fix build and test against newer go versions

### LINKED ISSUE #251: Travis is failing because of go 1.6 not being supported by go-metrics
Maybe we should disable the 1.6 tests and test 1.8 instead because it contains `sort.Slice`. Another possibility is to vendor/pin go-metrics to some version before it broke 1.6 compatibility.

JelteF:
This fixes #251

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: go
 
 go:
-    - 1.6
+    # Disabled until https://github.com/armon/go-metrics/issues/59 is fixed
+    # - 1.6
+    - 1.8
+    - 1.9
     - tip
 
 install: make deps


JelteF:
I'm quite sure the failure is caused by some test case failing randomly. Could someone with permissions restart it?

duckie:
@preetapan released the first stable `1.0.0` with a failed build status. This is very bad practice and I'm surprised a respectable team of developers such as Hashicorp let this slip. Maybe since raft is just a dependency of the other products the other builds are reference. 

I know how it is to work for a small successful company,, you dont have much time, but it is very small work to fix this. Please drop support of `1.6` so that we can move on with a green light on the continuous integration chain.

preetapan:
@duckie I've merged this PR. We currently have some timing-dependent tests that don't do well in Travis so we run the test suite locally on a more powerful machine before each release.  Travis results not going to be 100% reliable till we can improve the tests.

duckie:
Sorry if it felt like finger pointing, this wasnt my point. I edited my previous post to reflect it. I know how frustrating timed tests can be, but maybe such tests should be simply disabled then, when a failure is known not to reflect reality.

I knew that appealing to your love for good practives would make a hit, this wasnt a very honest move I admit.

Thanks for merging, I was waiting on it to move on with my code.

PR #256, opened 2017-10-24
Make shutdown graceful and improve and speed up election

JelteF:
I tried to make the commit messages speak for themselves. I made this mainly because I want to be able to kill a node nicely so it can be updated, even if it's the leader. The following are the changes I made (some depend on the ones before):

- Add HeartbeatInterval configuration
- Always accept elections for newer terms
- Allow election timeout to be shorter than heartbeat timeout
- Make maximum backoff duration configurable
- Make just started nodes wait twice the maximum backoff time before starting an election
- Make shutdown graceful by notifying followers

Setting a low heartbeat timeout is not always an option, because this can cause issues when the node is very busy. This way the election timeout can be as low as the network allows, as nodes are not doing anything else during an election. This allows us to set the election timeout to 50ms and when the shutdown is done gracefully this means that the cluster will only have a downtime of something like 200ms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: go
 
 go:
-    - 1.6
+    # Disabled until https://github.com/armon/go-metrics/issues/59 is fixed
+    # - 1.6
+    - 1.8
+    - 1.9
     - tip
 
 install: make deps
--- a/api.go
+++ b/api.go
@@ -156,6 +156,11 @@ type Raft struct {
 	// is indexed by an artificial ID which is used for deregistration.
 	observersLock sync.RWMutex
 	observers     map[uint64]*Observer
+
+	// startedAt is used to store the time at which raft started running.
+	startedAt time.Time
+	// receivedRequestVote stores if the node has received a request to vote.
+	receivedRequestVote bool
 }
 
 // BootstrapCluster initializes a server's storage with the given cluster
--- a/api.go
+++ b/api.go
@@ -123,6 +123,11 @@ type Raft struct {
 	shutdownCh   chan struct{}
 	shutdownLock sync.Mutex
 
+	// Shutting down boolean to check if raft is shutting down, protected by a
+	// mutex.
+	shuttingDownLock sync.Mutex
+	shuttingDown     bool
+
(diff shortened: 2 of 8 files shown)


JelteF:
Hmm, tests are failing. I will take a look at that tomorrow.

JelteF:
I fixed most of the failing tests. For more insight on one of the changes please read my post to the raft-dev mailinglist: https://groups.google.com/forum/#!topic/raft-dev/JEtBYaPpHXo

JelteF:
Ok, I found the problem with the final failing test. I had removed a check that I thought was unnecessary. But after looking at it better it seemed like it could be the cause to the issue. Which is now fixed by adding the change back.

JelteF:
@slackpad or @preetapan any feedback on this would be greatly apreciated.

connor4312:
We were occasionally running into the problem in the referenced issue. We swapped to using the fork in this PR and so far things are looking good. Would be great to see it merged -- or at least followed up on :)

JelteF:
@connor4312 thanks for trying out the PR. We also use it in production at the company I work at. 

Regarding the discussion around this. Part of this happened on the mailing list thread where @preetapan replied with the results from an internal Hashicorp meeting: https://groups.google.com/forum/#!topic/raft-dev/JEtBYaPpHXo As you can read there, my solution can result in excessive leader switching when one node has an intermittent connection to the other nodes, which is why they don't want to accept it as is. So keep that in mind when using this fork. At my company we decided that the gain from the changes was worth this edge case problem. Mainly because the problem that it can cause is very unlikely to occur in our infrastructure and the issue that it solves was happening regularly.

I tried to get @preetapan to discuss a bit more with me on the topic, but I didn't get a response anymore. I would like to change the PR to use the pre-vote RPC, but I would like to have some more feedback before doing that.

JelteF:
BTW, be sure to fork/vendor our fork. This branch might be removed if the PR ever gets merged.

s-christoff:
Howdy @JelteF, 
We really appreciate your contribution. We are going to close this pull request because it is stale against the rest of the code. But we may have someone look at it again in the future!

PR #264, opened 2017-11-29
Incorporate the ServerID in node identity verification.

fd:
* encode/decode ServerID in transports
* track the leader ServerID
* use the leader ServerID as a second identifier

Votes will be rejected when `candidate(id,addr)` doesn't match up with `leader(id,addr)`. The reasoning here is that a change of the ServerID or the ServerAddress indicates a change in the peer process (This is especially important in more dynamic environments like kubernetes where address can be reused by different nodes).

I'm aware of the library-v2-stage-two branch but it seems to be lagging behind. Feel free to reject this PR.

### DIFF
--- a/api.go
+++ b/api.go
@@ -87,8 +87,9 @@ type Raft struct {
 	lastContactLock sync.RWMutex
 
 	// Leader is the current cluster leader
-	leader     ServerAddress
-	leaderLock sync.RWMutex
+	leaderID      ServerID
+	leaderAddress ServerAddress
+	leaderLock    sync.RWMutex
 
 	// leaderCh is used to notify of leadership changes
 	leaderCh chan bool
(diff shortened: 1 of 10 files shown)


fd:
Tests are to be failing on `go1.6` because `github.com/armon/go-metrics` is no longer compatible.

```
go test -timeout=60s .
# github.com/armon/go-metrics
../../armon/go-metrics/inmem_endpoint.go:72: undefined: sort.Slice
../../armon/go-metrics/inmem_endpoint.go:86: undefined: sort.Slice
../../armon/go-metrics/inmem_endpoint.go:113: undefined: sort.Slice
```

JelteF:
If you merge master the 1.6 errors should be fixed now.

fd:
@JelteF are the integration tests flaky?

JelteF:
Yeah they are apparently: https://github.com/hashicorp/raft/pull/255#issuecomment-349138170

s-christoff:
Howdy @fd,  thank you so much for your contribution. 
We've changed how we handle server identification in recent releases and therefore this PR is stale. We're going to close this for now, but you can still comment to discuss.

fd:
Hi there @s-christoff, you are right this PR is stale. Totally forgot about it as well. 😂

PR #267, opened 2018-01-17
Add functionality to temporarily close and reload raft stream layer

chelseakomlo:


net_transport.go
```diff
@@ -199,6 +205,28 @@ func (n *NetworkTransport) SetHeartbeatHandler(cb func(rpc RPC)) {
 	n.heartbeatFn = cb
 }
 
+// UseStream re-initializes the stream layer
+func (n *NetworkTransport) UseStream(s StreamLayer) {
```
preetapan: Consider making this return an error and do nothing if CloseStream was not previously called to prevent leaking connections due to programmer error. 
One way to do this would be to set a boolean that the stream layer has been closed when it receives an event on the Done channel in the context used here, and check against the state of that boolean before reinitializing the stream layer.

dadgar:
@chelseakomlo and I paired on this and changed the functionality to be just to close the existing connections. We will be modifying the StreamLayer such that new connections will be using a new TLS wrapper config.

PR #270, opened 2018-01-24
Copy the contents of an InmemSnapshotStore when opening a snapshot

### LINKED ISSUE #269: InmemSnapshotStore.Open returns a reader with no content if used twice
This happens when trying to open the same snapshot a second time. The current code in of InmemSnapshotStore.Open() looks like:

```go
return &m.latest.meta, ioutil.NopCloser(m.latest.contents), nil
```

which means that a second call would get a zero-length reader, since m.latest.contents is a bytes.Buffer that can only be read once, and returns io.EOF the second time.

freeekanayaka:
This fixes #269 and is needed since otherwise a second call to
InmemSnapshotStore.Open would result in a zero-length content, because
bytes.Buffer can only be read once.

### DIFF
--- a/inmem_snapshot.go
+++ b/inmem_snapshot.go
@@ -82,7 +82,10 @@ func (m *InmemSnapshotStore) Open(id string) (*SnapshotMeta, io.ReadCloser, erro
 		return nil, nil, fmt.Errorf("[ERR] snapshot: failed to open snapshot id: %s", id)
 	}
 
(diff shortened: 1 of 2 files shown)


freeekanayaka:
Both the failures in Travis seem unrelated and due to flaky tests. You might want to retrigger the build (I don't have privileges for that).

RobbieMcKinstry:
Thank you! LGTM!

PR #271, opened 2018-01-30
Clarify AddVoter and AddNonvoter docs

axw:
Update the AddVoter and AddNonvoter doc comments to indicate that they will update the address of existing cluster members. They currently state that, if the members are currently in the configuration, the methods "do nothing".


kyhavlov:
Thanks!

Issue #273, opened 2018-02-08
Race between AppendEntriesPipeline and Disconnect in InmemTransport


freeekanayaka:
The current code of AppendEntriesPipeline() in InmemTransport reads:

```go
	i.RLock()
	peer, ok := i.peers[target]
	i.RUnlock()
	if !ok {
		return nil, fmt.Errorf("failed to connect to peer: %v", target)
	}
	pipeline := newInmemPipeline(i, peer, target)
	i.Lock()
	i.pipelines = append(i.pipelines, pipeline)
	i.Unlock()
	return pipeline, nil
```

which means that if a peer gets deleted via Disconnect() right after the i.RUnlock() call, then AppendEntriesPipeline will succeed in creating the pipeline and the transport will never be really disconnected (as calls to inmemPipeline.AppendEntries will keep succeeding).

I'm attaching a sample program below which reproduces this race. Just build it and run it in a while loop and it will eventually fail because of the above issue.

freeekanayaka:
```go
package main

import (
	"io"
	"log"
	"strconv"
	"time"

	"github.com/hashicorp/raft"
)

func main() {
	n := 3 // Number of nodes
	for i := 0; i < n; i++ {
	}

	// Create raft dependencies
	configs := newConfigs(n)
	stores := newStores(n)
	snapshots := newSnapshotStores(n)
	transports := newTranports(n)

	// Bootstrap initial configuration
	configuration := newConfiguration(n)
	for i := 0; i < n; i++ {
		err := raft.BootstrapCluster(
			configs[i],
			stores[i],
			stores[i],
			snapshots[i],
			transports[i],
			configuration,
		)
		if err != nil {
			log.Fatalf("bootstrap error: %v", err)
		}
	}

	// Setup notification channels
	notifyChs := make([]chan bool, n)
	for i := 0; i < n; i++ {
		notifyChs[i] = make(chan bool, 0)
		configs[i].NotifyCh = notifyChs[i]
	}

	// Start raft instances.
	fsms := newFSMs(n)
	rafts := make([]*raft.Raft, n)
	for i := 0; i < n; i++ {
		raft, err := raft.NewRaft(
			configs[i],
			fsms[i],
			stores[i],
			stores[i],
			snapshots[i],
			transports[i],
		)
		if err != nil {
			log.Fatalf("start error: %v", err)
		}
		rafts[i] = raft
	}

	// Wait for a leader to be elected.
	var i int
	select {
	case <-notifyChs[0]:
[...]

PR #274, opened 2018-02-08
Fix race between AppendEntriesPipeline and Disconnect in InmemTransport

freeekanayaka:
Acquire the write lock unconditionally at the beginning of
AppendEntriesPipeline, to prevent Disconnect from racing with it by
deleting a peer inbetween the read and write lock.

With this fix the sample program attached to #273 does not fail anymore.


mkeeler:
For the record, the tests pass locally. It seems like we have some flakiness issues with the raft tests similar to the Consul tests.

freeekanayaka:
Great to see all the pending PRs moving, thanks and keep on @mkeeler !

Issue #268, opened 2018-01-22
The Raft.Shutdown() API can end up in a deadlock


freeekanayaka:
I've noticed that occasionally some unit tests of [dqlite](https://github.com/CanonicalLtd/dqlite) hang forever.

I **believe** that the reason is a potential deadlock between Raft.Shutdown() and the Raft.runSnapshots() goroutine. The timeline of events looks roughly like:

1. The user calls Raft.Shutdown().Error()
2. Roughly at the same time runSnapshots() decides that it's time to kick Raft.takeSnapshot()
3. The Raft.takeSnapshot() method manages to run a bit and then block on [configReq.Error()](https://github.com/hashicorp/raft/blob/master/snapshot.go#L155) call. This can happen if the shutdown signal is being dispatched to the shutdownCh but it's not quite done, and also the call to ```r.configurationsCh <- configReq``` at line 151 succeeded (because configurationsCh is buffered with size 8), but the future was not actually processed by the main raft goroutine (either runLeader, runCandidate or runFollower)
3. The main goroutine of Raft (either runLeader, runCandidate or runFollower) receives the shutdown signal and exits **without** actually processing the pending request in configurationsCh, hence leaving takeSnapshot blocked.
4. The Error() method of future returned by Raft.Shutdown() is blocked waiting for all raft goroutines to finish, but runSnapshots will never return because it's blocked on takeSnapshot() which is in turn blocked on configReq.Error().

Hope this makes sense.

To see an example of a unit test that failed as above, look [here](https://travis-ci.org/CanonicalLtd/dqlite/builds/331818891#L539).

I'm pasting the relevant Go traceback below as well for reference.

freeekanayaka:
Traceback from the Travis run linked above:

```
2018/01/22 13:57:28 panic: test timed out after 1m0s
2018/01/22 13:57:28 
2018/01/22 13:57:28 goroutine 350 [running]:
2018/01/22 13:57:28 testing.startAlarm.func1()
2018/01/22 13:57:28 	/home/travis/.gimme/versions/go1.9.linux.amd64/src/testing/testing.go:1145 +0xf9
2018/01/22 13:57:28 created by time.goFunc
2018/01/22 13:57:28 	/home/travis/.gimme/versions/go1.9.linux.amd64/src/time/sleep.go:170 +0x44
2018/01/22 13:57:28 
2018/01/22 13:57:28 goroutine 1 [chan receive]:
[...]

PR #277, opened 2018-02-12
Fix deadlock between Shutdown() and takeSnapshot()

freeekanayaka:
This change makes sure that a takeSnapshot() call which is run about
at the same time as Shutdown() does not block indefinitely on a
ConfigurationFuture request.

The situation can happen if takeSnapshot() manages to push the request
to configurationCh (which is buffered) but the
follower/candidate/leader loop exists before processing it, in turn
making the shutdownFuture.Error() returned by Shutdown() block
indefinitely (since it indirectly waits for takeSnapshot to finish).

With this change the sample program attached to #268 doesn't fail
anymore.

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -152,9 +152,24 @@ func (r *Raft) takeSnapshot() (string, error) {
 	case <-r.shutdownCh:
 		return "", ErrRaftShutdown
 	}
-	if err := configReq.Error(); err != nil {
-		return "", err
+
+	// Wait for the future to complete, but abort if a shutdown happens in
+	// the meantime, and our configure request did not get processed
+	// despite it was successfully sent to the buffered configurationsCh
+	// above (see #268).
+	errorCh := make(chan error)
+	go func() {
+		errorCh <- configReq.Error()
+	}()
+	select {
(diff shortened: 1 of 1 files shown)


freeekanayaka:
@mkeeler your diff sounds fine to me (although if I had to make a nit, it'd be probably better to wait for at least one more use case of something like deferError.ShutdownCh before introducing such an ad-hoc API).

Please make sure that the sample program I attached to #277 does not crash anymore with your diff applied.

freeekanayaka:
I don't if it still applies, it surely did when I pushed the PR. I had even attached a program to reproduce the issue, which you can use to check it.

Poluect:
Hey guys! Are there any plans to fix this issue? example code mentioned in #268, still can reproduce the issue from actual master branch, and I found this bug while searching for some clues on why nomad server is freezing sometimes in our system, seems like a go routine responsible for dealing with allocations hangs, and then found raft calls inside. Not sure if this is a reason of our nomad freezes, but at least this bug still can be reproduced. Thanks!

i0rek:
Will look at this soon.

PR #278, opened 2018-02-12
Make TestNetworkTransport_CloseStreams less flaky

freeekanayaka:
This change improves the robustness of the
TestNetworkTransport_CloseStreams test, which might occasionally fail
on Travis (as in [1]) if Transport 1 consumes RPCs fast enough to not
stress Transport 2's pool as expected.

The test is still flaky in principle, but I belive it should be good
enough to make CI green most (all?) the times.

[1] https://travis-ci.org/hashicorp/raft/jobs/340532532


freeekanayaka:
I added some more commits with timing-related fixes, and improving diagnostics in case of error. You find the description and rational in each commit message.

Another failure that I've seen in Travis in [this](https://travis-ci.org/hashicorp/raft/jobs/340571449) build should be hopefully fixed by #277 (and if not, at least we'll get more details from the console).

Travis has now finally given me a green run, I hope merging this will help you reviewing the other PRs that I pushed, since they should have more chances to pass CI after this lands.

integ_test.go
```diff
@@ -149,15 +149,18 @@ WAIT:
...
-		panic(fmt.Errorf("timeout waiting for future %v", f))
-	})
-	defer timer.Stop()
-	return f.Error()
+func WaitFuture(f Future, t testing.TB) error {
+	if helper, ok := t.(testingHelper); ok {
+		helper.Helper()
+	}
+	msg := "timeout waiting for future %v"
+	return waitForError(t, f.Error, 2*time.Second, msg, f)
```
freeekanayaka: A bump from 200ms to 2s might seem high, and actually the most frequent failure that I've seen so far occurs when TestRaft_Integ runs WaitFuture(env1.raft.Snapshot(), t), as apparently Travis might be slow at syncing the snapshot file descriptor to disk. See for example [this](https://travis-ci.org/hashicorp/raft/jobs/340602304) build.

If it's preferred, we could keep the default timeout in WaitFuture lower, even the current 200ms, and add a more generic WaitFutureWIthTimeout that lets TestRaft_Integ specify a custom timeout in some call sites. WaitFuture would use it internally, and we could set the timeout to 2s only for the env1.raft.Snapshot() case (and possibly any other that might involve disk I/O).

freeekanayaka:
I confirm that tests in Travis are more stable now, as shown by PR #279 which includes this fix and the other I've pushed (especially #277).

In that branch, so far I've only seen TestRaft_UserRestore fail (see this [build](https://api.travis-ci.org/v3/job/340611119/log.txt))

freeekanayaka:
I rebased this branch on master, to pull in the last two commits. A new Travis run was triggered and it's again green.

I pushed also a new version of the branch that contains all the pending requests that I submitted (#279), and that is green as well.

On the contrary, all recent builds that do not contain these fixes have failed (including the recently merged ones):

- https://travis-ci.org/hashicorp/raft/builds/340710243
- https://travis-ci.org/hashicorp/raft/builds/340709344
- https://travis-ci.org/hashicorp/raft/builds/348613042

freeekanayaka:
What's upstream's opinion on this PR? It seems a safe no-brainer to me, since it only touches a few spots in _test.go files and I think I showed that it makes tests less flaky, providing analysis and explanation in the commit messages.

james-lawrence:
I agree I'd like to see these merged. have test issues in my packages due to these issues. making it harder to determine logic bugs because the underlying test code has issues.

ncabatoff:
Sorry we never followed up on this.  There has been some effort this year to improve test flakiness, I'm not sure whether this is still needed - I haven't seen TestNetworkTransport_CloseStreams failing much lately.  Regardless, if you'd like us to have another look, please open a new PR and we'll try to be more responsive this time.

PR #281, opened 2018-03-03
Make the InmemTransport.timeout field configurable

freeekanayaka:
The NewInmemTransportWithTimeout contructor supports specifying a
custom RPC timeout for an InmemTransport instance.

This timeout appears to be the only non-configurable one when creating
test-oriented Raft instances (i.e. using the in-memory version of the
various dependencies). All other timeouts can be set using Config.

Being able to tweak this timeout is useful when running tests on
machines under relatively high load, where a low timeout might produce
spurious errors.

### DIFF
--- a/inmem_transport.go
+++ b/inmem_transport.go
@@ -43,9 +43,11 @@ type InmemTransport struct {
 	timeout    time.Duration
 }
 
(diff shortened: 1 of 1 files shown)


mkeeler:
All tests pass locally

PR #294, opened 2018-08-04
remove setting last contact

### LINKED ISSUE #285: last contact is set due to a vote request?
why is last contact set when a vote request is received. it seems to have nothing to do with actual leadership. and seems incorrect. maybe I'm missing something but if a vote request is received (and granted) then that says nothing about contact with the leader node. as obviously that node may not end up elected.

```
func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest)
```

james-lawrence:
fixes #285, it also doesn't make any sense, having a candidate accepted by another node doesn't make it a leader.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1226,7 +1226,7 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 	}
 
 	resp.Granted = true
-	r.setLastContact()
+
 	return
 }
 


mkeeler:
@james-lawrence I am going to close this as the current behavior is intentional and required. See the explanation for #285 for more details.

PR #306, opened 2019-01-28
Leadership transfer

i0rek:
This PR is implementing the leadership transfer extension described in the [thesis chap 3.10](https://ramcloud.stanford.edu/~ongaro/thesis.pdf).

Background:

Consul is performing some setup after acquiring leadership. It is possible that the setup fails, but there is no good way to step down as a leader. It is possible to use `DemoteVoter` as show in https://github.com/hashicorp/consul/pull/5247, but this is suboptimal because it relies on Consul's autopilot to promote the old leader to a voter again.
Since there is a perfectly fine way described in the thesis: leadership transfer extension, we decided to implement that instead. Doing it this way also helps other teams, since it is more generic.

The necessary steps to perform are:

1. Leader picks target to transition to
2. Leader stops accepting client requests
3. Leader makes sure to replicate logs to the target
4. Leader sends TimeoutNow RPC request
5. Target receives TimeoutNow request, which triggers an election
6a. If the election is successful, a message with the new term will make the old leader step down
6b. if after electiontimeout the leadership transfer did not complete, the old leader resumes operation

Todo:

* [ ] mention leadership transfer in the readme
* [x] add fuzzy test for leadership transfer
* [x] make the leader semi respond to stuff until transition is either completed or failed
* [x] if after ~electiontimeout the leadership transfer did not complete, the old leader resumes operation
* [x] add leadershiptransfer flag to `RequestVote` as described in [chap 4.2.3 Disruptive servers](https://ramcloud.stanford.edu/~ongaro/thesis.pdf)
* [x] expire `LeaderLeaseTimeout` before leadership transfer as described in [chap 6.4.1 Using clocks to reduce messaging for read-only queries](https://ramcloud.stanford.edu/~ongaro/thesis.pdf)
* [x] are we calling the replication correctly?
* [x] add documentation in the code

Resources:

* https://github.com/etcd-io/etcd/tree/master/raft

### DIFF
--- a/api.go
+++ b/api.go
@@ -97,6 +97,8 @@ type Raft struct {
 	// leaderState used only while state is leader
 	leaderState leaderState
 
+	candidateFromTransitionLeadership bool
+
 	// Stores our local server ID, used to avoid sending RPCs to ourself
 	localID ServerID
 
@@ -469,7 +471,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		configurationsCh:       make(chan *configurationsFuture, 8),
 		bootstrapCh:            make(chan *bootstrapFuture),
 		observers:              make(map[uint64]*Observer),
-		transitionLeadershipCh: make(chan *transitionLeadershipFuture),
+		transitionLeadershipCh: make(chan *transitionLeadershipFuture, 64),
 	}
 
 	// Initialize as a follower.
--- a/api.go
+++ b/api.go
@@ -1035,7 +1035,10 @@ func (r *Raft) AppliedIndex() uint64 {
 // it is unlikely that another server is starting an election, it is very
 // likely that the target server is able to win the election.  Note that raft
 // protocol version 3 is not sufficient to use LeadershipTransfer. A recent
-// version of that library has to be used that includes this feature.
+// version of that library has to be used that includes this feature.  Using
+// transfer leadership is safe however in a cluster where not every node has
+// the latest version. If a follower cannot be promoted, it will fail
+// gracefully.
 func (r *Raft) LeadershipTransfer() Future {
 	if r.protocolVersion < 3 {
 		return errorFuture{ErrUnsupportedProtocol}
@@ -1053,7 +1056,9 @@ func (r *Raft) LeadershipTransfer() Future {
 // server in the arguments in case a leadership should be transitioned to a
 // specific server in the cluster.  Note that raft protocol version 3 is not
 // sufficient to use LeadershipTransfer. A recent version of that library has
-// to be used that includes this feature.
+// to be used that includes this feature. Using transfer leadership is safe
+// however in a cluster where not every node has the latest version. If a
+// follower cannot be promoted, it will fail gracefully.
 func (r *Raft) LeadershipTransferToServer(id ServerID, address ServerAddress) Future {
 	if r.protocolVersion < 3 {
 		return errorFuture{ErrUnsupportedProtocol}
--- a/api.go
+++ b/api.go
@@ -49,6 +49,8 @@ var (
 	// ErrCantBootstrap is returned when attempt is made to bootstrap a
 	// cluster that already has state present.
 	ErrCantBootstrap = errors.New("bootstrap only works on new clusters")
+
+	ErrLeadershipTransferInProgress = errors.New("leadership transfer in progress")
 )
 
 // Raft implements a Raft node.
--- a/api.go
+++ b/api.go
@@ -50,6 +50,8 @@ var (
 	// cluster that already has state present.
 	ErrCantBootstrap = errors.New("bootstrap only works on new clusters")
 
+	// ErrLeadershipTransferInProgress is returned when the leader is rejecting
+	// client requests because it is attempting to transfer leadership.
 	ErrLeadershipTransferInProgress = errors.New("leadership transfer in progress")
 )
 
@@ -99,6 +101,9 @@ type Raft struct {
 	// leaderState used only while state is leader
 	leaderState leaderState
 
+	// candidateFromLeadershipTransfer is used to indicate that this server became
+	// candidate because the leader tries to transfer leadership. This flag is
+	// used in requestvote to express that a leadership transfer is going on.
 	candidateFromLeadershipTransfer bool
 
 	// Stores our local server ID, used to avoid sending RPCs to ourself
@@ -162,6 +167,8 @@ type Raft struct {
 	observersLock sync.RWMutex
 	observers     map[uint64]*Observer
 
+	// leadershipTransferCh is used to start a leadership transfer from outside of
+	// the main thread.
 	leadershipTransferCh chan *leadershipTransferFuture
 }
 
--- a/api.go
+++ b/api.go
@@ -103,7 +103,8 @@ type Raft struct {
 
 	// candidateFromLeadershipTransfer is used to indicate that this server became
 	// candidate because the leader tries to transfer leadership. This flag is
-	// used in requestvote to express that a leadership transfer is going on.
+	// used in RequestVoteRequest to express that a leadership transfer is going
+	// on.
 	candidateFromLeadershipTransfer bool
 
 	// Stores our local server ID, used to avoid sending RPCs to ourself
@@ -1026,6 +1027,13 @@ func (r *Raft) AppliedIndex() uint64 {
 	return r.getLastApplied()
 }
 
+// LeadershipTransfer will transfer leadership to a server in the cluster.
+// This can only be called from the leader, or it will fail. The leader will
+// stop accepting client requests, make sure the target server is up to date and
+// starts the transfer with a TimeoutNow message. This message has the same
+// effect as if the election timeout on the on the target server fires. Since it
+// is unlikely that no other server is starting an election, it is very likely
+// that the target server is able to win the election.
 func (r *Raft) LeadershipTransfer() Future {
 	if r.protocolVersion < 3 {
 		return errorFuture{ErrUnsupportedProtocol}
@@ -1039,6 +1047,9 @@ func (r *Raft) LeadershipTransfer() Future {
 	return r.initiateLeadershipTransfer(s.ID, s.Address)
 }
 
+// LeadershipTransferToServer does the same as LeadershipTransfer but takes a
+// server in the arguments in case a leadership should be transitioned to a
+// specific server in the cluster.
 func (r *Raft) LeadershipTransferToServer(id ServerID, address ServerAddress) Future {
 	if r.protocolVersion < 3 {
 		return errorFuture{ErrUnsupportedProtocol}
--- a/api.go
+++ b/api.go
@@ -1027,7 +1027,7 @@ func (r *Raft) LeadershipTransfer() Future {
 		return errorFuture{errors.New("cannot find peer")}
 	}
 
-	return r.leadershipTransfer(s.ID, s.Address)
+	return r.initiateLeadershipTransfer(s.ID, s.Address)
 }
 
 func (r *Raft) LeadershipTransferToServer(id ServerID, address ServerAddress) Future {
@@ -1035,5 +1035,5 @@ func (r *Raft) LeadershipTransferToServer(id ServerID, address ServerAddress) Fu
 		return errorFuture{ErrUnsupportedProtocol}
 	}
 
-	return r.leadershipTransfer(id, address)
+	return r.initiateLeadershipTransfer(id, address)
 }
--- a/api.go
+++ b/api.go
@@ -1044,12 +1044,7 @@ func (r *Raft) LeadershipTransfer() Future {
 		return errorFuture{ErrUnsupportedProtocol}
 	}
 
-	s := r.pickServer()
-	if s == nil {
-		return errorFuture{errors.New("cannot find peer")}
-	}
-
-	return r.initiateLeadershipTransfer(s.ID, s.Address)
+	return r.initiateLeadershipTransfer(nil, nil)
 }
 
 // LeadershipTransferToServer does the same as LeadershipTransfer but takes a
@@ -1064,5 +1059,5 @@ func (r *Raft) LeadershipTransferToServer(id ServerID, address ServerAddress) Fu
 		return errorFuture{ErrUnsupportedProtocol}
 	}
 
-	return r.initiateLeadershipTransfer(id, address)
+	return r.initiateLeadershipTransfer(&id, &address)
 }
--- a/api.go
+++ b/api.go
@@ -1032,7 +1032,7 @@ func (r *Raft) AppliedIndex() uint64 {
 // stop accepting client requests, make sure the target server is up to date and
 // starts the transfer with a TimeoutNow message. This message has the same
 // effect as if the election timeout on the on the target server fires. Since it
-// is unlikely that no other server is starting an election, it is very likely
+// is unlikely that another server is starting an election, it is very likely
 // that the target server is able to win the election.
 func (r *Raft) LeadershipTransfer() Future {
 	if r.protocolVersion < 3 {
--- a/api.go
+++ b/api.go
@@ -49,6 +49,10 @@ var (
 	// ErrCantBootstrap is returned when attempt is made to bootstrap a
 	// cluster that already has state present.
 	ErrCantBootstrap = errors.New("bootstrap only works on new clusters")
+
+	// ErrNoTransitionLeadershipTarget is returned wher attempt is made to
+	// transition leadership but there is no peer to transition to.
+	ErrNoTransitionLeadershipTarget = errors.New("tried to transition leadership, but didn't find a peer")
 )
 
 // Raft implements a Raft node.
@@ -1021,7 +1025,26 @@ func (r *Raft) TransitionLeadership() Future {
 		return errorFuture{ErrNotLeader}
 	}
 
-	r.transitionLeadership()
+	s := r.pickTransferLeadershipTarget()
+	if s == nil {
+		return errorFuture{ErrNoTransitionLeadershipTarget}
+	}
+
+	r.transitionLeadership(s.ID)
+
+	return nil
+}
+
+func (r *Raft) TransitionLeadershipToServer(id ServerID) Future {
+	if r.protocolVersion < 3 {
+		return errorFuture{ErrUnsupportedProtocol}
+	}
+
+	if r.getState() != Leader {
+		return errorFuture{ErrNotLeader}
+	}
+
+	r.transitionLeadership(id)
 
 	return nil
 }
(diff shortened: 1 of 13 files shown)


i0rek:
@mitchellh Absolutely, I added the todos!

banks:
@tylertreat FYI. I know you were interested in this!

raft.go
```diff
@@ -652,6 +702,75 @@ func (r *Raft) verifyLeader(v *verifyFuture) {
...
+			r.logger.Printf("[DEBUG] raft: %v", err.Error())
+			future.respond(err.Error())
+			return
+		}
+	}
+
+	// Step 3: reset leader lease. Since a new leader might be choosen in a bit,
+	// we don't want this leader to have the right to answer read request based on
+	// this lease.
+	r.leaderState.lease = time.After(0)
```
banks: Hmm I'm not sure this is right for a very confusing reason!

The concept of a lease in our leader implementation is not really quite the same as the lease mechanism described in `6.4.1 Using clocks to reduce messaging for read-only queries` of the thesis.

That lease mechanism's only purpose is to support almost-linearizable reads (assuming clocks are enough in sync) without the message round trips. 

Our raft lib doesn't actually have a concept of fully linearizable reads in the same way - we assume that reads are made directly from the FSM implemented on top of this on the leader (not fully linearizable since the leader might have lost an election during partition but not noticed yet because still inside election timeout). Where fully-consistent reads are needed (i.e. `?consistent` in Consul) we still don't process the read through raft log as some raft implementations do, instead we first get a snapshot of current leader FSM state, then run verifyLeader which essentially re-runs leader election to ensure a quorum still support the leader, and then answer the request from the snapshot that we had which is linearizable because you know the server was still the only leader at the point that snapshot was valid.

The lease mechanism in our raft lib, is setup in a similar way to the one in the thesis, but in practice it's a timer that just tells the leader how often to check heartbeats are still coming in.

So if you stop the lease here, then as soon as this method exits below (right after sending TimeoutNow, but _probably before_ the actual new leader takes over, the leaderloop above will activate the `<-lease` case, re-check the heartbeats for a quorum are still coming in, **and then rest the lease timeout again**, in other words this is kind of a no-op - it just causes the heartbeat check to happen a few ms sooner than it might.

Worst case though, if heartbeats have stopped for a while (for example because it took us a few seconds to replicate log entries above) then this will mean that immediately after we send `TimeoutNow` and return here, the leader will hit the `<-lease` case and then find it's lost lease and step down - possibly before the new leader had a chance to win it's election and take over.

Ha ha the thesis in section six says:

>  Problems due to stale reads have already been discovered in two third-party Raft implementations [45]

One of those was this one before we implemented verifyLeader 😄 .

Actually our verifyLeader mechanism technically isn't the same as the one proposed in section 6 either... We don't record the leader's commit index before and wait for it to be applied to the FSM etc. in practice given the bounds on how fast other things can happen I think it's sufficient.

The tl;dr here is that I think we can just ignore this and be no less correct than now - reads that we consider "linearizable" use a different mechanism that perform their own "heartbeats" (in thesis language, actually we call them votes in verifyLeader) anyway so we effectively don't use the `lease` as a way to assume reads are consistent.

raft.go
```diff
@@ -652,6 +702,75 @@ func (r *Raft) verifyLeader(v *verifyFuture) {
...
+	for s.nextIndex <= r.getLastIndex() {
+		r.logger.Printf("[DEBUG] raft: waiting for replication to catch up before sending TimeoutNow")
+		err := &deferError{}
+		err.init()
+		s.triggerDeferErrorCh <- err
+		if err.Error() != nil {
+			r.logger.Printf("[DEBUG] raft: %v", err.Error())
+			future.respond(err.Error())
+			return
+		}
```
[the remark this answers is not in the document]
i0rek: You are correct! It could potentially sit here forever trying to catch up. I will change it so that it is time out.

raft.go
```diff
@@ -652,6 +702,75 @@ func (r *Raft) verifyLeader(v *verifyFuture) {
...
+	// Step 1: set this field which stops this leader from responding to any client requests.
+	r.leaderState.leadershipTransferInProgress = true
+
+	// Step 2: make sure the target server is up to date
+	s, ok := r.leaderState.replState[future.ID]
+	if !ok {
+		err := fmt.Errorf("cannot find replication state for %v, aborting leadership transfer", future)
+		r.logger.Printf("[DEBUG] raft: %v", err)
+		future.respond(err)
+		return
```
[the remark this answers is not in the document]
i0rek: You are right we need to add `r.leaderState.leadershipTransferInProgress = false` here. I don't think we can defer that since the happy path needs it to be enabled. Because the target becomes leader after we returned from this function and we cannot accept client requests in the meanwhile.

raft.go
```diff
@@ -505,6 +522,14 @@ func (r *Raft) leaderLoop() {
...
 
+		case future := <-r.leadershipTransferCh:
+			if r.leaderState.leadershipTransferInProgress {
+				r.logger.Printf("[DEBUG] raft: %s", ErrLeadershipTransferInProgress)
+				future.respond(ErrLeadershipTransferInProgress)
+				continue
+			}
+			r.leadershipTransfer(future)
+
 		case <-r.leaderState.commitCh:
```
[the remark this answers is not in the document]
mkeeler: Looks like you brought up the blocking nature of `leadershipTransfer` and it potentially taking a long time causing other issues in a different comment.

If we make it not block then we definitely should handle this case here.

api.go
```diff
@@ -461,6 +475,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		configurationsCh:      make(chan *configurationsFuture, 8),
 		bootstrapCh:           make(chan *bootstrapFuture),
 		observers:             make(map[uint64]*Observer),
+		leadershipTransferCh:  make(chan *leadershipTransferFuture, 64),
```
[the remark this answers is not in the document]
i0rek: I randomly picked that number. That was before I returned `ErrLeadershipTransferInProgress` instead of waiting to be able to write to the channel. I changed it to use `0` capacity.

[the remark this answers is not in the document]
i0rek: oh well, of course it needs to be 1. 🤦‍♂️

i0rek:
I addressed the fact, that the replication needs to be included in the timeout as well. I think it is much better now, because not only timeouts, but also loss of leadership can stop the transfer at any point in time: https://github.com/hashicorp/raft/blob/ded6ce1d88c0dba7140992d16eb9c138f09ba38d/raft.go#L540-L558.

i0rek:
I also defer resetting the flag for leadershiptransfer in progress now: https://github.com/hashicorp/raft/blob/ded6ce1d88c0dba7140992d16eb9c138f09ba38d/raft.go#L735-L736.

api.go
```diff
@@ -1006,3 +1021,34 @@ func (r *Raft) LastIndex() uint64 {
...
+	if s == nil {
+		return errorFuture{errors.New("cannot find peer")}
+	}
+
+	return r.initiateLeadershipTransfer(s.ID, s.Address)
+}
+
+// LeadershipTransferToServer does the same as LeadershipTransfer but takes a
+// server in the arguments in case a leadership should be transitioned to a
+// specific server in the cluster.
```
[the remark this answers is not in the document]
i0rek: It will update the server and if it happens to take longer than `electionTimeout` it will abort the transfer.

raft.go
```diff
@@ -652,6 +702,75 @@ func (r *Raft) verifyLeader(v *verifyFuture) {
...
+	// leadership is not lost after an election timeout, the current leader
+	// resumes operation and continues as if nothing has happened. We are setting
+	// everything up here, although we yet have to send TimeoutNowRequest. This is
+	// being done so that we don't miss leadership loss.
+	go func() {
+		select {
+		case err := <-verify.errCh:
+			r.logger.Printf("[DEBUG] raft: resetting leadership transfer: %s", err)
+			future.respond(nil)
+			r.leaderState.leadershipTransferInProgress = false
```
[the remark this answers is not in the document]
i0rek: I put this into a defer: https://github.com/hashicorp/raft/blob/c2eb642215d568b9925f86781a9b629c5fd1457a/raft.go#L755-L759

But since it is still being read from the mainloop, I also put it into an atomic int.

raft.go
```diff
@@ -1468,3 +1594,55 @@ func (r *Raft) setState(state RaftState) {
...
+	for _, server := range r.configurations.latest.Servers {
+		if server.ID != r.localID {
+			return &server
+		}
+	}
+	return nil
+}
+
+// pickServer returns the first server that is not itself.
+func (r *Raft) pickServer() *Server {
```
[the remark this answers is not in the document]
i0rek: I changed the implementation to do exactly that.

raft.go
i0rek: thank you for that explanation! removed it from the pr.

raft.go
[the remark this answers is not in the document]
banks: It partially blocks (replication blocks but then it runs the rest in a go routine and returns) in the implementation I commented on which is both a performance issue and a bug! Not checked the recent changes here yet though which maybe resolved that.

i0rek:
ok, I am at a point where I addressed the feedback except these two points: 

1) **protocol version:** there is no point checking for version 3 - because not all v3 raft has leadership transfer. should we create v4 for it?
2) **which actions need to be blocked?** The reason why I did it this way is the paper says `1. The prior leader stops accepting new client requests.` which I interpreted as every request that comes from the API. Since only the leader can transfer leadership, and the leader initiates every action, I think what I did is enough. I might be wrong.

raft.go
banks: Might be nice to keep a comment in the code to link to this for when we come back later and wonder why we didn't follow the thesis on that?

i0rek:
In a call with @banks we decided, that we don't cut a new version. I will make it clear in the docs that you have to run the latest code in order to use leadership transfer - that v3 is not sufficient.

With regards to blocking actions coming through `commitCh`, I don't think we should do that. The commitCh receives a message when an entry is commited: accepted by the majority of the followers. This doesn't change the raft state. But it changes the state of the leader. However even when we chose to not process these messages, we are basically missing stuff since we are still the leader. If we happen to transfer to a follower that has this commit, everything is fine. If we transfer to a follower that doesn't, it won't win the election. Not processing commitCh doesn't change that.

i0rek:
If a leader attempts to transfer leadership to a follower that doesn't have that feature, it will fail gracefully on the follower and the leader. In that case the old leader keeps being the leader:

```
    2019/02/14 00:17:16 [ERR] raft-net: Failed to decode incoming command: unknown rpc type 3
```

and on the leader as well:

```
    2019/02/14 00:17:16 [WARN] consul: failed to transfer leadership: failed to make TimeoutNow RPC to dd9a2744-96ec-78fc-f69c-d7376e207946: EOF
```

raft.go
```diff
@@ -505,6 +541,43 @@ func (r *Raft) leaderLoop() {
...
+
+			// Track leadership status with the notify channel.
+			verify := &verifyFuture{}
+			verify.init()
+			r.leaderState.notify[verify] = struct{}{}
+			stopCh := make(chan error, 1)
+			doneCh := make(chan error, 1)
+
+			go func() {
+				select {
```
i0rek: also read from `shutdownCh` to make sure we don't hang around longer than we need to.

api.go
```diff
@@ -1006,3 +1021,34 @@ func (r *Raft) LastIndex() uint64 {
...
+
+// LeadershipTransfer will transfer leadership to a server in the cluster.
+// This can only be called from the leader, or it will fail. The leader will
+// stop accepting client requests, make sure the target server is up to date and
+// starts the transfer with a TimeoutNow message. This message has the same
+// effect as if the election timeout on the on the target server fires. Since it
+// is unlikely that no other server is starting an election, it is very likely
+// that the target server is able to win the election.
+func (r *Raft) LeadershipTransfer() Future {
+	if r.protocolVersion < 3 {
```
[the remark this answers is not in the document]
mkeeler: For the sake of completeness, the answer to my question is that we don't currently support auto-negotiation of Raft protocol versions. So its not so simple to upgrade existing clusters.

raft.go
```diff
@@ -77,12 +77,14 @@ type commitTuple struct {
...
-	replState  map[ServerID]*followerReplication
-	notify     map[*verifyFuture]struct{}
-	stepDown   chan struct{}
+	commitCh                     chan struct{}
+	commitment                   *commitment
+	inflight                     *list.List // list of logFuture in log index order
+	replState                    map[ServerID]*followerReplication
+	notify                       map[*verifyFuture]struct{}
+	stepDown                     chan struct{}
+	lease                        <-chan time.Time
```
banks: Do we need to move this any more? I think you did it because originally you were waiting on it in the transfer until I pointed out that wasn't necessary.

I don't think it matters much so happy to leave it if that's easier but not sure it's necessary to pull  it out of the leader loop method any more?

raft.go
```diff
@@ -341,6 +356,8 @@ func (r *Raft) runLeader() {
 	r.leaderState.replState = make(map[ServerID]*followerReplication)
 	r.leaderState.notify = make(map[*verifyFuture]struct{})
 	r.leaderState.stepDown = make(chan struct{}, 1)
+	r.leaderState.lease = time.After(r.conf.LeaderLeaseTimeout)
```
banks: Hmm is this right? It seems to be e remnant of the original lease stuff that turned out not to be relevant?

raft.go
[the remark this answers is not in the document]
banks: For future reference the reason this is fine is that `commitCh` is used to be notified by replication code when a log has been acked by enough followers to be considered committed. Allowing this still is important in case the leadership transfer doesn't work and we are still leader but dropped the ball on tracking which logs got committed by followers.

raft_test.go
```diff
@@ -2395,6 +2406,275 @@ func TestRaft_ProtocolVersion_Upgrade_2_3(t *testing.T) {
...
+	t.Skip("How do I test this?")
+}
+
+func TestRaft_LeadershipTransferToUnresponsiveServer(t *testing.T) {
+	t.Skip("How do I test this?")
+}
+
+func TestRaft_LeadershipTransferResetsLeaderLease(t *testing.T) {
+	t.Skip("How do I test this?")
+}
```
banks: Shouldn't even be a test any more with lease stuff removed?

raft_test.go
```diff
@@ -2395,6 +2406,275 @@ func TestRaft_ProtocolVersion_Upgrade_2_3(t *testing.T) {
...
+
+func TestRaft_LeadershipTransferReplicationFails(t *testing.T) {
+	t.Skip("How do I test this?")
+}
+
+func TestRaft_LeadershipTransferToUnresponsiveServer(t *testing.T) {
+	t.Skip("How do I test this?")
+}
```
banks: Should we just remove these if we have no way to test them?

raft.go
```diff
@@ -1468,3 +1632,71 @@ func (r *Raft) setState(state RaftState) {
...
+			continue
+		}
+		state, ok := r.leaderState.replState[server.ID]
+		if !ok {
+			continue
+		}
+		// return early if this server is up to date
+		if state.nextIndex > target {
+			return &server
+		}
```
banks: Is this necessary?

The loop isn't long and would always find the most up-to-date anyway so I don't think it's a big performance win. In some cases it might even work out to be less-than-optimal for transfer time because there may have been 10 new logs recorded since our `getLastIndex` above and the first server is all 10 behind while another server in the list might be only 1 behind.

raft.go
```diff
@@ -1468,3 +1632,71 @@ func (r *Raft) setState(state RaftState) {
...
+// pickServer returns the follower that is most up to date.
+func (r *Raft) pickServer() *Server {
+	target := r.getLastIndex()
+	var pick *Server
+	var current uint64
+	for _, server := range r.configurations.latest.Servers {
+		if server.ID == r.localID {
+			continue
+		}
+		state, ok := r.leaderState.replState[server.ID]
```
banks: Is this OK? In general leaderState is expected to only be accessed by the single goroutine running the leader loop. Does this always run there? I think it does but worth checking. If so I suggest we update the comments on these new methods that assume they are part of the leader loop to make that explicit?

james-lawrence:
> I should also ask - the failing CI tests, are those same tests failing in master and/or passing locally?

@banks, there are a bunch of minor fixes (by freeekanayaka) in PR form already that improve the CI stability of master. might want to consider merging those and rebasing on top.

raft_test.go
i0rek: agreed!

raft_test.go
i0rek: agreed!

raft.go
i0rek: This is accounted for because we have

```
defer func() { close(leftLeaderLoop) }()
```

which will close leftLeaderLoop and trigger the case.

raft.go
i0rek: Great catch, it wasn't called from the leaderloop! Plus I found another place where `leaderstate` is accessed outside of the `leaderloop`. Fixed both things.

i0rek: https://github.com/hashicorp/raft/pull/306/commits/20a87dbeaa9522f3e5305dc1fe2c705a8ed22fbb

raft.go
i0rek: Agreed. Fixed in https://github.com/hashicorp/raft/pull/306/commits/8deca55fc7984ab0a0dd85b9ab29bccdc6711b6e.

raft.go
```diff
@@ -1484,3 +1664,66 @@ func (r *Raft) setState(state RaftState) {
...
+func (r *Raft) lookupServer(id ServerID) *Server {
+	for _, server := range r.configurations.latest.Servers {
+		if server.ID != r.localID {
+			return &server
+		}
+	}
+	return nil
+}
+
+// pickServer returns the follower that is most up to date.
```
banks: Should we comment this is only to be called in leader loop?

raft.go
```diff
@@ -506,6 +543,74 @@ func (r *Raft) leaderLoop() {
...
+				case err := <-doneCh:
+					if err != nil {
+						r.logger.Debug(err.Error())
+					}
+					future.respond(err)
+				}
+			}()
+
+			// leaderState.replState is accessed here before
+			// starting leadership asynchronously transfer because
```
banks:
```suggestion
			// starting leadership transfer asynchronously because
```

raft.go
```diff
@@ -665,6 +790,54 @@ func (r *Raft) verifyLeader(v *verifyFuture) {
...
+		doneCh <- nil
+		return
+	default:
+	}
+
+	// Step 1: set this field which stops this leader from responding to any client requests.
+	r.setLeadershipTransferInProgress(true)
+	defer func() { r.setLeadershipTransferInProgress(false) }()
+
+	for repl.nextIndex <= r.getLastIndex() {
```
banks: This access of the `followerReplication.nextIndex` made me wonder about thread safety. Presumably this is racey since the replication thread is still running and could be writing to the `nextIndex` concurrently?

raft.go
i0rek: done!

raft.go
i0rek: Using `atomic` now to protect from the race. Great catch!

PR #308, opened 2019-02-24
checkLeaderLease: count only voters into quorum

santhosh-tekuri:
currently checkLeaderLease is including all nodes.
but it should include only voters into quorum.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -844,27 +844,34 @@ func (r *Raft) leadershipTransfer(id ServerID, address ServerAddress, repl *foll
 // contact. This must only be called from the main thread.
 func (r *Raft) checkLeaderLease() time.Duration {
 	// Track contacted nodes, we can always contact ourself
-	contacted := 1
+	contacted := 0
 
 	// Check each follower
 	var maxDiff time.Duration
 	now := time.Now()
-	for peer, f := range r.leaderState.replState {
-		diff := now.Sub(f.LastContact())
-		if diff <= r.conf.LeaderLeaseTimeout {
-			contacted++
-			if diff > maxDiff {
-				maxDiff = diff
+	for _, server := range r.configurations.latest.Servers {
+		if server.Suffrage == Voter {
+			if server.ID == r.localID {
+				contacted++
+				continue
 			}
-		} else {
-			// Log at least once at high value, then debug. Otherwise it gets very verbose.
-			if diff <= 3*r.conf.LeaderLeaseTimeout {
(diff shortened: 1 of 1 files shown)


santhosh-tekuri:
travis build is failing because of dependencies:

```
net_transport_test.go:6:2: cannot find package "github.com/stretchr/testify/require" in any of:
	/home/travis/.gimme/versions/go1.8.linux.amd64/src/github.com/stretchr/testify/require (from $GOROOT)
	/home/travis/gopath/src/github.com/stretchr/testify/require (from $GOPATH)
```

rboyer:
I think if you rebased it on `master` now that https://github.com/hashicorp/raft/pull/309 is merged the unrelated test failure you were seeing should go away.

santhosh-tekuri:
trying to trigger travis build

santhosh-tekuri:
trying to trigger travis build

santhosh-tekuri:
anybody there? it has been 12 days since I submit this pull request

RobbieMcKinstry:
Hi @santhosh-tekuri ! I'm arguing with @s-christoff about this, but I think we found some agreement on this. Going to get another pair of eyes to confirm our logic is correct, since I'm fairly new to Raft.

I suspect that excluding non-voters from the `checkLeaderLease` quorum will make this check less robust.
Consider this (pathological) scenario where you have 5 voters and 45 nonvoters. Now suppose that a network partition occurs and the leader is able to get a quorum of voters but a bulk of non-voters are on the other partition. Since most of the nodes in the cluster are unable to receive replication, the leader should step down with the hope that the new leader is unaffected by the partition (i.e. overlaps the two partitions). Otherwise, you run the risk of having the quorum of voting notes collecting a large number of logs, none of which are able to be replicated to the bulk of the nodes in the cluster (as the replicants are on the other partition). (Though I suppose the leader will be taken down anyway when the other partition comes back online with a higher term).

santhosh-tekuri:
Hi,

current logic of checkLeaderLease is wrong. suppose quorum is 3. if leader
is able to contact 3 non-voters only(with no other voters) it still thinks
that it can extend the lease.

thanks
Santhosh

s-christoff:
Howdy @santhosh-tekuri , good catch! We agree. Could you please rebase? It should be good to go!

santhosh-tekuri:
@s-christoff rebased with master. 

travis seems failed because of some data-race issue

s-christoff:
Hey @santhosh-tekuri  - thank you for this contribution!

PR #309, opened 2019-02-25
ensure 'make deps' downloads test dependencies like testify

rboyer:


rboyer:
@i0rek I think PR #308 is hitting this, and I noticed in your leadership transfer branch that you mitigated this a little bit differently.

PR #310, opened 2019-03-04
Fixes #282

### LINKED ISSUE #282: What is this?
So what is this? Is it really needed? https://github.com/hashicorp/raft/blob/v1.0.0/replication.go#L133

s-christoff:
Clarifies function of `CommitTimeout`

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -137,7 +137,8 @@ RPC:
 		case <-s.triggerCh:
 			lastLogIdx, _ := r.getLastLog()
 			shouldStop = r.replicateTo(s, lastLogIdx)
-		case <-randomTimeout(r.conf.CommitTimeout): // TODO: what is this?
+		// Check and update leaderCommit of followers
+		case <-randomTimeout(r.conf.CommitTimeout):
 			lastLogIdx, _ := r.getLastLog()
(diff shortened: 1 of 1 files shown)


replication.go
```diff
@@ -137,7 +137,8 @@ RPC:
 		case <-s.triggerCh:
 			lastLogIdx, _ := r.getLastLog()
 			shouldStop = r.replicateTo(s, lastLogIdx)
-		case <-randomTimeout(r.conf.CommitTimeout): // TODO: what is this?
+		// Check and update leaderCommit of followers
```
banks: Up to you on how verbose you want to be here but I'd probably go with something more like:

```suggestion
		// This is _not_ our heartbeat mechanism but is to ensure
		// followers quickly learn the leader's commit index when 
		// raft commits stop flowing naturally. The actual heartbeats 
		// can't do this to keep them unblocked by disk IO on the 
		// follower. See https://github.com/hashicorp/raft/issues/282.
```

s-christoff: This is definitely more clear, which is what I strive for. I was just concerned by adding *too much* comments, but I like this way better. Thank you :)

Issue #312, opened 2019-03-19
Raft.Shutdown() with InmemTransport hangs after leader changed


babbageclunk:
We sometimes see the [Juju](https://github.com/juju/juju) raft tests timeout because Raft.Shutdown has hung. 

I've got a standalone reproduction (not that small though) that hangs reliably. Digging into it the InmemTransport only applies its timeout on receiving responses back from the peer, but not sending them.

```go
package main

import (
	"fmt"
	"io"
	"log"
	"os"
	"sync"
	"time"

	"net/http"
	_ "net/http/pprof"

	"github.com/hashicorp/raft"
)

func main() {
	go func() {
		log.Println(http.ListenAndServe("localhost:6060", nil))
	}()

	syncStderr := &syncWriter{target: os.Stderr}
	r0, t0, err := newRaft("node-0", syncStderr)
	if err != nil {
		panic(err)
	}

	f := r0.BootstrapCluster(raft.Configuration{
		Servers: []raft.Server{{
			ID:      "node-0",
			Address: t0.LocalAddr(),
		}},
	})
	if err := f.Error(); err != nil {
		panic(err)
	}

	// Wait for r0 to be leader.
	select {
	case isLeader := <-r0.LeaderCh():
		if !isLeader {
			panic("r0 wasn't leader")
		}
	case <-time.After(time.Second):
		panic("timed out waiting for r0 to be leader")
	}

	// Make two more and add them to the cluster.
	r1, t1, err := newRaft("node-1", syncStderr)
	if err != nil {
		panic(err)
	}
	r2, t2, err := newRaft("node-2", syncStderr)
	if err != nil {
		panic(err)
	}
	connectTransports(t0, t1, t2)
	f1 := r0.AddVoter("node-1", t1.LocalAddr(), 0, 0)
	f2 := r0.AddVoter("node-2", t2.LocalAddr(), 0, 0)
	if err := f1.Error(); err != nil {
		panic(err)
	}
	if err := f2.Error(); err != nil {
		panic(err)
	}

	// Shut the leader down.
	if err := r0.Shutdown().Error(); err != nil {
		panic(err)
	}
	if err := t0.Close(); err != nil {
		panic(err)
	}

	// Wait until one of the other nodes is leader.
	var newLeader *raft.Raft
	var name string
loop:
	for {
		select {
		case value := <-r1.LeaderCh():
			if value {
				newLeader = r1
				name = "node-1"
				break loop
			}
		case value := <-r2.LeaderCh():
[...]

PR #313, opened 2019-03-19
InmemTransport: respect timeout when sending

babbageclunk:
InmemTransport would block indefinitely when trying to make an RPC to a peer who has already shut down. This was causing some of [Juju](https://github.com/juju/juju)'s tests for our raft machinery to hang sometimes when stopping raft nodes - see #312 for code that will reproduce the situation. 

Guarding the send with a timeout (like there is on the receive) fixes it and seems like the right thing to do here.

### DIFF
--- a/inmem_transport.go
+++ b/inmem_transport.go
@@ -147,10 +147,16 @@ func (i *InmemTransport) makeRPC(target ServerAddress, args interface{}, r io.Re
 
 	// Send the RPC over
 	respCh := make(chan RPCResponse)
-	peer.consumerCh <- RPC{
+	req := RPC{
 		Command:  args,
 		Reader:   r,
 		RespChan: respCh,
+	}
+	select {
+	case peer.consumerCh <- req:
+	case <-time.After(timeout):
+		err = fmt.Errorf("send timed out")
+		return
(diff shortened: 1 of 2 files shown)


babbageclunk:
Hmm - I think the integration test that failed is flaky. If I run it in a loop against master it fails within 10 runs.

api.go
```diff
@@ -437,17 +437,17 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
...
-		conf:            *conf,
-		fsm:             fsm,
-		fsmMutateCh:     make(chan interface{}, 128),
-		fsmSnapshotCh:   make(chan *reqSnapshotFuture),
-		leaderCh:        make(chan bool),
-		localID:         localID,
-		localAddr:       localAddr,
-		logger:          logger,
-		logs:            logs,
+		protocolVersion:       protocolVersion,
```
babbageclunk: I didn't mean to include this formatting change, the file got gofmt'd when I was adding debugging.

s-christoff:
Thanks for fixing this odd timing bug in our testing system!

PR #317, opened 2019-04-17
Go modules and use ugorji/go/codec instead of hashicorp/go-msgpack/codec

mkeeler:
Moved the fuzzy package into github.com/hashicorp/raft-fuzzy.

Ran `go mod init` and `go mod tidy`

For the msgpack update I ran

`goimport-rewrite -r github.com/hashicorp/go-msgpack/codec:github.com/ugorji/go/codec .`

There were a handful of places our usage of codec needed some updating to reflect that the decode options got put into another struct within the msgpack handle where they used to be at the top level.


jefferai:
Seems reasonable, with the same comment I made on the other package about the 5 years of changes to codec.

mkeeler:
After getting all these open and to a point I could actually test out Consul I did find some incompatibilities in encodings of the old to new library. For now I have shelved the update and might take another pass at fixing up message passing codecs at some point in time.

Issue #320, opened 2019-04-26
Leveled Logging of Events


Lz-Gustavo:
Similar to #272 ,
I was thinking that would be good for Observability to implement a leveled log of events, since every log output to stderr is invoked like this:

    `r.logger.Printf("[WARN] raft: no known peers, aborting election")`

I know that I can use it by providing a user-defined log to config struct and implement the logger interface to my own needs, but I think a more transparent approach would be a better idea. Being able to simply pass a log level to config, let's say "WARN", and it would only print entries of similar or higher priority (i.e. "WARN", "ERROR" and "FATAL"), greatly reducing interrupts to output unnecessary information in some cases.

I would like to know if there's any interested on this idea? I'll probably gonna implement that on my application some way or another, I was wondering if that's a necessary pull request.

Of course it would add a new layer of abstraction to every logger invocation, so a good performance log library is a must. I was thinking about [Logrus](https://github.com/sirupsen/logrus) or [log15](https://github.com/inconshreveable/log15), do you have any other suggestions?

RobbieMcKinstry:
Hi @Lz-Gustavo ! Thanks for opening this. Leveled logging is definitely an upgrade we'd consider merging. Our library of choice for leveled logging is [hc-log](https://github.com/hashicorp/go-hclog). 

One note though. Consul (which uses hashicorp/raft) hasn't yet upgraded to hc-log, so we would just have to check that the API would be compatible before merging.

Does this give you sufficient information to answer your question? Thank you again!

Lz-Gustavo:
Yea of course it does, thanks @RobbieMcKinstry for the suggestions! I'll definitely look forward about hc-log.

RobbieMcKinstry:
Closed by https://github.com/hashicorp/raft/pull/321

Issue #272, opened 2018-02-05
Improve observability by adding opentracing and/or structured logging


freeekanayaka:
The current use of stdlib's log.Logger for emitting events is not very convenient with regard to observability/tracing.

Some ideas for improving this aspect would be:

- Add [opentracing](/opentracing/opentracing-go) support (e.g. by adding a Config.Tracer configuration option and instrumenting the code accordingly)

- Replace stdlib's log.Logger with [hclog.Logger](/hashicorp/go-hclog) (backward-compatibility with Config.Logger and Config.LogOutput should not be too hard)

- Add more events that can be observed with raft.Observer (as per #249)

Is there interest for any of the above? I could work on that if desired.

RobbieMcKinstry:
Hi @freeekanayaka ! This issue resurfaced in https://github.com/hashicorp/raft/issues/320 , but I wanted to add a little context over here for clarity. Your right, the lib could benefit from increased traceability. Leveled logging would definitely help.  The second option you provide is perfect I think: we'd greatly appreciate a PR to use hclog.Logger as long as it's backwards-compatible.
Thank you for your insights. I'll have to think more about the other two suggestions as well.

freeekanayaka:
Hello @RobbieMcKinstry, unfortunately I'm not using hashicorp/raft anymore, because I had to switch to a C implementation of raft (see [here](https://github.com/CanonicalLtd/raft)). So I'll have to backtrack on this proposal, hope somebody else can pick it up.

RobbieMcKinstry:
Thanks for letting me know! Sad to see you go, but that's looks like a cool C implementation!

RobbieMcKinstry:
Closing this issue as it was resolved by https://github.com/hashicorp/raft/pull/321

PR #321, opened 2019-04-29
Leveled log

Lz-Gustavo:
Trying to fix some observability issues #272 , #320 

This PR excludes support to a generic user-defined log implementing sdt Logger interface, to support hashicorp's [hclog](https://github.com/hashicorp/go-hclog) Logger interface. Allowing both the easy configuration of a desired level restriction to Config struct, like bellow:

`config := raft.DefaultConfig( )
config.LogLevel = "WARN"
raft, err := raft.NewRaft( config, ... ) `

And the configuration of an implemented hclog.Logger interface.

### DIFF
--- a/config.go
+++ b/config.go
@@ -3,8 +3,9 @@ package raft
 import (
 	"fmt"
 	"io"
-	"log"
 	"time"
+
+	"github.com/hashicorp/go-hclog"
 )
 
 // These are the versions of the protocol (which includes RPC messages as
@@ -190,9 +191,13 @@ type Config struct {
 	// Defaults to os.Stderr.
 	LogOutput io.Writer
 
-	// Logger is a user-provided logger. If nil, a logger writing to LogOutput
-	// is used.
-	Logger *log.Logger
+	// LogLevel represents a log level. If a no matching string is specified,
+	// hclog.NoLevel is assumed.
+	LogLevel string
+
+	// Logger is a user-provided hc-log logger. If nil, a logger writing to
+	// LogOutput with LogLevel is used.
+	Logger hclog.Logger
 }
 
 // DefaultConfig returns a Config with usable defaults.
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -208,6 +213,7 @@ func DefaultConfig() *Config {
 		SnapshotInterval:   120 * time.Second,
 		SnapshotThreshold:  8192,
 		LeaderLeaseTimeout: 500 * time.Millisecond,
+		LogLevel:           "TRACE",
```
RobbieMcKinstry: Do we want TRACE to be the default level? I suspect INFO is more reasonable, and at very least since we don't have any messages below debug, we should set it to debug.

Lz-Gustavo: I totally agree. When I first set it to TRACE I was thinking about preserving every old logger output. As you said, since we don't actually have any messages below debug, it's more reasonable setting it to DEBUG.

RobbieMcKinstry:
Hey @Lz-Gustavo, thank you so much for this contribution! 
It looks like some tests still need to be updated to use the correct logger:

``` 
./integ_test.go:114:14: cannot use env.logger (type *log.Logger) as type hclog.Logger in assignment:
    *log.Logger does not implement hclog.Logger (missing Debug method)
./raft_test.go:78:14: cannot use newTestLogger(t) (type *log.Logger) as type hclog.Logger in assignment:
    *log.Logger does not implement hclog.Logger (missing Debug method)
./raft_test.go:631:19: cannot use newTestLoggerWithPrefix(t, string(configuration.Servers[i].ID)) (type *log.Logger) as type hclog.Logger in assignment:
    *log.Logger does not implement hclog.Logger (missing Debug method)
```

We have been discussing about replacing the usage of `Sprintf` with using hc-log with key/value pairs instead, but we think that should be a separate PR in case this produces any breaking changes downstream.

Lz-Gustavo:
I'm sorry, I've forgot about test_files on the first push. Now they're all running smoothly.

About Sprintf usage, I first though about using hclog.Fmt() with it's own key/value pairs, but that would result in some small modifications on the logging output string. I tried to preserve the same standard string format.

RobbieMcKinstry:
Perfect, thank you very much!

RobbieMcKinstry:
Thank you very much for your contribution @Lz-Gustavo ! This is a much needed boost to observability!

PR #322, opened 2019-04-29
Introduce GetConfig

### LINKED ISSUE #298: Obtaining peers before starting Raft processing
In RobustIRC (which I’m trying to migrate from raft v0 to raft v1), we verify that the clock of the node which is about to join the network is synchronized (well enough) with the remaining nodes: https://github.com/robustirc/robustirc/blob/38fd78c7f0b39fe25a78d33805b89b26617363c3/robustirc.go#L398-L417

This ensures that accidental clock drift doesn’t e.g. expire all active sessions or prevent new messages from being accepted due to incorrect throttling state.

To verify the clocks, the node needs to know the server addresses of the other nodes in the network *before starting Raft processing*. Up until now, we have used raft.NewJSONPeers, but that no longer exists. I can’t see an exported function yet for obtaining the configuration from the combination of snapshot+log.

How can callers accurately obtain the most recent configuration? If there is no way currently, would you accept a PR which adds such a function?

Thanks!

stapelberg:
fixes #298

### DIFF
--- a/api.go
+++ b/api.go
@@ -383,6 +383,18 @@ func HasExistingState(logs LogStore, stable StableStore, snaps SnapshotStore) (b
 	return false, nil
 }
 
+// GetConfig returns the latest configuration based on snapshots and logs,
+// without actually starting a Raft node. This can be useful to obtain the
+// server addresses to do validation before joining the network.
+func GetConfig(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps SnapshotStore, trans Transport) (Configuration, error) {
+	conf.NoBackgroundWork = true
+	r, err := NewRaft(conf, fsm, logs, stable, snaps, trans)
+	if err != nil {
+		return Configuration{}, err
+	}
+	return r.configurations.Clone().latest, nil
+}
+
 // NewRaft is used to construct a new Raft node. It takes a configuration, as well
 // as implementations of various interfaces that are required. If we have any
 // old state, such as snapshots, logs, peers, etc, all those will be restored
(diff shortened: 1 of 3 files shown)


stapelberg:
Possibly we can further restrict the signature of GetConfig, but this was the simplest possible thing I could send out for review to confirm the overall direction before going down the rabbit hole too deeply :)

Please let me know what you think,

s-christoff:
Hey @stapelberg, thank you so much for your contribution!  
I think it would be great to have a generic way to recreate the entire FSM and config from disk without starting Raft. In [restoreCluster](https://github.com/hashicorp/raft/blob/2c551690b5c0eb05ef7f4ad72ed01f7f6ce3fcb6/api.go#L842), we already do something very similar to this. I think if we can use `restoreCluster` as a base and refactor it to take out the portions that satisfies this need we can have a more generalized function that helps more users. 
Please let me know your thoughts on this and I would love to help in any way!

stapelberg:
> Hey @stapelberg, thank you so much for your contribution!

Thanks for taking a look!

> I think it would be great to have a generic way to recreate the entire FSM and config from disk without starting Raft. In [restoreCluster](https://github.com/hashicorp/raft/blob/2c551690b5c0eb05ef7f4ad72ed01f7f6ce3fcb6/api.go#L842), we already do something very similar to this. I think if we can use `restoreCluster` as a base and refactor it to take out the portions that satisfies this need we can have a more generalized function that helps more users.
> Please let me know your thoughts on this and I would love to help in any way!

Sorry, I’m not very familiar with the internals of this package, so I don’t really follow. Would you like to take my change and adjust it as you see fit, or merge it and do a follow-up change, or just do a different change entirely that covers my use-case? I don’t care much about how we get there, just about the end result :) 

Thanks,

stapelberg:
@s-christoff Friendly ping? :)

stapelberg:
How can we move this forward? I still can’t update my Raft network to a more recent version of the package :(

s-christoff:
Hey, we updated this in #369!  :)
Thank you for your patience with this.

PR #323, opened 2019-04-30
Add go mod

RobbieMcKinstry:
This PR adds Go modules. It also bumps our testing version on Travis to include v1.12.


go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/hashicorp/raft
+
+go 1.12
+
+require (
+	github.com/armon/go-metrics v0.0.0-20190430140413-ec5e00d3c878
+	github.com/boltdb/bolt v1.3.1 // indirect
+	github.com/hashicorp/go-hclog v0.9.0
+	github.com/hashicorp/go-msgpack v0.5.4
```
briankassouf: go-msgpack has had a release since this PR, might want to re-init so it pulls in the latest version

RobbieMcKinstry: Thanks!

RobbieMcKinstry:
@briankassouf  I nuked the `go.mod` and `go.sum` files and re-inited with `go mod init` as suggested.

RobbieMcKinstry:
I'm going to go ahead and merge this. Sorry if you think it needs another tweak. I added an environment variable to Travis so it uses `go mod` on relevant versions. This is causing an error on other builds, since they require `tip`, but was necessary to test this PR works.

Merging this PR will fix the other builds, which is why I'm going to do it. Worst case scenario we can revert the commit or hose these 2 files.

Issue #299, opened 2018-11-02
Guidance on chosing a ServerID


stapelberg:
The documentation states “ServerID is a unique string identifying a server for all time.”

The documentation raft.ProtocolVersion states “Update the server's config to a better ID”.

Given that I’m currently using ServerAddresses as IDs, I’m wondering how a “better” ID should be chosen, and what makes it better. In my mode of operation, the network address is permanent — if the network address has to change, the node is removed and re-added to the cluster.

Given the above, are there any downsides to always setting ServerID to ServerAddress?

s-christoff:
Hey @stapelberg, thank you for bringing this up! 

In the documentation for `raft.ProtocolVersion` the statement “Update the server's config to a better ID” is geared towards people moving from ProtocolVersion 0 -> 1. In the older protocol versions the ServerID was set to the Network Address by default. However with the implementation ServerID it sets the ServerID to a UUID and persists the ID alongside your other raft data. I can see how this could be a bit confusing and I'll be updating the documentation to reflect more clearly. 

The only downside I could foresee happening is a node leaving the cluster and trying to rejoin the cluster with the same address without the removal completing. However, the likelihood of this happening may be very slim dependent on how your ServerAddresses are being generated.

stapelberg:
Thanks, this is much clearer. For my particular use-case (https://robustirc.net/), the server addresses are handed out to end users and hence have to remain stable anyway, so I’ll just stick with server id == server address.

PR #324, opened 2019-05-02
Update Docs to remove "better" #299

s-christoff:
Removing the usage of "better" to be more explicit in stating we use the ServerAddress instead of the network address.


RobbieMcKinstry:
LGTM

PR #325, opened 2019-05-02
Add Linter

s-christoff:
- GoFmt applied to all of Raft
- Travis now runs against Go 1.12 

🎉🎉🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,13 @@ go:
     # - 1.6
     - 1.8
     - 1.9
+    - 1.12
     - tip
 
-install: make deps
+install: 
+    - make deps 
+    - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin latest 
+
 script:
     - make integ
 
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,7 @@ GO_VERSION  = $(shell go version)
 
 # Look for versions prior to 1.10 which have a different fmt output
 # and don't lint with gofmt against them.
-ifneq (,$(findstring go version go1.7, $(GO_VERSION)))
-	FMT=
-else ifneq (,$(findstring go version go1.8, $(GO_VERSION)))
+ifneq (,$(findstring go version go1.8, $(GO_VERSION)))
 	FMT=
 else ifneq (,$(findstring go version go1.9, $(GO_VERSION)))
 	FMT=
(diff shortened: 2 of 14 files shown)


.travis.yml
```diff
@@ -5,9 +5,13 @@ go:
     # - 1.6
     - 1.8
     - 1.9
+    - 1.12
     - tip
 
-install: make deps
+install: 
+    - make deps 
+	- curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin latest
```
RobbieMcKinstry: This indent is curious

RobbieMcKinstry: Should be equally aligned

Makefile
```diff
@@ -9,12 +9,19 @@ integ: test
 fuzz:
 	go test -timeout=300s ./fuzzy
 	
-deps:
+deps: dep-linter
 	go get -t -d -v ./...
 	echo $(DEPS) | xargs -n1 go get -d
 
+lint:
```
RobbieMcKinstry: lint should be added a dependency of the `integ` rule on line 6

Makefile
```diff
@@ -9,12 +9,19 @@ integ: test
...
+	gofmt -s -w .
+	golangci-lint run -c .golangci-lint.yml
+
+dep-linter:
+	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin latest
```
RobbieMcKinstry: This line still doesnt work. TIL In Make `$()` evaluates a variable instead of running a subshell. So its looking for the variable named `go env GOPATH` which returns the empty string.

api.go
```diff
@@ -243,13 +243,14 @@ func BootstrapCluster(conf *Config, logs LogStore, stable StableStore,
 // the usual APIs in order to bring the cluster back into a known state.
 func RecoverCluster(conf *Config, fsm FSM, logs LogStore, stable StableStore,
 	snaps SnapshotStore, trans Transport, configuration Configuration) error {
+	var err error
```
RobbieMcKinstry: So, with all these new predeclarations, I wonder if there's a clean way to hoist them up to the top of the function...

net_transport_test.go
```diff
@@ -76,8 +76,8 @@ func TestNetworkTransport_CloseStreams(t *testing.T) {
 		t.Fatalf("err: %v", err)
 	}
 	defer trans2.Close()
-
-	for i := 0; i < 2; i++ {
+	var i int
```
RobbieMcKinstry: Yeah we should definitely do a refactoring pass before we merge. This is  a bug waiting to happen IMHO

RobbieMcKinstry:
As per our meeting, we're going to break this into two PRs: one for the refactoring of the `RecoverCluster` fn, and another for the linter.

s-christoff:
Moved into two new branches - #329 , #328

PR #327, opened 2019-05-10
chore: Add probot

endocrimes:
This is a first draft at a GitHub Probot configuration for managing those issues
with a long enough cycle time that it shouldn't catch cases where there is
active discussion, and it'll ignore anything we've explicitly called out an
enhancement, bug, or docs change, but will prompt people for a response if
we're waiting-reply etc.

I'm especially looking for feedback on the language usage, because I think the
thresholds are about correct for the volume on this repo.

Link to source/docs: https://github.com/probot/stale

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -20,8 +20,10 @@ staleLabel: "waiting-reply"
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: |
   Hey there,
-  We wanted to check in on this request since it has been inactive for 90 days. 
-  Are you having the same issue with the latest version of Raft? Have you reviewed the latest [godocs](https://godoc.org/github.com/hashicorp/raft)? 
+  We wanted to check in on this request since it has been inactive for at least 90 days.
+  Have you reviewed the latest [godocs](https://godoc.org/github.com/hashicorp/raft)? 
+  If you think this is still an important issue in the latest version of [the Raft library](https://github.com/hashicorp/raft/compare/) or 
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,38 @@
...
+  - "thinking"
+  - "docs"
+
+# Label to use when marking as stale
+staleLabel: "waiting-reply"
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: |
+  Hey there,
+  We wanted to check in on this request since it has been inactive for 90 days.
```
banks: should we say "at least 90 days"? some will be way older initially and who knows how tight the timing is? Not a big deal though.

.github/stale.yml
```diff
@@ -0,0 +1,38 @@
...
+  - "docs"
+
+# Label to use when marking as stale
+staleLabel: "waiting-reply"
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: |
+  Hey there,
+  We wanted to check in on this request since it has been inactive for 90 days. 
+  Are you having the same issue with the latest version of Raft? Have you reviewed the latest [godocs](https://godoc.org/github.com/hashicorp/raft)?
```
banks: This could sound to some a little too forward like we are expecting users to have been keeping up to date.

Perhaps we could word it more like:

> If you think this is still an important issue in the latest version of [the raft library](...) or [its documentation](...) please feel let us know what you are seeing now and we'll keep it open for investigation.

PR #328, opened 2019-05-13
Add Linter 

s-christoff:
Add Linter to only run on Go versions > 1.11

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,7 @@ GO_VERSION  = $(shell go version)
 
 # Look for versions prior to 1.10 which have a different fmt output
 # and don't lint with gofmt against them.
-ifneq (,$(findstring go version go1.7, $(GO_VERSION)))
-	FMT=
-else ifneq (,$(findstring go version go1.8, $(GO_VERSION)))
+ifneq (,$(findstring go version go1.8, $(GO_VERSION)))
 	FMT=
 else ifneq (,$(findstring go version go1.9, $(GO_VERSION)))
 	FMT=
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -16,12 +29,19 @@ ci.integ: ci.test
...
+deps: dep-linter
 	go get -t -d -v ./...
 	echo $(DEPS) | xargs -n1 go get -d
 
+lint:
+	gofmt -s -w .
+	golangci-lint run -c .golangci-lint.yml $(FMT) .
+
+dep-linter:
+	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(ENV)/bin v1.16.0
```
i0rek: I am always a little hesitant to pipe from curl to sh in that way. Very easy to get screwed that way. Especially since this would be done for every `deps` run.

RobbieMcKinstry: @i0rek Would you prefer I curl a known release on GH and checksum it? Sorry, I've never scripted package management the right way before, I don't know what I'm doing.

alvin-huang: Would just `go get`ting github.com/golangci/golangci-lint/cmd/golangci-lint work?

alvin-huang: Just took a look at their docs and what you have is their recommendation, I am fine with it. It might be easier to bubble the version up to the top of the Makefile for visibility and I would suggest decoupling `dep-linter` from `deps`. `Deps` is more for go dependencies whereas dep-linter is just installing this binary so it may be semi confusing if I just want to build and may never want to run `make lint` to have a binary pulled all the time.

PR #334, opened 2019-05-23
Add circle config

alvin-huang:
I removed the test for go1.10 since it does not have the `go mod` subcommands and would need to go back to the vendor route. Not hugely difficult but I can add that back later if we want to support it.


s-christoff:
🥇

PR #338, opened 2019-06-07
Clarify the BootstrapCluster behaviour and error expectation

### LINKED ISSUE #336: Potentially misleading documentation in BootstrapCluster method
I'm observing different behaviour than the documented behaviour for [`func (*Raft) BootstrapCluster`](https://godoc.org/github.com/hashicorp/raft#Raft.BootstrapCluster). The documentation for this method says the following:

> BootstrapCluster is equivalent to non-member BootstrapCluster but can be called on an un-bootstrapped Raft instance after it has been created. This should only be called at the beginning of time for the cluster, and **you absolutely must make sure that you call it with the same configuration on all the Voter servers**. There is no need to bootstrap Nonvoter and Staging servers.

However, it appears as though this method need only be called _once_, at the beginning of the cluster on only a _single_ server, not all of the voting servers. In the cases where either:

1. All of the servers are defined in the `raft.Configuration`, or
2. Only the first server is defined in the configuration and additional nodes are added using [`func (*Raft) AddVoter`](https://godoc.org/github.com/hashicorp/raft#Raft.AddVoter)

Any additional calls to `BootstrapCluster` from _any_ participating voter server cause the following error to occur:

> bootstrap only works on new clusters

Any idea if this is the expected behaviour or erroneous documentation for this method?

sam3d:
This PR aims to resolve #336.

With clarification from @RobbieMcKinstry, this PR includes the following clarifications:

- The sentence "you absolutely must make sure that you call it with the same configuration on all the Voter servers" has been completely removed. The actual meaning of this sentence was "If you are going to call it on all Voter servers, it must be with the same configuration". However, it reads as "This method _must_ be called on all Voter servers for the bootstrap process to work", which is incorrect. Instead, that sentence now clarifies that on any voting server that calls `BootstrapCluster`, the configuration must be identical.

- There is also a small additional paragraph that says that the cluster can only be bootstrapped from a single Voter server (which is true). It also says that any further attempts to bootstrap can be safely ignored. The combination of these two remarks provides the necessary inference (without the potential to mislead), along with the change in the first paragraph, that it may be called on all participating Voter servers, _as long_ as they all have the same configuration, and that the error from subsequent calls can be ignored.

- This documentation is also completely cloned to the other Raft instance `BootstrapCluster` method for completeness.

Hope this is all OK and makes sense, please let me know if you have any questions about my rationale here! 😄

### DIFF
--- a/api.go
+++ b/api.go
@@ -591,10 +591,17 @@ func (r *Raft) restoreSnapshot() error {
 
 // BootstrapCluster is equivalent to non-member BootstrapCluster but can be
 // called on an un-bootstrapped Raft instance after it has been created. This
-// should only be called at the beginning of time for the cluster, and you
-// absolutely must make sure that you call it with the same configuration on all
-// the Voter servers. There is no need to bootstrap Nonvoter and Staging
-// servers.
+// should only be called at the beginning of time for the cluster with a
+// configuration listing all Voter servers. There is no need to bootstrap
+// Nonvoter and Staging servers.
+//
+// A cluster can only be bootstrapped once from a single participating Voter
+// server. Any further attempts to bootstrap will return an error that can be
+// safely ignored.
+//
+// One sane approach is to bootstrap a single server with a configuration
+// listing just itself as a Voter, then invoke AddVoter() on it to add other
+// servers to the cluster.
 func (r *Raft) BootstrapCluster(configuration Configuration) Future {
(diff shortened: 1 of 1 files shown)


RobbieMcKinstry:
Hey @sam3d ! Thanks for working through this. I appreciate the effort you made to land this PR. :)

sam3d:
You're welcome! 😄

PR #339, opened 2019-06-14
Add an optional ConfigurationStore interface 

briankassouf:
FSMs that satisfy this interface will be sent configuration change updates after they are fully committed to the log

### DIFF
--- a/configuration.go
+++ b/configuration.go
@@ -32,6 +32,19 @@ func (s ServerSuffrage) String() string {
 	return "ServerSuffrage"
 }
 
+// ConfigurationStore provides an interface that can be implemented by clients
+// to store configuration updates made in the replicated log.
+type ConfigurationStore interface {
+	// StoreConfig is invoked once a log entry containing a configuration change
+	// is committed. It takes the index at which the configuration was written
+	// and the configuration value.
+	StoreConfig(index uint64, configuration Configuration) error
+}
+
+type nopConfigurationStore struct{}
+
+func (s nopConfigurationStore) StoreConfig(_ uint64, _ Configuration) error { return nil }
+
 // ServerID is a unique string identifying a server for all time.
 type ServerID string
 
--- a/configuration.go
+++ b/configuration.go
@@ -39,6 +39,9 @@ func (s ServerSuffrage) String() string {
 // persistent FSM state can behave as a complete snapshot, and be able to recover
 // without an external snapshot just for persisting the raft configuration.
 type ConfigurationStore interface {
+	// ConfigurationStore is a superset of the FSM functionality
+	FSM
+
 	// StoreConfiguration is invoked once a log entry containing a configuration
 	// change is committed. It takes the index at which the configuration was
 	// written and the configuration value.
--- a/configuration.go
+++ b/configuration.go
@@ -32,7 +32,7 @@ func (s ServerSuffrage) String() string {
 	return "ServerSuffrage"
 }
 
-// ConfigurationStore provides an interface that can be implemented by FSMs
+// ConfigurationStore provides an interface that can optionally be implemented by FSMs
 // to store configuration updates made in the replicated log.
 type ConfigurationStore interface {
 	// StoreConfiguration is invoked once a log entry containing a configuration
--- a/configuration.go
+++ b/configuration.go
@@ -33,10 +33,10 @@ func (s ServerSuffrage) String() string {
 }
 
 // ConfigurationStore provides an interface that can optionally be implemented by FSMs
-// to store configuration updates made in the replicated log. In general this is only 
-// necessary for FSMs that mutate durable state directly instead of applying changes 
-// in memory and snapshotting periodically. By storing configuration changes, the 
-// persistent FSM state can behave as a complete snapshot, and be able to recover 
+// to store configuration updates made in the replicated log. In general this is only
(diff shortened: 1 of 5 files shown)


raft_test.go
```diff
@@ -1703,6 +1785,9 @@ func TestRaft_UserSnapshot(t *testing.T) {
 	conf := inmemConfig(t)
 	conf.SnapshotThreshold = 50
 	conf.TrailingLogs = 10
+	// Set protocl to a lower version to avoid applying configuration entries on
```
briankassouf:
```suggestion
	// Set protocol to a lower version to avoid applying configuration entries on
```

configuration.go
```diff
@@ -32,6 +32,19 @@ func (s ServerSuffrage) String() string {
 	return "ServerSuffrage"
 }
 
+// ConfigurationStore provides an interface that can be implemented by FSMs
```
banks: nit:
```suggestion
// ConfigurationStore provides an interface that can optionally be implemented by FSMs
```

configuration.go
```diff
@@ -32,6 +32,19 @@ func (s ServerSuffrage) String() string {
...
+// to store configuration updates made in the replicated log.
```
banks: Might be nice to add a line to say why they might want to as a hint for people trying to figure out how to implement raft themselves. Something like. "In general this is only necessary for FSMs that mutate durable state directly instead of applying changes in memory and snapshotting periodically. By storing configuration changes, the persistent FSM state can behave as a complete snapshot, and be able to recover without an external snapshot just for persisting the raft configuration."

fsm.go
```diff
@@ -50,12 +50,28 @@ func (r *Raft) runFSM() {
...
 			metrics.MeasureSince([]string{"raft", "fsm", "apply"}, start)
+
+		case LogConfiguration:
+			configStore, ok := r.fsm.(ConfigurationStore)
+			if !ok {
+				// Return early to avoid incrementing the index and term for
+				// an unimplemented operation.
+				if req.future != nil {
+					req.future.respond(nil)
+				}
```
banks: Would it be cleaner to set this up in a defer at the start of the closure rather than duplicate it here?

raft_test.go
```diff
@@ -1703,6 +1785,9 @@ func TestRaft_UserSnapshot(t *testing.T) {
 	conf := inmemConfig(t)
 	conf.SnapshotThreshold = 50
 	conf.TrailingLogs = 10
+	// Set protocol to a lower version to avoid applying configuration entries on
+	// bootstrap
+	conf.ProtocolVersion = 1
```
banks: Didn't look at the context but does this mean this test will not be covering the v2+ case of the thing it's meant to be testing any more? If so, is there a way we can make it pass reliably without limiting the coverage of the current protocol version - like have the mock able to turn off config persistence where we don't care about testing that?

briankassouf: I went down this path originally but it was turning out to be a major change to the testing framework. I don't believe this test covers any differences between v1 and v2. It just tests that snapshot returns errors in certain circumstances

banks: 🤔 Perhaps I'm missing something but is it not possible to have another field on MockStore that optionally enables the new interface and if it's not true then just make the new Implementation method a no-op?

Then all old tests would work just as before presumably since there is no state change for config updates but we could still write the new test with just a one liner to enable the ConfigStore interface?

briankassouf: The issue is the test expects Raft to report a 0 Applied index on first snapshot. But if the FSM even implements the ConfigurationStore interface then the FSM will increment its applied index for the initial bootstrap call. So a no-op isn't enough, we would need two FSM types, one that implements the new interface and one that doesn't. This is where the large amount of test framework changes came in, the cluster object assumes the FSM's type to be *MockFSM a number of places

banks: Gotcha.

I can let it go if there is no clean way but would it work to do something like:

```go
type MockPersistentFSM struct {
  MockFSM
  configurations []Configuration
}
// Implement ConfigurationStore for MockPersistentFSM
```

Then the existing tests all get the same behaviour as before, new tests can use the new type and any helper that expects `*MockFSM` can be passed `&fsm.MockFSM`?

If it gets more complex than that I guess this is reasonable for now and is not super likely to change in the future but it feels like at least a _potential_ footgun if a future change to the code path that is protocol version dependent will seem to pass all tests fine but actually won't be exercised here.

configuration.go
```diff
@@ -32,6 +32,23 @@ func (s ServerSuffrage) String() string {
 	return "ServerSuffrage"
 }
 
+// ConfigurationStore provides an interface that can optionally be implemented by FSMs
+// to store configuration updates made in the replicated log. In general this is only
+// necessary for FSMs that mutate durable state directly instead of applying changes
+// in memory and snapshotting periodically. By storing configuration changes, the
+// persistent FSM state can behave as a complete snapshot, and be able to recover
+// without an external snapshot just for persisting the raft configuration.
+type ConfigurationStore interface {
```
mitchellh: Random drive-by comment: if this interface is only useful for FSMs, should you embed the FSM interface in here as well?

I tend to find that to be a fairly self-documenting behavior I appreciate.

briankassouf: Good idea, i'll add that

PR #340, opened 2019-06-15
Add a config option specifying no snapshot restore on startup

briankassouf:


api.go
```diff
@@ -545,23 +547,23 @@ func (r *Raft) restoreSnapshot() error {
...
+			source.Close()
+			if err != nil {
+				r.logger.Error(fmt.Sprintf("Failed to restore snapshot %v: %v", snapshot.ID, err))
+				continue
+			}
 
+			r.logger.Info(fmt.Sprintf("Restored from snapshot %v", snapshot.ID))
+		}
 		// Update the lastApplied so we don't replay old logs
 		r.setLastApplied(snapshot.Index)
```
i0rek: Is it correct to leave this here since nothing was actually restored?

api.go
```diff
@@ -291,19 +291,21 @@ func RecoverCluster(conf *Config, fsm FSM, logs LogStore, stable StableStore,
...
+			err = fsm.Restore(source)
+			// Close the source after the restore has completed
+			source.Close()
+			if err != nil {
+				// Same here, skip and try the next one.
+				continue
+			}
 		}
 
 		snapshotIndex = snapshot.Index
```
i0rek: Is it correct to leave this here since nothing was actually restored?

briankassouf:
@i0rek Good question, yeah we still need a way to tell Raft which index the FSM is currently at even if we don't need to restore a snapshot to the FSM. This way if there are still pending logs they will be replayed on top of the FSM. The Snapshot Metadata is still used to do this. In Vault's case the snapshot metadata represents the current state of the FSM.

jefferai:
Test failure just normal flakiness; merging.

PR #347, opened 2019-07-11
Fix broken link to Raft paper

tcarrio:
The link used for Raft is outdated. It seems the original, hosted at stanford.edu (https://ramcloud.stanford.edu/wiki/download/attachments/11370504/raft.pdf) is no longer available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ to port Consul to these new interfaces.
 
 ## Protocol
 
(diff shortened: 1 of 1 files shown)


tcarrio:
Since the [raft GitHub organization](https://github.com/raft) hosts this paper on their repository, this should be a better location to maintain the link. The co-author of the Raft paper manages that repository himself, and GitHub acts a CDN to serve it.

PR #348, opened 2019-07-11
add Leaving stage

mengskysama:
https://github.com/hashicorp/consul/issues/6113
https://github.com/hashicorp/raft/pull/348

### DIFF
--- a/api.go
+++ b/api.go
@@ -134,6 +134,9 @@ type Raft struct {
 	shutdownCh   chan struct{}
 	shutdownLock sync.Mutex
 
+	// leaving is used for mark leave status
+	leaving bool
+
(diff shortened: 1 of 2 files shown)


s-christoff:
Hey @mengskysama, thank you so much for your contribution! I think I'm missing some context on how this relates to the the problem in Consul you referenced, could you elaborate more for me?

s-christoff:
Hey @mengskysama, thank you for bringing this to our attention. We are discussing different ways to go about solving this problem - with our primary focus to add a new entry in the log that states the server is intending to leave and then afterwards process the leave. I am going to close this PR for now - and we will continue to look into the issues that have been opened.

PR #351, opened 2019-07-12
Updates around linting errors, formatting, and more.

tcarrio:
This provides corrections to comment structure, inclusion of new comments on exported types, explicit aliases on hyphenated imports, renaming of variables that don't follow convention in Go, removal of redundant typing on variable assignment, removed unnecessary else block, etc.

I have yet to dig through `log_cache.go`, which I may knock out tomorrow, but these updates are ready whenever.

### DIFF
--- a/config.go
+++ b/config.go
@@ -8,8 +8,8 @@ import (
 	"github.com/hashicorp/go-hclog"
 )
 
-// These are the versions of the protocol (which includes RPC messages as
-// well as Raft-specific log entries) that this server can _understand_. Use
+// ProtocolVersion is the version of the protocol (which includes RPC messages
+// as well as Raft-specific log entries) that this server can _understand_. Use
 // the ProtocolVersion member of the Config object to control the version of
 // the protocol to use when _speaking_ to other servers. Note that depending on
 // the protocol version being spoken, some otherwise understood RPC messages
@@ -88,13 +88,15 @@ import (
 type ProtocolVersion int
 
 const (
+	// ProtocolVersionMin is the minimum protocol version
 	ProtocolVersionMin ProtocolVersion = 0
-	ProtocolVersionMax                 = 3
+	// ProtocolVersionMax is the maximum protocol version
+	ProtocolVersionMax = 3
 )
 
-// These are versions of snapshots that this server can _understand_. Currently,
-// it is always assumed that this server generates the latest version, though
-// this may be changed in the future to include a configurable version.
+// SnapshotVersion is the version of snapshots that this server can understand.
+// Currently, it is always assumed that the server generates the latest version,
+// though this may be changed in the future to include a configurable version.
 //
 // Version History
 //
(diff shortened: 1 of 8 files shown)


api.go
```diff
@@ -9,9 +9,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/hashicorp/go-hclog"
+	hclog "github.com/hashicorp/go-hclog"
```
jefferai: I've not seen linters making this change before and this is a convention understood by the compiler. I don't think it's necessary.

tcarrio: Yeah that can be reverted, I guess it's something about explicit vs implicit in the linter. It's functional both ways too so no worries

config.go
```diff
@@ -88,13 +88,15 @@ import (
 type ProtocolVersion int
 
 const (
+	// ProtocolVersionMin is the minimum protocol version
 	ProtocolVersionMin ProtocolVersion = 0
-	ProtocolVersionMax                 = 3
+	// ProtocolVersionMax is the maximum protocol version
+	ProtocolVersionMax = 3
```
s-christoff: This is nitpicky -
But I liked the _look_ of it where it was aligned with `ProtocolVersionMin` - can we revert it to how it was or is this one of the formatting errors you're trying to fix?

config.go
```diff
@@ -112,8 +114,10 @@ const (
 type SnapshotVersion int
 
 const (
+	// SnapshotVersionMin is the minimum snapshot version
 	SnapshotVersionMin SnapshotVersion = 0
-	SnapshotVersionMax                 = 1
+	// SnapshotVersionMax is the maximum snapshot version
+	SnapshotVersionMax = 1
```
s-christoff: Same spacing comment as in ProtocolVersion.

config.go
```diff
@@ -88,13 +88,15 @@ import (
...
 	ProtocolVersionMin ProtocolVersion = 0
-	ProtocolVersionMax                 = 3
+	// ProtocolVersionMax is the maximum protocol version
+	ProtocolVersionMax = 3
 )
 
-// These are versions of snapshots that this server can _understand_. Currently,
-// it is always assumed that this server generates the latest version, though
-// this may be changed in the future to include a configurable version.
+// SnapshotVersion is the versions of snapshots that this server can _understand_.
```
s-christoff: Hi! 
Nitpicking here:
```
// SnapshotVersion is the version of snapshots that this server can understand.
// Currently, it is always assumed that the server generates the latest version,
// though this may be changed in the future to include a configurable version.
```

s-christoff:
This looks great, thank you so much for doing this! 
I think I just have that one _teeny_ comment and then if you could rebase we'll get this merged. :)

tcarrio:
Absolutely! I'll get a chance to finish it up tomorrow 👍

tcarrio:
@s-christoff rebased and squashed and looking green :+1:

s-christoff:
Thank you again! :)

PR #353, opened 2019-07-18
Add support for extensions in log entries

jefferai:
This allows supporting middleware FSM/API wrappers. Test functions were moved to a public file so that middleware can use raft FSMs/clusters for tests.


jefferai:
Note: test failures are unrelated. I could keep poking with empty commits until one works given I'm seeing different test failures on different builds, but doesn't seem worth it.

api.go
```diff
@@ -637,6 +646,15 @@ func (r *Raft) Leader() ServerAddress {
...
 func (r *Raft) Apply(cmd []byte, timeout time.Duration) ApplyFuture {
 	metrics.IncrCounter([]string{"raft", "apply"}, 1)
+
+	return r.ApplyLog(Log{Data: cmd}, timeout)
+}
+
+// ApplyLog performs Apply but takes in a Log directly. The only values
+// currently taken from the submitted Log are Data and Extensions.
+func (r *Raft) ApplyLog(log Log, timeout time.Duration) ApplyFuture {
+	metrics.IncrCounter([]string{"raft", "apply_with_log"}, 1)
```
banks: Is there value in making this a new metric name? My concern is that people with existing monitoring setup likely have `raft.apply` graphed and assume it covers all `apply` calls. If their workload slowly transitions over time to include more chunked calls then they won't see those in their existing graphs as the chunks will only get counted as `apply_with_log`.

I wonder if it would be better to just make this `apply` like now and remove the increment from `Apply` above to avoid double counting. We can record a separate counter in Consul if we want to make it clear how many applies are chunked vs not chunked where we actually have context to know.

It seems like an implementation detail of the client whether this should be separate or not so I vote for keeping it just as `apply` and letting clients differentiate in how/what they are applying where they have meaningful context for that. (e.g. Consul could add a `raft.chunked` counter to be explicit about why it's different/important).

api.go
```diff
@@ -14,6 +14,15 @@ import (
 	"github.com/armon/go-metrics"
 )
 
+const (
+	// This is the current suggested max size of the data in a raft log entry.
+	// This is based on current architecture, default timing, etc. Clients can
+	// ignore this value if they want as there is no actual hard checking
+	// within the library. As the library is enhanced this value may change
+	// over time to reflect current suggested maximums.
+	SuggestedMaxDataSize = 512 * 1024
```
banks: Maybe add a sentence like:

> Increasing beyond this risks RPC IO taking too long and preventing timely heartbeat signals which are sent in serial in current transports, potentially causing leadership instability.

log.go
```diff
@@ -47,6 +47,12 @@ type Log struct {
 
 	// Data holds the log entry's type-specific data.
 	Data []byte
+
+	// Extensions holds an opaque byte slice of information for middleware. It
+	// is up to the client of the library to properly modify this as it adds
+	// layers and remove those layers when appropriate. This value is a part of
+	// the log, so very large values could cause timing issues.
+	Extensions []byte
```
banks: I'd like to test this for backward compatibility. I think it should be OK since msgpack should be able to decode old messages into this struct and new messages into the old struct although it will strip the Extensions field in the process.

For chunking old servers stripping is probably not a big deal because they will crash when they try to decode/apply the partial chunk anyway and so upgrade will need to be managed by client or operator somehow.

But in general, I'm not sure if we need to make this change more explicit somehow. The way we've dealt with compatibilities in raft before is via protocol numbers which are negotiated by raft peers at start and can be forced to older version to ensure compatibility through an upgrade. That is fairly heavyweight though e.g. in Consul requires two rolling restarts to safely upgrade servers through a raft protocol change when we could just automatically disallow using the new chunking feature until all servers are running a new version.

We should probably _at least_ document in this comment that setting this is only safe when all raft peers are known to be running a version of raft with this change in though? Violating the assumption that log playback on all replicas will behave the same seems risky in general here. Imagine another extension that isn't chunking and doesn't result in invalid payloads but just alters how the FSM should apply the log - in that case failure of the client system to manage versioning manually would cause silent FSM corruption and divergence between replicas.

banks: Devils advocate says that it's the library client's responsibility to ensure that FSMs are backwards compatible on top of this library. Since this doesn't actually change the encoding format or cause anything to inherently break between peers other than the interpretation of the log messages themselves, it can stay like it is and we should just document (somehow?) that clients need to manage compatibility concerns themselves.

This is all assuming that this is actually backwards compatible and will be decoded by an old client from a log or RPC without crashing etc. I think that's the case but we should test it to be sure before accepting.

api.go
jefferai: I'm not sure if there's any value and I'm fine just moving apply there, will do so.

api.go
jefferai: Done, thanks.

log.go
jefferai: I'll add some backwards compat testing in the chunking library. And I'll add some documentation as you suggested.

As for the upgrade path I also took the position that it's up to the client. So long as you don't trigger an extension -- in this case, chunking, by going above the limit (which is the same as it was before) -- then you don't run the risk of triggering chunking. So for any scenario in which we tell someone to turn on larger value support in Consul, we'd also tell them that all Consul nodes _must_ be at least version X. And if we ever publicly document that option, we would do the same thing, e.g. warning, do not make higher than the default if all Consul servers are not at least version X.

jefferai: I've added an upgrade warning here. I'll work on the backwards compat tests shortly.

jefferai: I did some testing with Consul, one node using raft including Extensions and one without. As I think you suspected, for a normal sized apply that didn't trigger ApplyLog, everything worked fine. When I went through ApplyLog it caused a decoding panic on the Consul node prior to this PR. (Technically it was an error; Consul just panics in this situation).

jefferai:
Tests pass locally.

PR #357, opened 2019-07-24
Export leader field

aalda:
Export the Leader field from LeaderObservation to make accessible the observation value.

### DIFF
--- a/observer.go
+++ b/observer.go
@@ -18,7 +18,7 @@ type Observation struct {
 
 // LeaderObservation is used for the data when leadership changes.
 type LeaderObservation struct {
-	leader ServerAddress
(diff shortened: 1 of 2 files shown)


s-christoff:
Hey @aalda , thank you so much for this contribution! I wanted to know if you have any particular use case for this change?

aalda:
Hi @s-christoff, I was using a custom observer to update some metadata about the cluster state, including the Leader address. But later, I found a more convenient way to do it. Anyway, I think that exporting that field makes the LeaderObservation usable and consistent with the PeerObservation.

RobbieMcKinstry:
LGTM

PR #360, opened 2019-08-04
Replaced logging with hclog throughout raft.

elliotcourant:
This simply replaces the existing logging implementation with hclog to
make everything consistent throughout the library.

### DIFF
--- a/fuzzy/transport.go
+++ b/fuzzy/transport.go
@@ -5,8 +5,8 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/go-hclog"
 	"io"
-	"log"
 	"os"
 	"path/filepath"
 	"sync"
@@ -32,10 +32,10 @@ type appendEntries struct {
 type transports struct {
 	sync.RWMutex
 	nodes map[string]*transport
-	log   *log.Logger
+	log   hclog.Logger
 }
 
-func newTransports(log *log.Logger) *transports {
+func newTransports(log hclog.Logger) *transports {
 	return &transports{
 		nodes: make(map[string]*transport),
 		log:   log,
@@ -65,7 +65,7 @@ type TransportHooks interface {
 }
 
 type transport struct {
-	log        *log.Logger
+	log        hclog.Logger
 	transports *transports
 	node       string
 	ae         []appendEntries
@@ -99,7 +99,7 @@ func (t *transport) sendRPC(target string, req interface{}, resp interface{}) er
 	t.transports.RLock()
 	tt := t.transports.nodes[target]
 	if tt == nil {
-		t.log.Printf("sendRPC target=%v but unknown node (transports=%v)", target, t.transports.nodes)
+		t.log.Info("sendRPC unknown node", "target", target, "transports", t.transports.nodes)
 		t.transports.RUnlock()
 		return fmt.Errorf("Unknown target host %v", target)
 	}
@@ -132,7 +132,7 @@ func (t *transport) sendRPC(target string, req interface{}, resp interface{}) er
 		}
 		rpc.Command = &appEnt
 	default:
-		t.log.Printf("Unexpected request type %T %+v", req, req)
+		t.log.Warn("unexpected request type", "type", fmt.Sprintf("%T", req), "request", req)
 	}
 	var result *raft.RPCResponse
 	if t.hooks != nil {
@@ -152,7 +152,7 @@ func (t *transport) sendRPC(target string, req interface{}, resp interface{}) er
 			}
 		}
 	}
-	//t.log.Printf("sendRPC %v -> %v : %+v", t.node, target, rpc.Command)
+	// t.log.Printf("sendRPC %v -> %v : %+v", t.node, target, rpc.Command)
 	if result == nil {
 		tt.consumer <- rpc
 		cr := <-rc
@@ -165,7 +165,7 @@ func (t *transport) sendRPC(target string, req interface{}, resp interface{}) er
 			result.Error = err
 		}
 	}
-	//t.log.Printf("sendRPC %v <- %v: %+v %v", t.node, target, result.Response, result.Error)
+	// t.log.Printf("sendRPC %v <- %v: %+v %v", t.node, target, result.Response, result.Error)
 	buff = bytes.Buffer{}
 	codec.NewEncoder(&buff, &codecHandle).Encode(result.Response)
 	codec.NewDecoderBytes(buff.Bytes(), &codecHandle).Decode(resp)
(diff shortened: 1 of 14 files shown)


fuzzy/transport.go
```diff
@@ -152,7 +152,7 @@ func (t *transport) sendRPC(target string, req interface{}, resp interface{}) er
 			}
 		}
 	}
-	//t.log.Printf("sendRPC %v -> %v : %+v", t.node, target, rpc.Command)
+	// t.log.Printf("sendRPC %v -> %v : %+v", t.node, target, rpc.Command)
```
s-christoff: Are these supposed to be here? I'm kind of confused on their purpose 🤔

fuzzy/transport.go
```diff
@@ -165,7 +165,7 @@ func (t *transport) sendRPC(target string, req interface{}, resp interface{}) er
 			result.Error = err
 		}
 	}
-	//t.log.Printf("sendRPC %v <- %v: %+v %v", t.node, target, result.Response, result.Error)
+	// t.log.Printf("sendRPC %v <- %v: %+v %v", t.node, target, result.Response, result.Error)
```
s-christoff: There's one here too!

net_transport.go
```diff
@@ -310,7 +318,7 @@ func (n *NetworkTransport) getProviderAddressOrFallback(id ServerID, target Serv
 	if n.serverAddressProvider != nil {
 		serverAddressOverride, err := n.serverAddressProvider.ServerAddr(id)
 		if err != nil {
-			n.logger.Printf("[WARN] raft: Unable to get address for server id %v, using fallback address %v: %v", id, target, err)
+			n.logger.Warn(fmt.Sprintf("unable to get address for server id %s, using fallback address %v", id, target), err)
```
s-christoff: Could we use `hclog.Fmt()` instead of `fmt.Sprintf` here? 
Just something I thought of while looking this over. :)

testing.go
```diff
@@ -354,7 +359,8 @@ func (c *cluster) pollState(s RaftState) ([]*Raft, uint64) {
 // GetInState polls the state of the cluster and attempts to identify when it has
 // settled into the given state.
 func (c *cluster) GetInState(s RaftState) []*Raft {
-	c.logger.Printf("[INFO] Starting stability test for raft state: %+v", s)
+	c.logger.Info("starting stability test", "raft-state", s)
+	// c.logger.Printf("[INFO] Starting stability test for raft state: %+v", s)
```
s-christoff: Another comment that I think should be deleted, maybe? 😛

s-christoff:
Hey @elliotcourant , thank you SO much for this. Seriously, this is something we've been wanting to have done for awhile. We all really appreciate it (and we all rejoiced when we saw this PR!)

A couple nitpick-y things I commented on above but thought I would expand on. There's some usage of `fmt.Sprintf` throughout this, I think it may be preferred to use `hclog.Fmt()` here. It's no big deal either way.

I noted on some comments I saw that may have meant to be deleted, but they also may be necessary, if they are let me know!

Again, thank you 🙏

elliotcourant:
@s-christoff Thank you for the quick follow up! 

It would make sense to me to remove those commented out log entries. At this point they could easily be added back in, but since they are not being used they could also be checked with a debugger instead. I will remove them for now.

I'll make the adjustments to use `hclog.Fmt()`. Should all log entries using `fmt` be updated as well? (Including ones not touched by this MR?)

Thank you very much!

elliotcourant:
Kind of went nuts and made all of the logging consistent. There are a few lingering fmt.Sprintfs in some tests but in the primary code everything should be consistent with hclog now!

net_transport.go
elliotcourant: I've redone all of the logging that had formatting and either rephrased it to better fit with the hclog key/value format and/or added hclog.Fmt() instead of Sprintf.

testing.go
```diff
@@ -411,17 +416,18 @@ func (c *cluster) GetInState(s RaftState) []*Raft {
...
 
 		case t, ok := <-timer.C:
 			if !ok {
-				c.FailNowf("[ERR] Timer channel errored")
+				c.FailNowf("timer channel errored")
 			}
-			c.logger.Printf("[INFO] Stable state for %s reached at %s (%d nodes), %s from start of poll, %s from cluster start. Timeout at %s, %s after stability",
-				s, inStateTime, len(inState), inStateTime.Sub(pollStartTime), inStateTime.Sub(c.startTime), t, t.Sub(inStateTime))
+
+			c.logger.Info(fmt.Sprintf("stable state for %s reached at %s (%d nodes), %s from start of poll, %s from cluster start. Timeout at %s, %s after stability",
```
elliotcourant: This one is kind of a tricky one to replace and is only used for testing so I'm going to leave it in.

fuzzy/transport.go
```diff
@@ -132,7 +132,7 @@ func (t *transport) sendRPC(target string, req interface{}, resp interface{}) er
 		}
 		rpc.Command = &appEnt
 	default:
-		t.log.Printf("Unexpected request type %T %+v", req, req)
+		t.log.Warn("unexpected request type", "type", fmt.Sprintf("%T", req), "request", req)
```
elliotcourant: Forgot this one

testing.go
elliotcourant: @s-christoff are you alright with leaving this one in?

elliotcourant:
I had missed several others. Now everything (except for the one test) is using a normal hclog format, or is using hclog.Fmt to format strings.

@s-christoff I hope this is alright, if its not I have not squashed my commits yet so I can revert any changes that you deem unnecessary.

testing.go
s-christoff: Yeah! LGTM :)

raft_test.go
```diff
@@ -1856,7 +1856,7 @@ func TestRaft_ProtocolVersion_Upgrade_2_3(t *testing.T) {
 	// Use the new ID-based API to add the server with its ID.
 	future := c.Leader().AddVoter(c1.rafts[0].localID, c1.rafts[0].localAddr, 0, 1*time.Second)
 	if err := future.Error(); err != nil {
-		c.FailNowf("[ERR] err: %v", err)
+		c.FailNowf("err: %v", err)
```
endocrimes: Are these test changed accidental? They’re kinda useful for gripping test output

elliotcourant: They are intentional. The hclog library should include the log level in the entry every time anyway; and `c.FailNowf` is just a wrapper around an hclog call now. So it will include `ERR` or `ERROR` in square brackets for those entries.

s-christoff:
🤦‍♀️ Hit the wrong button, sorry

Issue #358, opened 2019-07-24
Snapshot can attempt to delete a negative range from log


banks:
Debugging something in consul I noticed raft log messages like this:

```
raft: Compacting logs from 26108 to 6079
```

Looking at https://github.com/hashicorp/raft/blob/ff523e1c1515a5b33cc3153a34c984093fed06bb/snapshot.go#L231-L235

This is in fact the min/max values we are passing to `DeleteRange` on the log store.

Our `LogStore` interface doesn't define what the behaviour should from implementations in that case although our `raft-boltdb` store that we use everywhere happens to make that a no-op so it's not causing actual issue other than a confusing log message.

I suggest we  add a bounds check just before those lines to verify that max is strictly larger than min and if not just return possibly with a nice log like "no logs to truncate".

jscode017:
Are anyone on this one?
If not may I take this as my first issue?

banks:
Please do take it! Thanks for contributing!

catsby:
https://github.com/hashicorp/raft/pull/362 was opened, thanks @jscode017 :D

banks:
Thanks @dineshba good spot.

PR #362, opened 2019-09-14
issue358 Snapshot can attempt to delete a negative range from log

jscode017:
My first try for this issue and the project.

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -228,6 +228,11 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
 	// after the snapshot to be removed.
 	maxLog := min(snapIdx, lastLogIdx-r.conf.TrailingLogs)
 
+	if maxLog <= minLog {
+		r.logger.Info("no logs to truncate")
+		return nil
+	}
+
 	r.logger.Info("compacting logs", "from", minLog, "to", maxLog)
 
 	// Compact the logs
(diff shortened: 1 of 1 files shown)


snapshot.go
```diff
@@ -228,6 +228,11 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
 	// after the snapshot to be removed.
 	maxLog := min(snapIdx, lastLogIdx-r.conf.TrailingLogs)
 
+	if maxLog <= minLog {
```
catsby: Should this be greater-than instead? @banks suggested "max is strictly larger than min". I lack some context but note that when I read "strictly larger" I wouldn't expect `<=`

catsby: I would personally also do `minLog > maxLog` instead of `maxLog < minLog` for the conditional, I feel that would improve the readability. I had to read it 2x to realize what you had was correct, I think my mind just expected min/max and not max/min, but I admit that's my personal bias.

catsby:
Fixes https://github.com/hashicorp/raft/issues/358

snapshot.go
banks: Yeah the strictly part is necessar.

See the interface https://github.com/hashicorp/raft/blob/ff523e1c1515a5b33cc3153a34c984093fed06bb/log.go#L85 says the max an min are "inclusive" which means that min == max is actually valid and it will cause exactly one log entry to be deleted. This condition would consider `min:1000, max:1000` to be invalid and say "no logs to truncate" when in fact the current behaviour would be to truncate exactly one log with index 1000.

So yeah strict inequality is more correct here.

jscode017:
Thanks, change it to strict inequality

banks:
Thanks again @jscode017!

Issue #352, opened 2019-07-16
fuzzy: Unable to run fuzzy test


nilsocket:
Unable to run tests under fuzzy folder.

It's a build error

I tried by checking out to v1.1.0, even there it is failing.

Thank you.

catsby:
I opened a PR for this here: https://github.com/hashicorp/raft/pull/363

Unfortunately I don't know for absolute certain that it fixes the issues that @nilsocket  experienced, there are some missing details on the specifics but at least with that change they tests compile, run, and pass.

nilsocket:
@catsby , I think I just ran `go test` in that directory, I think it's safe to close this issue, if you have fixed it.

Thank you.

catsby:
@nilsocket awesome, thanks. It should get closed automatically when #363 is merged

PR #363, opened 2019-09-17
Fix fuzzy

catsby:
This PR does three things in the `fuzzy/` directory, but the main thing is in the method `sendNApplies` where if the `Leader()` method returns `nil` due to hitting the timeout, then `sendNApplies` simply returns an empty slice instead of trying triggering a nil pointer dereference. 

It also extends the leader timeout as I found that alone helped many of the tests to pass. 

In the Makefile, I added support for `TESTARGS` so you can invoke `make fuzz` and run only specific tests, or use other things like `-count` or such. I also extended the timeout, as I found running the `TestRaft_FuzzyLeadershipTransfer` test alone was taking ~140 seconds on my laptop. 

I believe this fixes https://github.com/hashicorp/raft/issues/352 , in such that the tests no longer panic

Test results from my 2016 laptop:

```
[raft][fix_fuzzy](4)$ make fuzz 
go test  -timeout=20m ./fuzzy
ok      github.com/hashicorp/raft/fuzzy 933.596s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,10 +2,10 @@ DEPS = $(go list -f '{{range .TestImports}}{{.}} {{end}}' ./...)
 TEST_RESULTS_DIR?=/tmp/test-results
 
 test:
-	go test -timeout=60s -race .
+	go test $(TESTARGS) -timeout=60s -race .
 
 integ: test
-	INTEG_TESTS=yes go test -timeout=25s -run=Integ .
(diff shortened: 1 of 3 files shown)


catsby:
I amended my PR with commit bb46780 that just adds support for `$TESTARGS` to the other steps that run `go test`

PR #364, opened 2019-09-24
Enable FSM batching through optionally implemented interface

briankassouf:
Currently this library operates under the assumption that the FSM apply will be much faster than the Log apply. Based on this assumption the library does Log apply batching but no FSM level batching. When using an in-memory FSM this isn't a problem, but when the FSM is doing additional work- such as persisting the data to disk- this causes high latencies. 

This PR introduces a new `BatchingFSM` interface which extends the `FSM` interface to add an `ApplyBatch([]*Log) []interface{}` function. This can optionally be implemented by clients to enable multiple logs to be applied to the FSM in batches. Up to `MaxAppendEntries` could be sent in a batch.

`ApplyBatch` is invoked once a batch of log entries has been committed and are ready to be applied to the FSM. `ApplyBatch` will take in an array of log entries. These log entries could be of a few types so clients should check the log type prior to attempting to decode the data attached.  Presently the `LogCommand` and `LogConfiguration` types will be sent.

The returned slice must be the same length as the input. The returned values will be made available in the `ApplyFuture` returned by raft.Apply method if that method was called on the same Raft node as the FSM.

### DIFF
--- a/fsm.go
+++ b/fsm.go
@@ -152,12 +152,15 @@ func (r *Raft) runFSM() {
 
 		var i int
 		for _, req := range reqs {
+			var resp interface{}
+			// If the log was sent to the FSM, retrieve the response.
+			if shouldSend(req.log) {
+				resp = responses[i]
+				i++
+			}
+
 			if req.future != nil {
-				// If the log was sent to the FSM, retrieve the response.
-				if shouldSend(req.log) {
-					req.future.response = responses[i]
-					i++
-				}
+				req.future.response = resp
 				req.future.respond(nil)
 			}
 		}
--- a/fsm.go
+++ b/fsm.go
@@ -31,7 +31,19 @@ type FSM interface {
 	Restore(io.ReadCloser) error
 }
 
+// BatchingFSM extends the FSM interface to add an ApplyBatch function. This can
+// optionally be implemented by clients to enable multiple logs to be applied to
+// the FSM in batches. Up to MaxAppendEntries could be sent in a batch.
 type BatchingFSM interface {
+	// ApplyBatch is invoked once a batch of log entries has been committed and
+	// are ready to be applied to the FSM. ApplyBatch will take in an array of
+	// log entries. These log entries could be of a few types clients should
+	// check the log type prior to attempting to decode the data attached.
+	// Presently the LogCommand and LogConfiguration types will be sent.
+	//
+	// The returned slice must be the same length as the input. The returned
+	// values will be made available in the ApplyFuture returned by Raft.Apply
+	// method if that method was called on the same Raft node as the FSM.
 	ApplyBatch([]*Log) []interface{}
 
 	FSM
@@ -101,15 +113,22 @@ func (r *Raft) runFSM() {
 			return
 		}
 
+		var lastBatchIndex, lastBatchTerm uint64
 		logs := make([]*Log, len(req))
 		for i, ct := range req {
 			logs[i] = ct.log
+			lastBatchIndex = ct.log.Index
+			lastBatchTerm = ct.log.Term
 		}
 
 		start := time.Now()
 		responses := batchingFSM.ApplyBatch(logs)
 		metrics.MeasureSince([]string{"raft", "fsm", "applyBatch"}, start)
 		metrics.SetGauge([]string{"raft", "fsm", "applyBatchNum"}, float32(len(req)))
+
+		// Update the indexes
+		lastIndex = lastBatchIndex
+		lastTerm = lastBatchTerm
 
 		for i, resp := range responses {
 			if resp != nil && req[i].future != nil {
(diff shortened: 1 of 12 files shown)


fsm.go
```diff
@@ -31,6 +31,24 @@ type FSM interface {
...
 
+// BatchingFSM extends the FSM interface to add an ApplyBatch function. This can
+// optionally be implemented by clients to enable multiple logs to be applied to
+// the FSM in batches. Up to MaxAppendEntries could be sent in a batch.
+type BatchingFSM interface {
+	// ApplyBatch is invoked once a batch of log entries has been committed and
+	// are ready to be applied to the FSM. ApplyBatch will take in an array of
+	// log entries. These log entries could be of a few types clients should
+	// check the log type prior to attempting to decode the data attached.
+	// Presently the LogCommand and LogConfiguration types will be sent.
```
banks: It's perhaps obvious but is it worth stating that the batch will always represent the logs in the order they have been committed and will not include any gaps. FSMs implementing this _must_ apply the commands in the batch in the same order (or equivalent to them being applied in same order) to maintain the FSM invariants.

fsm.go
```diff
@@ -31,6 +31,24 @@ type FSM interface {
...
+	//
+	// The returned slice must be the same length as the input. The returned
+	// values will be made available in the ApplyFuture returned by Raft.Apply
+	// method if that method was called on the same Raft node as the FSM.
```
banks: Should we suggest guidance on partial/total error? e.g. if the whole FSM is unavailable do all results need to contain the same error or just the first and the rest nil?

fsm.go
```diff
@@ -85,6 +105,67 @@ func (r *Raft) runFSM() {
...
+			lastBatchIndex = req.log.Index
+			lastBatchTerm = req.log.Term
+		}
+
+		var responses []interface{}
+		if len(sendLogs) > 0 {
+			start := time.Now()
+			responses = batchingFSM.ApplyBatch(sendLogs)
+			metrics.MeasureSince([]string{"raft", "fsm", "applyBatch"}, start)
+			metrics.SetGauge([]string{"raft", "fsm", "applyBatchNum"}, float32(len(reqs)))
```
banks: 🤔 what is this useful for? It's a gauge which is a single value not aggregated so each call will "override" the previous until the timeslice is up (e.g. 10s for statsd typically). I feel like if we want to indicate the average/max/min size of batches we'd need a different metric type. In pure statsd the only type that would work would unintuitively be a "timer". Other systems (Prometheus, Dogstatsd) have support for Histogram types which are what you need if you want to show the distribution (avg, max, etc) of sizes of things.

I'd suggest we just leave this out or now as I don't think it will be useful as it is - it's effectively uncontrolled random sampling of the size of the batch at any given metrics interval.

raft.go
```diff
@@ -627,33 +630,41 @@ func (r *Raft) leaderLoop() {
 				}
 			}
 
-			var numProcessed int
 			start := time.Now()
+			var groupReady []*list.Element
+			var groupFutures = make(map[uint64]*logFuture)
+			var lastInGroup uint64
```
banks: nit: I'd name this `lastIdxInGroup` to be obvious that it's an index not a pointer to an actual future or leg entry. Not a big deal though as it's only used locally here.

raft.go
```diff
@@ -627,33 +630,41 @@ func (r *Raft) leaderLoop() {
...
+					r.leaderState.inflight.Remove(e)
+				}
 			}
 
 			// Measure the time to enqueue batch of logs for FSM to apply
 			metrics.MeasureSince([]string{"raft", "fsm", "enqueue"}, start)
 
 			// Count the number of logs enqueued
-			metrics.SetGauge([]string{"raft", "commitNumLogs"}, float32(numProcessed))
+			metrics.SetGauge([]string{"raft", "commitNumLogs"}, float32(len(groupReady)))
```
banks: 🤔 I guess this is where you copied the Gauge thing from. I just don't think this does anything useful. I guess our existing stats for this are essentially meaningless too...

raft_batch_test.go
```diff
@@ -0,0 +1,26 @@
+// +build batchtest
+
+package raft
+
+func init() {
+	userSnapshotErrorsOnNoData = false
+}
+
+// NOTE: This is exposed for middleware testing purposes and is not a stable API
```
banks: We have a pattern in other modules to create a `testing.go` file that exposes Public API for test helpers for downstream dependencies. For example helpers to generate stub data or manipulate config or expose mocked out components.

I suggest we do the same here rather than have a test-time only dep documented this way.

fsm.go
briankassouf: All results should return the same error. I'll add this to the comment

raft.go
briankassouf: We should consider updating this in a separate PR

PR #365, opened 2019-09-26
Move peer id from metric name to labels in raft replication

oceyral:
Hi,

Current raft replication metrics names include the peer id of the follower to which replication is made. 

While the information is important, I feel having the id in the metric name is less than ideal as it makes the metric hard to track and graph over the lifecycle of a cluster.

See hashicorp/consul#4450 for an example where users have to work around this in order to use the metric.

This PR removes the ID from the metric name, and adds a `raft_peer_id` label containing that ID.

Example before: 
```
consul_raft_replication_heartbeat_77884991_4b4b_636a_46c3_e1521741ca26_count 1.099632e+06                              
consul_raft_replication_heartbeat_77884991_4b4b_636a_46c3_e1521741ca26_count 1.099632e+06                              
```
Example after:
```
consul_raft_replication_heartbeat_count{raft_peer_id="77884991_4b4b_636a_46c3_e1521741ca2"} 6 1.099632e+06                              
consul_raft_replication_heartbeat_count{raft_peer_id="77884991_4b4b_636a_46c3_e1521741ca26"} 1.099632e+06                              
```

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -326,7 +326,7 @@ func (r *Raft) sendLatestSnapshot(s *followerReplication) (bool, error) {
 		s.failures++
 		return false, err
 	}
-	metrics.MeasureSince([]string{"raft", "replication", "installSnapshot", string(s.peer.ID)}, start)
+	metrics.MeasureSinceWithLabels([]string{"raft", "replication", "installSnapshot"}, start, []metrics.Label{{Name: "raft_peer_id", Value: string(s.peer.ID)}})
 
 	// Check for a newer term, stop running
 	if resp.Term > req.Term {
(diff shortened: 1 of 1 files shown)


oceyral:
Hi,

Sorry for the delay, I was on holidays.

I can see why making it optional is necessary. I'll try to look into it when I have some time.

satyanash:
> The simplest thing we could do to get this accepted

@banks Another slightly inefficient, but backwards-compatible solution would be to emit _both_ the metrics.
This way, none of the older dashboards break, but new ones can also start working without requiring complex rewriting rules.
Also doesn't require messing around with configs and it "just works".

FYI there is an open issue regarding this on the consul repo as well: https://github.com/hashicorp/consul/issues/4450

A later consul version can then remove the deprecated metric name.

Let me know if you're okay with this, can submit a patch on top of the existing PR.

banks:
Both would work especially as most collectors and even Consul can filter out metrics to avoid collecting things that aren't being used.

mikemorris:
Is this fixed with the merge of https://github.com/hashicorp/raft/pull/409?

oceyral:
yes

PR #366, opened 2019-10-02
Update gauge metrics to histogram samples

briankassouf:
Existing gauge metrics `raft.commitNumLogs` and `raft.leader.dispatchNumLogs` have been updated to histogram samples. Gauge was not the correct metric type for these values and was causing samples to be dropped. Metrics sinks should be updated accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # UNRELEASED
 
+CHANGES
+
+* Existing gauge metrics `raft.commitNumLogs` and `raft.leader.dispatchNumLogs`
+  have been updated to histogram samples. Gauge was not the correct metric type
+  for these values and was causing samples to be dropped. Metrics sinks should
+  be updated accordingly. 
+
 FEATURES
 
 * Improve FSM apply performance through batching. Implementing the `BatchingFSM` interface enables this new feature [[GH-364](https://github.com/hashicorp/raft/pull/364)]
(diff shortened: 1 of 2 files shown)


banks:
🤔 Only concern here is I'm not 100% sure how most backends will handle a type change without a name change. Even if the timeseries store is smart enough to realise it's a different type with same name and so change processing and storage, existing dashboards that might rely on jus plotting the gauge value may well end up broken because now they are attempting to plot a single value when the actual metric is a histogram and needs some aggregation function.

I'm also not sure if it matters in that a rename will also break the dashboard in a different way - just no data any more.

I'm also not sure if it's worth going to the effort of testing this on lots of different backends to understand the subtleties. 

DD has this useful documentation: https://docs.datadoghq.com/developers/metrics/#modify-a-metric-s-type so at least for them you'd need to manual go and change the type and it would make all the old data for that metric garbage.

What do we think about choosing new names for these so that they get picked up as the new type automatically and people can just switch to the new name to update dashboard? That seems safer to me and simpler to document (use X now instead of Y) rather than trying to help users understand the subleties of type change in their TSDB/APM of choice?

I'd vote for `raft.commitLogsBatchSize` and `raft.dispatchLogsBatchSize` we could also change the new histogram you added for batching to be compatible: `raft.fsmApplyLogsBatchSize` or whatever?

PR #375, opened 2019-11-06
Fix: Failed to build since the inappropriate logger parameter

bufferx:
Just experienced one bug when execute the command "go install -v ./...", which shown that "fuzzy/node.go:24:26: undefined: log".

Seems that the parameter type of "logger" is not correct in "newRaftNode" function and we may prefer to use "hc.Logger" here.


s-christoff:
Hey, thanks for this! Good catch!

Issue #356, opened 2019-07-24
Stats calls take arbitrarily long when raft is being used heavily.


banks:
In Consul we often see any endpoint that needs to fetch raft stats like `/agent/self` hang for seconds or minutes on server nodes under write load.

It's usually a symptom of some other problem with raft or resource contention but looking at the code here: https://github.com/hashicorp/raft/blob/ff523e1c1515a5b33cc3153a34c984093fed06bb/api.go#L995-L1004

I notice a few ways we could make this better:
 - we call `getLast*` a few times which both take an exclusive lock on `lastLock`. I doubt this is highly contended though
 - we wait for configuration which comes through the main raft loop. If raft is busy restoring a snapshot or handling lots of diskIO then this can likely take a long time. I'm not sure of a simple answer here but we might get away with caching the config and pushing updates to it on reconfigurations then being able to read that under lock rather than wait for the main thread every time
 
I guess the second point is the real root cause during our observed issues.

catsby:
Hey @banks - I poked at the second point here for a bit, but I hit a wall. The problem I see is that `GetConfiguration()` and `future.Configuration()` end up calling [Configurations.Clone](https://github.com/hashicorp/raft/blob/c53e736e2b2a821b67f9c798793844b3fffd706c/configuration.go#L157) and/or [Configuration.Clone](https://github.com/hashicorp/raft/blob/c53e736e2b2a821b67f9c798793844b3fffd706c/configuration.go#L80), which both copy over an array of Servers. I’m not sure how you can cache those as the returns are expected to be deep copies. At least, the tests fail if they are not deep copies, meaning we can’t cache and return the same copy multiple times.

Am I missing something obvious, or is there a way to do this that I don't know?

catsby:
Related issue:

- https://github.com/hashicorp/raft/issues/302

banks:
@catsby I didn't have a firm plan or detailed look but I was thinking you would effectively denomalise that result into one that can be read under a lock outside of the main loop and then have the main loop update it (under lock) when it changes rather than have to wait for the main loop every time it needs to be read from. Does that make sense?

catsby:
Hey `stale` - yeah still planning on fixing this, thanks

banks:
This came up again in a new form today.

[...]

PR #379, opened 2020-01-07
Read latest configuration independently from main loop

### LINKED ISSUE #302: Too long Raft.Stats call because of not stable in time AppendEntries.StoreLogs routine
I have encountered a problem on cloud virtualized storage that the rpc `Raft.Stats` called from `consul` may periodically stall. As a consequence consul leader emits to log messages about not healthy followers at the same time leaving the cluster is safe.

The primary investigation has revealed that `Raft.Stats` stalls on getting configuration of raft node (`ConfigurationFuture` wrapper) that deals with follower loop (through `configurationsCh` channel inside `runFollower` routine). AFAIC each request to Follower including heartbeats and raft RPCs are handled sequentially. The output of RPC time metrics have shown that the `appendEntries` rpc or, more precisely `storeLogs` stage has wide spread of latency. It happens because of not stable synchronization of logs latency to persistent storage (`fdatasync` syscall in BoltDB storage backend). And separate measurement of `fdatasync` latency confirmed this hypothesis.

It's a problem of cloud provider. But such issue exposes the looseness of architecture - lightweight read requests to Follower have to wait block ones. What about not blocking reads based on snapshot before commit of logs (such as in MVCC scheme)? Is it possible and could be implemented?

i0rek:
Before reading the raft configuration required the main loop to work off the `configurationsCh`. This is problematic in different scenarios:

* node is under high load
* leadership is flapping

The main reason why that was the case is that the `configurations` attribute must not be accessed in parallel which was guaranteed because it was only ever read or written from the `leaderLoop`.

By adding a copy of the latest configuration (which is ultimately what was returned) and protecting it with its own mutex, it is now possible to access the latest configuration from outside of the leader loop.

Fixes #302, #356, and improves https://github.com/hashicorp/consul/issues/6852.

### DIFF
--- a/api.go
+++ b/api.go
@@ -136,7 +136,10 @@ type Raft struct {
 
 	// Tracks the latest configuration and latest committed configuration from
 	// the log/snapshot.
-	configurations          configurations
+	configurations configurations
+
+	// Holds a copy of the latest configuration which can be read
+	// independently from main loop.
 	latestConfiguration     Configuration
 	latestConfigurationLock sync.RWMutex
 
@@ -747,9 +750,8 @@ func (r *Raft) VerifyLeader() Future {
 	}
 }
 
-// GetConfiguration returns the latest configuration and its associated index
-// currently in use. This may not yet be committed. This must not be called on
-// the main thread (which can access the information directly).
+// GetConfiguration returns the latest configuration. This may not yet be
+// committed. The main loop can access this directly.
 func (r *Raft) GetConfiguration() ConfigurationFuture {
 	configReq := &configurationsFuture{}
 	configReq.init()
--- a/api.go
+++ b/api.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	metrics "github.com/armon/go-metrics"
@@ -140,8 +141,7 @@ type Raft struct {
 
 	// Holds a copy of the latest configuration which can be read
 	// independently from main loop.
-	latestConfiguration     Configuration
-	latestConfigurationLock sync.RWMutex
+	latestConfiguration atomic.Value
 
 	// RPC chan comes from the transport layer
 	rpcCh <-chan RPC
--- a/api.go
+++ b/api.go
@@ -136,7 +136,9 @@ type Raft struct {
 
 	// Tracks the latest configuration and latest committed configuration from
 	// the log/snapshot.
-	configurations configurations
+	configurations          configurations
+	latestConfiguration     Configuration
+	latestConfigurationLock sync.RWMutex
 
 	// RPC chan comes from the transport layer
 	rpcCh <-chan RPC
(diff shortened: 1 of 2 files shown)


banks:
Another pretty important issue this caused for future reference is when restoring a large snapshot the leader/follower loop might be blocked for seconds or minutes resulting in Stats() calls hanging "forever".

Thanks for getting to this @i0rek.

@catsby FYI I think you were looking into this at some point so just flagging the PR in case you missed it and are interested.

api.go
```diff
@@ -138,6 +138,11 @@ type Raft struct {
 	// the log/snapshot.
 	configurations configurations
 
+	// Holds a copy of the latest configuration which can be read
+	// independently from main loop.
+	latestConfiguration     Configuration
```
banks: Should we be returning the latest configuration or the latest _comitted_ configuration? I may well be misremembering but I feel like we shouldn't be using/returning the config until it's committed right?

api.go
```diff
@@ -746,19 +750,14 @@ func (r *Raft) VerifyLeader() Future {
...
-	select {
-	case <-r.shutdownCh:
-		configReq.respond(ErrRaftShutdown)
-		return configReq
-	case r.configurationsCh <- configReq:
-		return configReq
-	}
+	configReq.configurations = configurations{latest: r.getLatestConfiguration()}
+	configReq.respond(nil)
+	return configReq
```
banks: Does this make `case future := <-r.configurationsCh:` in leader/follower loop dead code now? Should we remove that?

api.go
banks: nm. `GetConfiguration` docs state:

> GetConfiguration returns the latest configuration. This may not yet be currently in use. This may not yet be committed...

SO this preserves that behaviour although frankly that seems strange to me that Consul or an application would "see" a config that tis not the one actually in use and may never get committed by a quorum 🤔.

I guess that is an edge case issue we can solve separately though this preserves same behaviour while fixing the performance issue.

api.go
```diff
@@ -138,6 +138,11 @@ type Raft struct {
 	// the log/snapshot.
 	configurations configurations
 
+	// Holds a copy of the latest configuration which can be read
+	// independently from main loop.
+	latestConfiguration     Configuration
+	latestConfigurationLock sync.RWMutex
```
banks: I think this would probably work well for most use cases but there is a risk that very heavy reads of `GetConfiguration` can starve the main loop on a config change.

For example in Consul if there were some script calling `agent info` on a server that accidentally got into a hot loop and called it constantly, it could bring the whole of raft grinding to a halt. While that's hopefully not super likely it feels like a pretty bad outcome.

I think in this case `atomic.Value` might be a better way to syncronize Load and Store of this copy. Then readers can read without blocking using `latestConfiguration.Load()` and high contention might slow the leaderloop a little due to cross-processor core contention on the cache line but it can't starve it indefinitely.

What do you think?

api.go
i0rek: We are still using it in our tests and adding an alternative would be more work. I decided to leave it as is so that this PR doesn't get out of hand. But it is essentially dead code now.

api.go
i0rek: I was wondering about the same thing, but wanted to preserve current behaviour.

api.go
i0rek: Thanks for the suggestion, sounds like a good solution. I forgot about the atomic package, but I too was wondering about contention. I am changing it to atomic!

api.go
i0rek: In theory `configurations` could be an atomic value, and we only ever load and store. That way I could get rid of the channel and fix the tests. But that would be quite some places to change.

api.go
i0rek: I agree that it would be better to use atomic and I will change it.

api.go
banks: 🤔 hmm. Making part of the internal workings atomic and accessing it directly in tests could introduce other issues though. Atomics prevent data races but not logical ones. I think the usage here is fine as we are limiting to only a single writer and wrapping concurrent readers so they don't have direct access but making internal state atomic and then tweaking it from tests could get brittle fast if we ever update or make assumptions about whether then atomic value.

Does leaving tests using a different code path to real callers cause issues? If not I'm inclined to leave it to.

Oh we also still use it in `takeSnapshot`. We also rely in those places on being able to get both the latest and the committed configs which means the GetConfiguration can't do the same job.

So I say it's OK to leave it for now as it is as those call sites need to lower level and more accurate answer, this just optimises for external callers.

api.go
```diff
@@ -138,6 +139,10 @@ type Raft struct {
 	// the log/snapshot.
 	configurations configurations
 
+	// Holds a copy of the latest configuration which can be read
+	// independently from main loop.
+	latestConfiguration atomic.Value
```
banks: For the record.. I looked it up and `atomic.Value` is itself a struct which means even as a non-pointer field like this it will be aligned to 64 bit boundaries and the atomically accessed pointer within it won't have alignment issues.

api.go
i0rek: 👍

PR #384, opened 2020-01-23
raft.LeaderCh() always deliver latest transition

notnoop:
This makes leaderCh always deliver the latest leadership transition
rather than dropping messages somewhat unexpectedly

Managing leadership transition is quite fragile and easy to get wrong.
Prior to this, `LeaderCh()` is unreliable and is prone to misuse.
For example, both Consul[1] and Nomad[2] started with LeaderCh and
ended up switching to manage their own channels after hitting issues.
But managing reliable leadership channels is tricky; both Consul[3] and
Nomad[4] had deadlock bugs related to leadership flagging!

Here, we aim to improve the library ergonomics for future library
applications as well as Nomad, by having stricter guarantees of
`raft.LeaderCh`.  Namely, receivers are guaranteed to receive a
notification if leadership transition event occured, and slow receivers
will be guaranteed to receive the latest view since their last receive*.

This seems like a reasonable interface for practical purposes.  If a
leadership flapped many times but but ended as a non-leader ultimately,
the receiver should just recored the final step down without attempting
to handle the intermediary states events.

One edge case is if a leader lost but then regained leadership quickly. The
application may need to issue a Barrier call and reconcile its state.
Thus, the library should still send notification events even if final
leadership value matches last delivered value.

* note that due to concurrency, last received value might still be
stale, if raft just recognized transition event but didn't send yet.
Receivers on a loop will get the event in subsequent receive.

[1] https://github.com/hashicorp/consul/pull/2911
[2] https://github.com/hashicorp/nomad/commit/c2b7ce92e79f2888964270228011ce1555dfb86d
[3] https://github.com/hashicorp/consul/issues/6852
[4] https://github.com/hashicorp/nomad/issues/4749

### DIFF
--- a/api.go
+++ b/api.go
@@ -503,7 +503,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		fsm:                   fsm,
 		fsmMutateCh:           make(chan interface{}, 128),
 		fsmSnapshotCh:         make(chan *reqSnapshotFuture),
-		leaderCh:              make(chan bool),
+		leaderCh:              make(chan bool, 1),
 		localID:               localID,
 		localAddr:             localAddr,
 		logger:                logger,
@@ -959,10 +959,17 @@ func (r *Raft) State() RaftState {
 	return r.getState()
 }
 
-// LeaderCh is used to get a channel which delivers signals on
-// acquiring or losing leadership. It sends true if we become
-// the leader, and false if we lose it. The channel is not buffered,
-// and does not block on writes.
+// LeaderCh is used to get a channel which delivers signals on acquiring or
+// losing leadership. It sends true if we become the leader, and false if we
+// lose it.
+//
+// Receivers can expect to receive a notification only if leadership
+// transition has occured.
+//
+// If receivers aren't ready for the signal, signals may drop and only the
+// latest leadership transition. For example, if a receiver receives subsequent
+// `true` values, they may deduce that leadership was lost and regained while
+// the the receiver was processing first leadership transition.
 func (r *Raft) LeaderCh() <-chan bool {
 	return r.leaderCh
 }
--- a/raft.go
+++ b/raft.go
@@ -364,7 +364,7 @@ func (r *Raft) runLeader() {
 	metrics.IncrCounter([]string{"raft", "state", "leader"}, 1)
 
 	// Notify that we are the leader
-	asyncNotifyBool(r.leaderCh, true)
+	overrideNotifyBool(r.leaderCh, true)
 
 	// Push to the notify channel if given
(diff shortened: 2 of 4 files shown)


api.go
```diff
@@ -959,10 +959,17 @@ func (r *Raft) State() RaftState {
...
-// and does not block on writes.
+// LeaderCh is used to get a channel which delivers signals on acquiring or
+// losing leadership. It sends true if we become the leader, and false if we
+// lose it.
+//
+// Receivers can expect to receive a notification only if leadership
+// transition has occured.
+//
+// If receivers aren't ready for the signal, signals may drop and only the
+// latest leadership transition. For example, if a receiver receives subsequent
```
banks:
```suggestion
// latest leadership transition will be received. For example, if a receiver receives consecutive
```

notnoop:
@banks Thanks for the review - yeah, I can spill the need to handle transition more specifically.  In Nomad land, I implemented this approach, just outside the library in https://github.com/hashicorp/nomad/pull/6977/files (would love your review there too :P).  I was hoping that we'd simply nomad's logic by incorporating the leadership channel behavior here and rely on this instead.  Based on our experience, requiring clients to configure their reliable channels properly is tricky.

Another alternative that may relieve consumers more and be less prone to misuse is to common and "best practices" bits of Consul or Nomad's monitorLoop and leaderLoop into the library and expose a `LeadershipUpdateDelegate` for application specific logic.  If this is something worth checking, I can implement it.

I'm motivated to address as much of this in raft library, because managing leadership transitions is very subtle now - and exposing a better api will hopefully avoid having other applications (Vault?!) rediscover Consul/Nomad bugs and easies fixing them globally.

banks:
Got it,

I'm definitely in favour of making the raft lib easier for other
implementers to use correctly, I guess I wonder if this is enough better
considering the still hefty gotchas involved to use it correctly.

Perhaps encapsulating the way Consul and Nomad do this via notifyCh in the
lib like you suggest is a better improvement?

notnoop:
> Perhaps encapsulating the way Consul and Nomad do this via notifyCh in the
> lib like you suggest is a better improvement?

Maybe - Defining the right interface isn't obvious.  Nomad and Consul monitor leadership loop have other concerns (reconciling Serf, tracking members, and enabling ACL progress), and I'm

i0rek:
I was wondering if this change would be technically speaking backwards incompatible. `leaderCh` was dropping messages before and still does, just other messages. From the outside the message dropping was unpredictable and now we are conscious about it. From an API perspective I don't see why this change would be a problem. 

> This change makes `leaderCh` useful (previous version was unusable)

agreed.

> also in extreme flapping, it does redundant unnecessary transitions

I was thinking that if we would drain notifyCh before doing anything, we could get the latest state as well and act on it. Avoiding unnecessary transitions.

PR #390, opened 2020-02-26
futures can react to shutdown

### LINKED ISSUE #268: The Raft.Shutdown() API can end up in a deadlock

i0rek:
This is implementing a very similar thing to https://github.com/hashicorp/raft/pull/277 but a little bit more generalized, like suggested by @mkeeler.

Fixes #277, fixes #268.

### DIFF
--- a/future.go
+++ b/future.go
@@ -84,9 +84,10 @@ func (e errorFuture) Index() uint64 {
 // deferError can be embedded to allow a future
 // to provide an error in the future.
 type deferError struct {
-	err       error
-	errCh     chan error
-	responded bool
+	err        error
+	errCh      chan error
+	responded  bool
+	ShutdownCh chan struct{}
 }
 
 func (d *deferError) init() {
@@ -103,7 +104,11 @@ func (d *deferError) Error() error {
 	if d.errCh == nil {
 		panic("waiting for response on nil channel")
 	}
-	d.err = <-d.errCh
+	select {
+	case d.err = <-d.errCh:
+	case <-d.ShutdownCh:
+		d.err = ErrRaftShutdown
+	}
 	return d.err
 }
 
--- a/snapshot.go
+++ b/snapshot.go
@@ -146,6 +146,7 @@ func (r *Raft) takeSnapshot() (string, error) {
 	// We have to use the future here to safely get this information since
 	// it is owned by the main thread.
 	configReq := &configurationsFuture{}
+	configReq.ShutdownCh = r.shutdownCh
 	configReq.init()
 	select {
 	case r.configurationsCh <- configReq:


snapshot.go
```diff
@@ -146,6 +146,7 @@ func (r *Raft) takeSnapshot() (string, error) {
 	// We have to use the future here to safely get this information since
 	// it is owned by the main thread.
 	configReq := &configurationsFuture{}
+	configReq.ShutdownCh = r.shutdownCh
```
notnoop: It may be nice to document when `ShutdownCh` needs to be set and what makes snapshotting special.  I assume it's because snapshotting runs in an internal gorotouine and can run regardless of leadership.  Other futures will be responded to immediately when not leader, or when raft loses leadership as a result of shutting down?

future.go
```diff
@@ -84,9 +84,10 @@ func (e errorFuture) Index() uint64 {
 // deferError can be embedded to allow a future
 // to provide an error in the future.
 type deferError struct {
-	err       error
-	errCh     chan error
-	responded bool
+	err        error
+	errCh      chan error
+	responded  bool
+	ShutdownCh chan struct{}
```
notnoop: very nit picky - can this field be private like the rest?

PR #391, opened 2020-02-27
fix TestRaft_LeadershipTransferLeaderRejectsClientRequests test

i0rek:
TestRaft_LeadershipTransferLeaderRejectsClientRequests was failing because `GetConfiguration` is no longer dependent on the mainloop and can be retrieved even while leadership transfer!

TestRaft_LeaderLeaseExpire is now passing because I changed the code to wait for the leader to step down. Maybe the recent leaderch changes changes something. 🤷‍♂

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -2027,7 +2027,6 @@ func TestRaft_LeadershipTransferLeaderRejectsClientRequests(t *testing.T) {
 		l.Apply([]byte("test"), 0),
 		l.Barrier(0),
 		l.DemoteVoter(ServerID(""), 0, 0),
-		l.GetConfiguration(),
 
 		// the API is tested, but here we are making sure we reject any config change.
 		l.requestConfigChange(configurationChangeRequest{}, 100*time.Millisecond),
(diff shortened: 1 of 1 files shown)


raft_test.go
```diff
@@ -1490,13 +1490,17 @@ func TestRaft_LeaderLeaseExpire(t *testing.T) {
 	c.Disconnect(follower.localAddr)
 
 	// Watch the leaderCh
-	select {
-	case v := <-leader.LeaderCh():
-		if v {
-			c.FailNowf("should step down as leader")
+	timeout := time.After(conf.LeaderLeaseTimeout * 2)
```
s-christoff: Would we be able to not use this timeout by implementing a retry testutil? Would love to see that improvement. :)

notnoop: Sadly, a retry testutil helper function wouldn't make this necessarily cleaner, since we are also blocking on a channel receive.

i0rek:
Merged since @notnoop is right re retry helper.

PR #398, opened 2020-04-08
Ignore non-voters in leadership transfer

f3nry:
A leader should not initiate an election on a non-voter, causing it to temporarily become a leader for a short period of time.

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -2094,6 +2094,29 @@ func TestRaft_LeadershipTransferLeaderReplicationTimeout(t *testing.T) {
 	}
 }
 
+func TestRaft_LeadershipTransferIgnoresNonvoters(t *testing.T) {
+	c := MakeCluster(2, t, nil)
+	defer c.Close()
+
+	follower := c.Followers()[0]
+
+	demoteFuture := c.Leader().DemoteVoter(follower.localID, 0, 1*time.Nanosecond)
+	if demoteFuture.Error() != nil {
+		t.Fatalf("demote voter err'd: %v", demoteFuture.Error())
+	}
+
+	future := c.Leader().LeadershipTransfer()
+	if future.Error() == nil {
+		t.Fatal("leadership transfer should err")
+	}
+
+	expected := "cannot find peer"
+	actual := future.Error().Error()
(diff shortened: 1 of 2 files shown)


raft_test.go
```diff
@@ -2094,6 +2094,29 @@ func TestRaft_LeadershipTransferLeaderReplicationTimeout(t *testing.T) {
 	}
 }
 
+func TestRaft_LeadershipTransferIgnoresNonvoters(t *testing.T) {
+	c := MakeCluster(2, t, nil)
+	defer c.Close()
+
+	follower := c.Followers()[0]
+
+	demoteFuture := c.Leader().DemoteVoter(follower.localID, 0, 1*time.Nanosecond)
```
briankassouf, line 2103: Could we make this a larger timeout or no timeout? Seems like 1 nanosecond could cause this to be flakey if go doesn't schedule the receiver right away.

f3nry: Thanks for the feedback - I changed this to zero.

f3nry:
@briankassouf Thank you for the review! I'm separately working on signing the CLA.

I'm wondering if there's additional places where we'd want to protect against this occurring or if this is sufficient. For example, should a node to reject a TimeoutNow request if it is currently a non-voter? Should it be an invalid state transition to become a candidate if a node doesn't have a vote?

i0rek:
@f3nry I think you are correct, a non-voter should not become a leader and should reject leadership transfers. 

There is an issue in Consul for your PR: https://github.com/hashicorp/consul/issues/7031. And another one that is similar: https://github.com/hashicorp/consul/issues/6979. The first one should be fixed by this PR. The last one is still open.

f3nry:
@i0rek It seems like the [second bug](https://github.com/hashicorp/raft/blob/72a6ed385dac0c8ee1628d12a9f0bb99abf39f19/raft.go#L216) is mostly protected by the configuration index check on line 207.. Under what conditions would `r.configurations.latestIndex == r.configurations.committedIndex` not pass?

f3nry:
@briankassouf CLA is signed and PR is ready to go!

briankassouf:
@f3nry Great! Thanks!

PR #399, opened 2020-04-16
more efficient append bytes

Himer:
` io.Copy(s.contents, bytes.NewReader(p)) ` , why use this way?


s-christoff:
Thank you for this!

PR #400, opened 2020-05-01
Return decoded candidate address on duplicate requestVote log

calvn:
The log returns the encoded candidate address on the `duplicate requestVote from` warning message which is not very useful during triage:
```
2020-04-28T18:21:24.265Z [WARN]  storage.raft: duplicate requestVote from: candidate=[91, 58, 58, 93, 58, 56, 50, 48, 49]
```

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1452,7 +1452,7 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 	if lastVoteTerm == req.Term && lastVoteCandBytes != nil {
 		r.logger.Info("duplicate requestVote for same term", "term", req.Term)
 		if bytes.Compare(lastVoteCandBytes, req.Candidate) == 0 {
-			r.logger.Warn("duplicate requestVote from", "candidate", r.trans.DecodePeer(req.Candidate))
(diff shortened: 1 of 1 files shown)


raft.go
```diff
@@ -1452,7 +1452,7 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 	if lastVoteTerm == req.Term && lastVoteCandBytes != nil {
 		r.logger.Info("duplicate requestVote for same term", "term", req.Term)
 		if bytes.Compare(lastVoteCandBytes, req.Candidate) == 0 {
-			r.logger.Warn("duplicate requestVote from", "candidate", req.Candidate)
+			r.logger.Warn("duplicate requestVote from", "candidate", r.trans.DecodePeer(req.Candidate))
```
s-christoff, line 1455: It looks like [something similar](https://github.com/hashicorp/raft/blob/e51cd9f557cc5e879cd8e41f4054725a41f262d0/raft.go#L1417) is done a couple lines up - could we use the `candidate` value instead?

calvn: Good catch, sure thing!

Issue #401, opened 2020-05-04
Get unavailable nodes from the leader?


ickby:
Currently the API does not support to get feedback on the state of nodes in the cluster. With the hearbeats send by the followers there should be a way to detect missing nodes by the leader, and that quite fast. It would be interesting to get this information, maybe in the observer. Do I overlook a possibility to get the info, or could this be added if not?

I want to implement auto-manage of a cluster, which removes nodes if they are not working anymore. But I need way to detect nodes failing, and it seems counter intuitive to add an additional heartbeat next to rafts one.

Wicked7000:
Just as an additional bump I'm also looking for this behaviour, currently when a node fails the cluster constantly keeps pinging it (When after X amount of time I'd just like to remove it from the cluster). It seems as though this information used to be available via PeerStore but no longer is?

szab100:
bump! this is quite important, I'd like to do the same, monitor missing nodes and auto-remove them.. this way I can have a stateless replicaSet/deployment with an init script that takes the headless service's DNS hostname that resolves to all active pods, iterate over them, find the first leader and join as follower ==> zeroconf auto cluster grouping.. But this cannot work properly without auto-removal..

ickby:
Resolved by #405. Thanks!

PR #405, opened 2020-06-01
Observe missed heartbeats

mattwalo32:
From the perspective of a Raft API user, there is currently no way to detect if the leader is unable to contact nodes. This PR allows users to register an observer that will make observations whenever the leader cannot heartbeat with a node.

Fixes issue #401

### DIFF
--- a/observer.go
+++ b/observer.go
@@ -1,8 +1,8 @@
 package raft
 
 import (
-	"time"
 	"sync/atomic"
+	"time"
 )
 
 // Observation is sent along the given channel to observers when an event occurs.
@@ -30,7 +30,7 @@ type PeerObservation struct {
 
 // FailedHeartbeatObservation is sent when a node fails to heartbeat with the leader
 type FailedHeartbeatObservation struct {
-	PeerID ServerID
+	PeerID      ServerID
 	LastContact time.Time
 }
 
--- a/observer.go
+++ b/observer.go
@@ -27,8 +27,8 @@ type PeerObservation struct {
 	Peer    Server
 }
 
-// HeartbeatObservation is sent when a node fails to heartbeat with the leader
-type HeartbeatObservation struct {
+// MissedHeartbeatObservation is sent when a node fails to heartbeat with the leader
+type MissedHeartbeatObservation struct {
 	PeerID ServerID
 }
 
--- a/observer.go
+++ b/observer.go
@@ -27,6 +27,7 @@ type PeerObservation struct {
 	Peer    Server
 }
 
+// HeartbeatObservation is sent when a node fails to heartbeat with the leader
 type HeartbeatObservation struct {
 	Peer ServerAddress
 }
--- a/observer.go
+++ b/observer.go
@@ -1,6 +1,7 @@
 package raft
 
 import (
+	"time"
 	"sync/atomic"
 )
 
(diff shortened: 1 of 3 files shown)


observer.go
```diff
@@ -27,6 +27,11 @@ type PeerObservation struct {
 	Peer    Server
 }
 
+// HeartbeatObservation is sent when a node fails to heartbeat with the leader
+type HeartbeatObservation struct {
```
fastest963, line 31: I wonder if instead we should call this `FailedHeartbeatObservation` or if we should instead always send an observation for every heartbeat and include a `Failed` boolean

mattwalo32: Yeah, I think `FailedHeartbeatObservation` would be a better choice. It seems like there could be a bit of overhead if we observed every heartbeat, especially if the heartbeat interval is low. Also I'd imagine that there aren't many cases where developers want to observe successful heartbeats.

fastest963:
@briankassouf Can you look at this? Pretty small change overall and would be a huge help to anyone trying to manage their cluster automatically. Right now we have to do duplicate heartbeats outside of raft just to understand which nodes are alive. It would be simpler to just get insight into the heartbeats that raft is already doing.

raft.go
```diff
@@ -881,6 +881,8 @@ func (r *Raft) checkLeaderLease() time.Duration {
 				} else {
 					r.logger.Debug("failed to contact", "server-id", server.ID, "time", diff)
 				}
+
+				r.observe(FailedHeartbeatObservation{PeerID: server.ID})
```
briankassouf, line 885: What do you think about adding the LastContact time for this peer here? That way decisions can be made based on how long this node has been unreachable.

mattwalo32: That's a good addition, I'll put that in

fastest963:
@briankassouf @mattwalo32 actually, where this was added only works for voters. Would it be better to put this in `Raft.heartbeat` instead after AppendEntries fails?

raft.go
```diff
@@ -881,6 +881,8 @@ func (r *Raft) checkLeaderLease() time.Duration {
 				} else {
 					r.logger.Debug("failed to contact", "server-id", server.ID, "time", diff)
 				}
+
+				r.observe(FailedHeartbeatObservation{PeerID: server.ID, LastContact: f.LastContact()})
```
jameshartig, line 885: Did you see my comment on the PR? I'm concerned that this only applies to voters (see line 865). When instead if this was done in the `heartbeat` method it would apply to any server.

mattwalo32: Yea, I'll add that in

jameshartig (approved):
LGTM

jameshartig:
This should be good to merge @briankassouf

tylertreat:
Would it be possible to somehow extend this to add an Observation for nodes that come back online after failing a heartbeat? I have a use case where I'd like to run some logic both when a follower times out _and_ if/when it comes back (e.g. imagine a case where a server is restarted). This would probably require the leader to track "failed" nodes in order to determine when they come back online.

ickby:
Could this be merged? I'm very interested in the change.

s-christoff (approved):
🤙

s-christoff:
Howdy, thank you for this contribution. Please update [the documentation ](https://pkg.go.dev/github.com/hashicorp/raft#Observation) for this change as well if you could and I'll get this change merged.

PR #409, opened 2020-07-08
Use peer id as metric label

coignetp:
Hey! 
I'm currently working on the consul integration at Datadog, and we want to scrape the prometheus endpoint to get consul metrics. In the current state several metrics are not consistently named, and contain parts that would be good as labels instead, and that makes scraping the endpoint and converting to Datadog metrics troublesome.
This PR use the peer ids as label and remove them from the metric names.

Example: `consul_raft_replication_appendEntries_logs_684ad51c_6143_f2ea_aaaa_8ab6cc5ccccc` -> `consul_raft_replication_appendEntries_logs`.
This PR affects:
* `consul_raft_replication_installSnapshot`
* `consul_raft_replication_heartbeat`
* `consul_raft_replication_appendEntries_rpc`
* `consul_raft_replication_appendEntries_logs`

This change is not backward compatible since the metric names will be different. The current metric parsing method won't work anymore.

Please tell me what you think about this change! 😄 

P.S. : I'd like to do the same thing for other consul metrics (here for example https://github.com/hashicorp/consul/blob/5e5dbedd47a5f875b60e241c5555a9caab595246/agent/http.go#L258)

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -328,6 +328,8 @@ func (r *Raft) sendLatestSnapshot(s *followerReplication) (bool, error) {
 	}
 	labels := []metrics.Label{{Name: "peer_id", Value: string(s.peer.ID)}}
 	metrics.MeasureSinceWithLabels([]string{"raft", "replication", "installSnapshot"}, start, labels)
+	// Duplicated information. Kept for backward compatibility.
+	metrics.MeasureSince([]string{"raft", "replication", "installSnapshot", string(s.peer.ID)}, start)
 
 	// Check for a newer term, stop running
 	if resp.Term > req.Term {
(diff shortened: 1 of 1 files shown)


coignetp:
Is the CI failing because of my change? I'm not sure I understand the error here

s-christoff:
Howdy - this is great! Thank you so much. 
Is there any way we could maintain the original metrics and add the new metrics as well to maintain backwards compatibility?

coignetp:
I updated the PR to keep sending both metrics

coignetp:
Should I do something more on this change? I'm not sure if the CI is failing because of this PR here🙂

s-christoff:
Howdy friend,
Thank you for your patience and hard work on this. It's started some internal discussions - which is great! We're (Consul) looking into deprecating other metadata-suffixed metrics in Consul and we are working through the details. We're going to figure some stuff out on our end, and will be looking to accept this in a few weeks.

coignetp:
Hi @s-christoff, is there any news about this?

coignetp:
Hi @s-christoff, sorry to bother you again, but how do you have any news here?

s-christoff (approved):
lgtm - sorry for the delay, lots of internal conversations happening. :p

PR #412, opened 2020-07-19
leader notifies followers for commitIndex in heartbeat rpc

vision9527:
Hello. I think that the leader will not notify followers specially for commitIndex  except normal AppendEntriesRpc(not heartbeat), it will bring about follower's commitIndex is not latest. 
In the raft paper(section 5.3) it suggests the leader keeps track of the highest index it knows to be committed, and it includes that index in future AppendEntries RPCs (including heartbeats) so that the other servers eventually find out.
Is it right in my understand?
image

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -375,6 +375,10 @@ func (r *Raft) heartbeat(s *followerReplication, stopCh chan struct{}) {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


vision9527:
The change has to do with this error?

vision9527:
I see

PR #413, opened 2020-07-20
Reject leadership transfer requests when in candidate state

briankassouf:
If `raft.LeadershipTransfer().Error()` is called while raft is in candidate state the caller will block until it transitions to follower or leader. If raft is in an error state and unable to elect a leader this could block indefinitely. This PR causes `LeadershipTransfer()` called on a candidate node to immediately return an error.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -311,6 +311,10 @@ func (r *Raft) runCandidate() {
 			// Reject any restores since we are not the leader
 			r.respond(ErrNotLeader)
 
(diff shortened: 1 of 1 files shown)


Legogris:
Do I understand it right if this is a fix for https://github.com/hashicorp/consul/issues/7750?

If so, is there a plan for making a new release of raft and upgrading raft in consul/nomad/vault?

We currently have a nomad cluster unoperational with the exact symptoms as the consul issue above.

PR #416, opened 2020-08-18
Move peer id from metric name to labels in raft replication

satyanash:
This PR builds on top of and closes #365 
- It emits new replication metrics which move the Peer ID into a label, from the metric name.
- As per the discussion with @banks in #365, we're keeping the old one around for backwards compatibility with existing dashboards.
- This issue has existed for a while now and has also been reported on the consul repo. (https://github.com/hashicorp/consul/issues/4450)

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -7,7 +7,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/armon/go-metrics"
+	metrics "github.com/armon/go-metrics"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


banks:
Hey @satyanash, thanks for making this PR.

I must apologise though - after having said this seemed like a reasonable approach and encouraging you to make a new PR I discovered we have someone on the team currently writing a proposal for a more holistic approach for how we enhance and deprecate metrics and potentially even change our metrics libraries. Hopefully that work will be done in the next few weeks and be in the next major release of our projects anyway.

Given that's already under discussion across a few teams here, we'll have to hold on merging this until we've decided overall on the best approach for evolving metrics, then hopefully we can either merge it or make the equivalent change under that new design.

Thanks again for the contribution though and again, my apologies for steering you this way!

mikemorris:
Is this fixed with the merge of https://github.com/hashicorp/raft/pull/409?

PR #419, opened 2020-08-25
startStopReplication: Handle Existing Peer Update

alexanderbez:
Currently, when `Raft#startStopReplication` is executed from a config change update, it handles two cases: adding a new peer or removing an existing peer. However, if a configuration change  resulted in a peer change (e.g. server address changed), `Raft#startStopReplication` won't catch this and heartbeats will be sent to the old address.

e.g.
1. Create a cluster {A, B, C}
2. Stop node A & wait for view change
3. Start A with new IP address
4. Observe `AddStaging` is handled by `Raft#appendConfigurationEntry`
5. Observe new configuration is persisted
6. Observe `Raft#startStopReplication` is called, but Raft keeps A  with the old server address
7. Observe heartbeats timeout

I'm not sure if we need to cleanup any routines or replication, so I just have the `followerReplication.server` field updated for now.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -473,10 +473,13 @@ func (r *Raft) startStopReplication() {
 		if server.ID == r.localID {
 			continue
 		}
+
 		inConfig[server.ID] = true
-		if _, ok := r.leaderState.replState[server.ID]; !ok {
+
+		s, ok := r.leaderState.replState[server.ID]
+		if !ok {
 			r.logger.Info("added peer, starting replication", "peer", server.ID)
-			s := &followerReplication{
+			s = &followerReplication{
 				peer:                server,
 				commitment:          r.leaderState.commitment,
 				stopCh:              make(chan uint64, 1),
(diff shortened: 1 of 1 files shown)


raft.go
```diff
@@ -489,10 +492,14 @@ func (r *Raft) startStopReplication() {
...
 				stepDown:            r.leaderState.stepDown,
 			}
+
 			r.leaderState.replState[server.ID] = s
 			r.goFunc(func() { r.replicate(s) })
 			asyncNotifyCh(s.triggerCh)
 			r.observe(PeerObservation{Peer: server, Removed: false})
+		} else if ok && s.peer.Address != server.Address {
+			r.logger.Info("updating peer", "peer", server.ID)
+			s.peer = server
```
briankassouf, line 502: I'm wondering if you need to call `asyncNotifyCh(s.triggerCh)` like we are above?

alexanderbez: Perhaps? What does that do? is it idempotent?

briankassouf: it tells raft to kick off replication to this server

https://github.com/hashicorp/raft/blob/ff338ca008a48ab472f7a9e748ad10db15c462c9/replication.go#L156-L158

alexanderbez: I see. We're only change the peer address here. So it depends on if the replication layer will pick up this change or not? If we do call `asyncNotifyCh`, is this idempotent? Specifically, is there any harm in calling it?

alexanderbez:
Going to merge this. During my testing of this change, replication was still occurring so I believe the `asyncNotifyCh(s.triggerCh)` call is unnecessary.

PR #424, opened 2020-08-29
error style fixes

ZYunH:
Error strings should not be capitalized except it starts with a variable name.


ZYunH:
The test failed, but looks like not related to this PR.

s-christoff (approved):
Thank you so much :)

PR #425, opened 2020-09-04
Make snapshot meta size optional

ncabatoff:
Desired for Vault snapshotting where it's expensive to compute size.
Changes seem to work, but unclear if the LimitReader was necessary - is the connection re-used?


JelteF:
I would suggest you to look at this PR of mine as well, which does the same thing.

https://github.com/hashicorp/raft/pull/247

PR #427, opened 2020-09-08
Make Raft.LeaderCh() return a new channel for each invocation

### LINKED ISSUE #426: No good way to watch leadership
`Raft.LeaderCh()` has two problems: It always returns the same channel and it doesn't prepopulate the channel with the current value. Reconstructing this has proven error prone/impossible.

`RegisterObserver()` was my next attempt, but the LocalAddress isn't exposed (except by the Transport, which not be available to library code).

I propose to make Raft.LeaderCh() return a new channel per call, and prepopulate that with the current value. That's a slight behavior change, but I don't expect it to break clients.

Jille:
.. and immediately send the current leadership state over the channel.

This way it can be used by multiple pieces of code with disrupting another.

Sending the value immediately avoids strange race conditions when RaftState gets updated at a slightly different moment.

fixes #426

### DIFF
--- a/api.go
+++ b/api.go
@@ -111,6 +111,12 @@ type Raft struct {
 	// leaderCh is used to notify of leadership changes
 	leaderCh chan bool
 
+	leaderChs     []chan bool
+	leaderChsLock sync.Mutex
+	// leaderChLastMessage is the last message sent over leaderCh that has been processed.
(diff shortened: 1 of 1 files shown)


Jille:
still relevant

Jille:
still relevant, discussion in #426 got some traction the other day actually

PR #439, opened 2021-02-01
Buffer applyCh to size MaxAppendEntries

alecjen:
This change implements the solution suggested in https://github.com/hashicorp/raft/pull/124 to enable batch processing off the raft applyCh

For context, our application uses this library, and we are currently bottlenecked by StoreLogs throughput. We believe we can boost this throughput by increasing batch size in StoreLogs, which fsync's on each call.

Appreciate all the help team 💯

### DIFF
--- a/config.go
+++ b/config.go
@@ -238,20 +238,20 @@ func ValidateConfig(config *Config) error {
 	}
 	if config.ProtocolVersion < protocolMin ||
 		config.ProtocolVersion > ProtocolVersionMax {
-		return fmt.Errorf("Protocol version %d must be >= %d and <= %d",
+		return fmt.Errorf("ProtocolVersion %d must be >= %d and <= %d",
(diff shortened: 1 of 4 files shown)


alecjen:
Actually, I'll just try to fix the race condition here by using a context and done channel, as suggested

api.go
```diff
@@ -498,7 +499,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 	// Create Raft struct.
 	r := &Raft{
 		protocolVersion:       protocolVersion,
-		applyCh:               make(chan *logFuture),
+		applyCh:               make(chan *logFuture, conf.ApplyChSize),
```
corymonroe-coinbase, line 502: Should this now just be `conf.MaxAppendEntries` given that it will at most pull that many off the channel.

PR #442, opened 2021-02-26
Add String() to LogType enum for easier debugging.

icco:
Found this useful while debugging a server that was using this library.

PR #443, opened 2021-02-26
Add CircleCI tests for Go 1.15 and 1.16

icco:
Noticed while submitting https://github.com/hashicorp/raft/pull/442 that this repo is missing tests for the latest two versions of Go.


dnephin:
Thank you for the PR! It would be great to run the tests against the more recent versions of Go.

There were some test flakes, but I re-ran the tests a few times and it looks green now, so I think we can merge!

PR #444, opened 2021-02-26
Make snapshot timing and trailing logs hot-reloadable in raft

banks:
These parameters are reasonably easy to make reloadable in a running raft instance.

Other parameters like the timeouts may be too although some require more thought about whether they are safe to change at runtime, however I'm sticking with these three right now as they are the critical parameters needed to get out of an otherwise unrecoverable situation detailed in https://github.com/hashicorp/consul/issues/9609.

There are other possible mitigations and solutions to that problem including my previous WIP PR that prevents snapshotting during replication entirely, but this has fewer uncertainties about safety and is the simplest change that gives a manual intervention option where almost none exists today.

If others are happy with this change, we can merge it and then allow Consul (and others) to have a way to re-configure snapshotting and log pruning on a leader without restarting it and loosing leadership.

## Open Questions

The API/Config change was the main gross part here. I went with this as the least-bad of a few bad options but I'm open to input if others thing there is a preferable way.

I didn't want to accept a whole `Config` even though some fields are not reloadable because then you either have to explicitly copy only some fields, or require the user ensure the rest of the config is identical to the current config (which they might not know as they can't read it) otherwise Bad Things would happen.

I also didn't want to make a separate `SetX` method for every field we make reloadable as I could see why at least a handful of others might be good to be reloadable later.

I also didn't want the internal state to continue reading from a plane `conf Config` field but just have to remember not to read reloadable fields from there but somewhere else because that seems error prone. This way forces all readers to access the most recent config and hopefully helps them consider whether the thing they are reading is reloadable or not so they can make sure the logic takes that into account (i.e. use same value in multiple operations in case it changes between them).

### DIFF
--- a/api.go
+++ b/api.go
@@ -642,7 +642,9 @@ func (r *Raft) config() Config {
 	return r.conf.Load().(Config)
 }
 
-// ReloadConfig updates the configuration of a running raft node.
+// ReloadConfig updates the configuration of a running raft node. If the new
+// configuration is invalid an error is returned and no changes made to the
+// instance.
 func (r *Raft) ReloadConfig(rc *ReloadableConfig) error {
 	r.confReloadMu.Lock()
 	defer r.confReloadMu.Unlock()
--- a/api.go
+++ b/api.go
@@ -81,8 +81,15 @@ type Raft struct {
 	// be committed and applied to the FSM.
 	applyCh chan *logFuture
 
-	// Configuration provided at Raft initialization
-	conf Config
+	// conf stores the current configuration to use. This is the most recent one
+	// provided. All reads of config values should use the config() helper method
+	// to read this safely.
+	conf atomic.Value
+
+	// confReloadMu ensures that only one thread can reload config at once since
+	// we need to read-modify-write the atomic. It is NOT necessary to hold this
+	// for any other operation e.g. reading config using config().
+	confReloadMu sync.Mutex
 
 	// FSM is the client state machine to apply commands to
 	fsm FSM
@@ -385,9 +392,9 @@ func RecoverCluster(conf *Config, fsm FSM, logs LogStore, stable StableStore,
 	return nil
 }
 
-// GetConfiguration returns the configuration of the Raft cluster without
-// starting a Raft instance or connecting to the cluster
-// This function has identical behavior to Raft.GetConfiguration
+// GetConfiguration returns the persisted configuration of the Raft cluster
+// without starting a Raft instance or connecting to the cluster. This function
+// has identical behavior to Raft.GetConfiguration.
 func GetConfiguration(conf *Config, fsm FSM, logs LogStore, stable StableStore,
 	snaps SnapshotStore, trans Transport) (Configuration, error) {
 	conf.skipStartup = true
@@ -505,7 +512,6 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 	r := &Raft{
 		protocolVersion:       protocolVersion,
 		applyCh:               applyCh,
-		conf:                  *conf,
 		fsm:                   fsm,
 		fsmMutateCh:           make(chan interface{}, 128),
 		fsmSnapshotCh:         make(chan *reqSnapshotFuture),
@@ -529,6 +535,8 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		observers:             make(map[uint64]*Observer),
 		leadershipTransferCh:  make(chan *leadershipTransferFuture, 1),
 	}
+
+	r.conf.Store(*conf)
 
 	// Initialize as a follower.
 	r.setState(Follower)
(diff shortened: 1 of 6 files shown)


banks:
We are still having some issues with tests in CI, but the suite passes locally for me and the new tests I added pass using `-race` too FWIW.

config.go
```diff
@@ -157,9 +157,10 @@ type Config struct {
 	// we can become a leader of a cluster containing only this node.
 	ShutdownOnRemove bool
 
-	// TrailingLogs controls how many logs we leave after a snapshot. This is
-	// used so that we can quickly replay logs on a follower instead of being
-	// forced to send an entire snapshot.
+	// TrailingLogs controls how many logs we leave after a snapshot. This is used
+	// so that we can quickly replay logs on a follower instead of being forced to
+	// send an entire snapshot. The value passed here is the initial setting used.
+	// This can be tuned during operation using
```
banks, line 163: TODO:

- [ ]  fix this comment and docs generally if we are happy with this API approach (duplicating fields)

api.go
```diff
@@ -624,6 +632,29 @@ func (r *Raft) restoreSnapshot() error {
 	return nil
 }
 
+func (r *Raft) config() Config {
+	return r.conf.Load().(Config)
+}
+
+// ReloadConfig updates the configuration of a running raft node.
```
banks, line 639:
```suggestion
// ReloadConfig updates the configuration of a running raft node. Only
// Config fields explicitly marked as reloadable will take effect.
```

banks, line 639:
```suggestion
// ReloadConfig updates the configuration of a running raft node. If the new config would be
// invalid it will return an error without changing the running node.
```

config.go
banks:
```suggestion
	// This can be tuned during operation using ReloadConfig.
```

briankassouf (approved):
One optional comment, but looks good otherwise!

config.go
```diff
@@ -218,6 +222,39 @@ type Config struct {
...
+// ReloadableConfig is the subset of Config that may be reconfigured during
+// runtime using raft.ReloadConfig. We choose to duplicate fields over embedding
+// or accepting a Config but only using specific fields to keep the API clear.
+// Reconfiguring some fields is potentially dangerous so we should only
+// selectively enable it for fields where that is allowed.
+type ReloadableConfig struct {
+	// TrailingLogs controls how many logs we leave after a snapshot. This is used
+	// so that we can quickly replay logs on a follower instead of being forced to
+	// send an entire snapshot. The value passed here is the initial setting used.
+	// This can be tuned during operation using
```
s-christoff, line 234: I feel like this isn't finished here, but I may be misunderstanding it!

snapshot.go
```diff
@@ -219,15 +219,19 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
 
 	// Check if we have enough logs to truncate
 	lastLogIdx, _ := r.getLastLog()
-	if lastLogIdx <= r.conf.TrailingLogs {
+
+	// Ise a consistent value for trailingLogs for the duration of this method
```
s-christoff, line 223: *Use

config.go
```diff
@@ -164,19 +164,23 @@ type Config struct {
...
+	// here is the initial setting used. This can be tuned during operation using
+	// ReloadConfig.
 	SnapshotInterval time.Duration
 
 	// SnapshotThreshold controls how many outstanding logs there must be before
 	// we perform a snapshot. This is to prevent excessive snapshots when we can
-	// just replay a small set of logs.
+	// just replay a small set of logs. The value passed here is the initial
+	// setting used. This can be tuned during operation using ReloadConfig.
 	SnapshotThreshold uint64
```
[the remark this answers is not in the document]
s-christoff: Stated below, 
" We choose to duplicate fields over embedding or accepting a Config but only using specific fields to keep the API clear."
I'm not entirely sold on the idea of embedding the ReloadableConfig into the "regular" Config because the separation of "this changes, and this doesn't" is clear and "feels safer" (which is probably not a good reason, I admit)

s-christoff (approved):
Some minor nits

config.go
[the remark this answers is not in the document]
banks: I was on the fence there - I wasn't sure if it was technically binary compatible to embed which tipped the balance although in practice most reasonable usages would probably have been unaffected. I tend to agree with Sarah though - the separation ends up being kind of OK with me in the end for keeping things clearly delineated?

config.go
banks: Nope, good catch. I thought I'd fixed this but only in one of the places!

banks, line 234:
```suggestion
	// send an entire snapshot. The value passed here updates the setting at runtime
	// which will take effect as soon as the next snapshot completes and truncation
	// occurs.
```

snapshot.go
banks:
```suggestion
	// Use a consistent value for trailingLogs for the duration of this method
```

banks:
Thanks for the reviews folks! Waiting for CI to be green after typo fixes and I'll merge.

PR #450, opened 2021-03-26
testing: improving logging output for tests

dnephin:
A [flaky test failed in CI](https://app.circleci.com/pipelines/github/hashicorp/raft/298/workflows/de4daf27-568e-4725-89f4-7ec5d40f4085/jobs/5033) and I noticed the log output was difficult to
follow. I believe the reason is that `testLoggerAdapter` was double
printing the logs.

In CI tests are run using `gotestsum` which runs `go test -json`. The
`-json` flag implies `-v`, so in CI tests are always run in verbose
mode. `testLoggerAdapter` was checking `testing.Verbose` and priting
logs twice, once to stdout, and once to `t.Log`.

Instead of double printing, we can handle this logic in `newTestLogger`
helpers. If verbose mode is enabled we can use a regular logger to get
real-time output of the logs. In "quiet" mode we continue to use `t.Log`
as the output for logs so that logs are only printed on failure.

Also removed a couple duplicate helper functions. [Setting `Level` to
the default is a no-op](https://github.com/hashicorp/go-hclog/blob/master/intlogger.go#L105-L108), so many of these functions were actually doing
the same thing.

It is important to note that in general sending logs to t.Log can be a
problem because if a goroutine outlives the main test function it will
cause a panic. This does not appear to be a problem right now, so I have
not changed it.

### DIFF
--- a/testing.go
+++ b/testing.go
@@ -27,7 +27,7 @@ func inmemConfig(t *testing.T) *Config {
 	conf.ElectionTimeout = 50 * time.Millisecond
 	conf.LeaderLeaseTimeout = 50 * time.Millisecond
 	conf.CommitTimeout = 5 * time.Millisecond
-	conf.Logger = newTestLeveledLogger(t)
+	conf.Logger = newTestLogger(t)
 	return conf
 }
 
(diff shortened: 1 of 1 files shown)


s-christoff (approved):
🤙

PR #451, opened 2021-03-26
(WIP) Updating underlying bbolt

s-christoff:
Once [this](https://github.com/hashicorp/raft-boltdb/pull/22) is merged, and the repository tagged I will update this properly.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/boltdb/bolt v1.3.1 // indirect
 	github.com/hashicorp/go-hclog v0.9.1
 	github.com/hashicorp/go-msgpack v0.5.5
-	github.com/hashicorp/raft-boltdb v0.0.0-20171010151810-6e5ba93211ea
+	github.com/hashicorp/raft-boltdb v0.0.0-20210326213752-22de16c8e211
 	github.com/stretchr/testify v1.3.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -7,7 +7,6 @@ require (
 	github.com/boltdb/bolt v1.3.1 // indirect
 	github.com/hashicorp/go-hclog v0.9.1
 	github.com/hashicorp/go-msgpack v0.5.5
-	github.com/hashicorp/raft-boltdb v0.0.0-20171010151810-6e5ba93211ea
+	github.com/hashicorp/raft-boltdb v0.0.0-20210326213752-22de16c8e211
```
notnoop, line 10: It looks like `hashicorp/raft-boltdb` and `boltdb/bolt` are only used for `fuzzy` tests. The dependency is a bit troublesome too, because it's a circular dependency (raft depends, raft-boltdb which in turns depends on raft!).

I'd suggest breaking out `fuzzy` as its own go module, and have it depend on raft and raft-boltdb.  Doing so avoids requiring raft consumers from depending on boltdb unnecessarily.

s-christoff: Thank you for this suggestion and I'll get that taken care of <3

s-christoff:
Resolved in #459

PR #453, opened 2021-04-01
Allow read of reloadable config values

banks:
Continuation of #444 this adds a method that allows callers to discover the current values of config in use for use in debugging and reporting APIs as well as being useful in tests.

### DIFF
--- a/api.go
+++ b/api.go
@@ -666,7 +666,10 @@ func (r *Raft) ReloadConfig(rc ReloadableConfig) error {
 
 // ReloadableConfig returns the current state of the reloadable fields in Raft's
 // configuration. This is useful for programs to discover the current state for
-// reporting to users or tests. It is safe to call from any goroutine.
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -664,6 +664,16 @@ func (r *Raft) ReloadConfig(rc ReloadableConfig) error {
 	return nil
 }
 
+// ReloadableConfig returns the current state of the reloadable fields in Raft's
+// configuration. This is useful for programs to discover the current state for
+// reporting to users or tests. It is safe to call from any goroutine.
```
banks, line 669:
```suggestion
// reporting to users or tests. It is safe to call from any goroutine. It is intended
// for reporting and testing purposes primarily; external synchronization would be
// required to safely use this in a read-modify-write pattern for reloadable 
// configuration options.
```

briankassouf (approved):

PR #454, opened 2021-04-08
Add a gauge to hold the last restore time that

banks:
This is updated on every restore including startup from disk as well as RPC (leader/user initiated restore).

It's a gauge to provide a constant reading of the last restore time which may have been days or weeks ago in a healthy server but is useful to be able to graph along side the trailing logs age to check if there is danger of restores taking longer than trailing logs can cover.

It differs from fsm.restore both because it is omitted at startup and because it is a gauge rather than a single point in time sample.

This is a follow up to #452 (and relates to https://github.com/hashicorp/consul/issues/9609) after realising the current restore metrics are not really suitable to easily monitor whether the last restore too dangerously long or not!

Having this gauge means operators only need to compare two gauge values over time to get a good indicator of if their cluster is in danger of becoming unrecoverable.

### DIFF
--- a/api.go
+++ b/api.go
@@ -605,9 +605,11 @@ func (r *Raft) restoreSnapshot() error {
 			}
 
 			if err := fsmRestoreAndMeasure(r.fsm, source); err != nil {
+				source.Close()
 				r.logger.Error("failed to restore snapshot", "id", snapshot.ID, "error", err)
 				continue
 			}
+			source.Close()
 
 			r.logger.Info("restored from snapshot", "id", snapshot.ID)
 		}
--- a/api.go
+++ b/api.go
@@ -323,7 +323,12 @@ func RecoverCluster(conf *Config, fsm FSM, logs LogStore, stable StableStore,
 			continue
 		}
 
-		start := time.Now()
+		// Note this is the one place we call fsm.Restore without the
+		// fsmRestoreAndMeasure wrapper since this function should only be called to
+		// reset state on disk and the FSM passed will not be used for a running
+		// server instance. If the same process will eventually become a Raft peer
+		// then it will call NewRaft and restore again from disk then which will
+		// report metrics.
 		err = fsm.Restore(source)
 		// Close the source after the restore has completed
 		source.Close()
@@ -331,9 +336,6 @@ func RecoverCluster(conf *Config, fsm FSM, logs LogStore, stable StableStore,
 			// Same here, skip and try the next one.
 			continue
 		}
-		// Update the gauge for the time it took to restore.
-		metrics.SetGauge([]string{"raft", "fsm", "lastRestoreTime"},
-			float32(time.Since(start).Milliseconds()))
 
 		snapshotIndex = snapshot.Index
 		snapshotTerm = snapshot.Term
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -583,6 +587,11 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 // of them can be restored. This is called at initialization time, and is
 // completely unsafe to call at any other time.
 func (r *Raft) restoreSnapshot() error {
+	// Measure time from here since latency on listing snapshots or having to try
+	// multiple snapshots all adds to the restore time that matters i.e. how long
+	// the server takes to restart.
+	start := time.Now()
```
dnephin, line 593: I see what you are saying here, but I wonder how that will work out in practice. It seems strange that the first time we report this gauge it is going to include a bunch of other time that won't be reported by the other two sites which report this metric.

I imagine there is plenty of other "startup" time that is not reflected in this metric as well. Time required for the process to start, read config, and get to this point.

I wonder if it would be better to report this metric the same as the others (only around the fsm.Restore), and document that this is a minimum time and that a sufficient buffer should be added.

Or, maybe use `raft.fsm.restore` in all 3 call sites to indicate the time just for restore, and only report the gauge from this one call site? If the looping over other snapshots is going to be a significant amount of time, then reporting the gauge from other places and not including that time is going to report a much lower number and could be very misleading.

api.go
```diff
@@ -323,13 +323,17 @@ func RecoverCluster(conf *Config, fsm FSM, logs LogStore, stable StableStore,
...
 		err = fsm.Restore(source)
 		// Close the source after the restore has completed
 		source.Close()
 		if err != nil {
 			// Same here, skip and try the next one.
 			continue
 		}
+		// Update the gauge for the time it took to restore.
+		metrics.SetGauge([]string{"raft", "fsm", "lastRestoreTime"},
+			float32(time.Since(start).Milliseconds()))
```
dnephin, line 336: A couple thoughts here.

1. The `Time` suffix in the name here could imply that this is a timer metric, since we use `Time` in other timers. Maybe `lastRestoreElapsed` or something else would help identify this as a gauge?
2. Generally metrics will only be reported from a single place. If we're going to report this metric in a few places I think it would be good to put it into a function and call that function from all the locations. The repetition isn't really a problem, but I do think that a function acts as a good hint to the reader that this metric breaks the common pattern of only being reported from a single place.

dnephin (commented):
Nice! Some thoughts/questions on specifics, but generally I think adding a metric like this is a really good idea.

api.go
banks: That's fair. I reasoned that since this is still within the single logical operation of `restoreSnapshot` method it made some sense. 

Honestly I'm not super sure why we loop here - if the most recent completed snapshot on disk isn't readable for some reason then it's almost certain that we either don't have an older one, or if we do that we don't still have enough logs to restore from it since we would have truncated our logs after taking the newer one.

But yeah since it's an edge case I'm not sure is even possible it's not a big deal to move it back inside the loop. That will allow using a function/emitting in one place anyway as you suggested.

api.go
banks: I went with `Duration` as a suffix and moved to a central func. Good call.

banks:
Thanks @dnephin I think this is a cleaner diff now!

dnephin (approved):
Nice, LGTM!

I was thinking of only moving the one metric to a function, but moving both metrics and the restore call works quite nicely as well.

PR #455, opened 2021-04-08
Document how FSM.Apply errors are retrieved from a Future

dnephin:
We noticed that the `ApplyFuture.Response()` return value was not being handled consistently in many places, which may lead to missed errors. This PR updates the docs to call attention to the problem.

### DIFF
--- a/future.go
+++ b/future.go
@@ -9,12 +9,13 @@ import (
 
 // Future is used to represent an action that may occur in the future.
 type Future interface {
-	// Error blocks until the future arrives and then
-	// returns the error status of the future.
-	// This may be called any number of times - all
-	// calls will return the same value.
-	// Note that it is not OK to call this method
-	// twice concurrently on the same Future instance.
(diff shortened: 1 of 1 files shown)


future.go
```diff
@@ -9,12 +9,14 @@ import (
...
-	// This may be called any number of times - all
-	// calls will return the same value.
-	// Note that it is not OK to call this method
-	// twice concurrently on the same Future instance.
+	// Error blocks until the future arrives and then returns the error status
+	// of the future. This may be called any number of times - all calls will
+	// return the same value.
+	// Note that it is not OK to call this method twice concurrently on the same
+	// Future instance.
+	// Note that Error() will only return generic errors related to raft, such
```
s-christoff, line 17: Minor itsy bitsy spider sized nit: You can just remove the Note (or both of them) and just keep that as part of the informational comment.

s-christoff (approved):
These seem like solid changes. Thanks for doing this :) 🤙

PR #458, opened 2021-04-12
Fix flaky TestRaft_LeadershipTransferLeaderRejectsClientRequests

dnephin:
This test is currently the test which flakes the most often.

The test would fail because of the select read-from-channel in the test. This
read-from-channel could race with the `leaderLoop` goroutine. If the
read-from-channel in the test won then the `leaderLoop` goroutine would
never receive the future, and the future would never respond, leaving the
test to timeout eventually.

The fix was to remove the select and read-from-channel from the test. This fix
caused the test to fail, because of the following call to `LeadershipTransferToServer`
would not being able to send to the channel immediately, resulting in an
ErrEnqueueTimeout instead of ErrLeadershipTransferInProgress.

To fix the secondary problem the test was changed to call
`LeadershipTransferToServer` after the channel is drained (by waiting on
all the futures).

Also added a `t.Logf` to the test, since that helped debug the problem I
thought it would be good to keep it around in case we encounter other
problems with this test.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -347,10 +347,7 @@ func (r *Raft) setLeadershipTransferInProgress(v bool) {
 
 func (r *Raft) getLeadershipTransferInProgress() bool {
 	v := atomic.LoadInt32(&r.leaderState.leadershipTransferInProgress)
-	if v == 1 {
-		return true
-	}
-	return false
+	return v == 1
 }
 
 func (r *Raft) setupLeaderState() {
--- a/raft_test.go
+++ b/raft_test.go
@@ -2059,17 +2059,17 @@ func TestRaft_LeadershipTransferLeaderRejectsClientRequests(t *testing.T) {
 		l.requestConfigChange(configurationChangeRequest{}, 100*time.Millisecond),
 	}
 	futures = append(futures, l.LeadershipTransfer())
-	select {
-	case <-l.leadershipTransferCh:
-	default:
-	}
-
-	futures = append(futures, l.LeadershipTransferToServer(ServerID(""), ServerAddress("")))
 
 	for i, f := range futures {
+		t.Logf("waiting on future %v", i)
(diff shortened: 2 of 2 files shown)


raft.go
```diff
@@ -347,10 +347,7 @@ func (r *Raft) setLeadershipTransferInProgress(v bool) {
 
 func (r *Raft) getLeadershipTransferInProgress() bool {
 	v := atomic.LoadInt32(&r.leaderState.leadershipTransferInProgress)
-	if v == 1 {
-		return true
-	}
-	return false
+	return v == 1
 }
```
dnephin, line 351: This is not related to the fix, just a small cleanup.

raft.go
```diff
@@ -347,10 +347,7 @@ func (r *Raft) setLeadershipTransferInProgress(v bool) {
 
 func (r *Raft) getLeadershipTransferInProgress() bool {
 	v := atomic.LoadInt32(&r.leaderState.leadershipTransferInProgress)
-	if v == 1 {
-		return true
-	}
-	return false
+	return v == 1
```
s-christoff, line 350: I did not know you could do this in Go - very cool :)

raft_test.go
```diff
@@ -2059,18 +2059,18 @@ func TestRaft_LeadershipTransferLeaderRejectsClientRequests(t *testing.T) {
 		l.requestConfigChange(configurationChangeRequest{}, 100*time.Millisecond),
 	}
 	futures = append(futures, l.LeadershipTransfer())
-	select {
-	case <-l.leadershipTransferCh:
-	default:
-	}
```
dnephin, line 2065: This was the core of the problem. This "read from channel" would race with the one in `leaderLoop`, and if this one won the race the test would block until the timeout.

s-christoff (approved):
🤙
This looks great.

PR #459, opened 2021-04-14
extract fuzzy as its own module

notnoop:
Fuzzy depends on raft-boltdb, which depends on raft as well. Having
fuzzy as its own module, breaks the circular dependency.

Also, having `fuzzy` avoids leaking the raft-boltdb dependency and
versioning constaints into consumers. This was a significant factor, as
`raft-boltdb` changed storage engine in
https://github.com/hashicorp/raft-boltdb/pull/22. However, given
raft-boltdb storage occured as a api major change, and both raft-boltdb
storage engines can coexist in the same go project, so it's less of an
issue now.


s-christoff (approved):
🤙 Thank you for doing this!

PR #430, opened 2020-09-15
[FIXED] LogCache should not cache on StoreLogs error

kozlovic:
LogCache was caching logs and then invoking store.StoreLogs().
However, if StoreLogs() fails to store, LogCache.Get() could still
return logs that had not been persisted into storage, which could
lead to a node failing to restart with a "log not found" panic.

This PR ensures that LogCache only cache logs if StoreLogs was
successful.

Resolves #429

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>

PR #460, opened 2021-04-14
[Updated] LogCache should not cache on StoreLogs error

### LINKED ISSUE #429: Restarting RAFT node could fail with `panic: log not found [recovered]` when using LogCache
Would get such stack:
```
panic: log not found [recovered]
	panic: log not found

goroutine 18 [running]:
testing.tRunner.func1.1(0x171e8e0, 0xc00018e190)
	/usr/local/go/src/testing/testing.go:988 +0x452
testing.tRunner.func1(0xc0001ca120)
	/usr/local/go/src/testing/testing.go:991 +0x600
panic(0x171e8e0, 0xc00018e190)
	/usr/local/go/src/runtime/panic.go:975 +0x3e3
github.com/hashicorp/raft.NewRaft(0xc00027a060, 0x188d7e0, 0xc0001c1080, 0x1892180, 0xc0001c0640, 0x188fa20, 0xc0001c0640, 0x188d7a0, 0xc0001824b0, 0x1895a20, ...)
	/Users/ivan/dev/go/src/github.com/hashicorp/raft/api.go:545 +0x2014
```
LogCache incorrectly caches logs prior to invoking store.StoreLogs(), which means that it would possibly cache logs that have not made it to the backend storage.
This could cause RAFT library to query for a log and get it (due to the improper cache) and keep track of a log index that is assumed to be committed while it is not. When restarting, the node would then try to look for log X and fail to find it in the backend store.

s-christoff:
Original author was #430 
[FIXED] LogCache should not cache on StoreLogs error
LogCache was caching logs and then invoking store.StoreLogs().
However, if StoreLogs() fails to store, LogCache.Get() could still
return logs that had not been persisted into storage, which could
lead to a node failing to restart with a "log not found" panic.

This PR ensures that LogCache only cache logs if StoreLogs was
successful.

Resolves #429

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>

### DIFF
--- a/log_cache_test.go
+++ b/log_cache_test.go
@@ -1,6 +1,8 @@
 package raft
 
 import (
+	"errors"
+	"sync"
 	"testing"
 )
 
@@ -84,5 +86,58 @@ func TestLogCache(t *testing.T) {
 	err = c.GetLog(34, &out)
 	if err != ErrLogNotFound {
 		t.Fatalf("err: %v", err)
+	}
+}
+
+type errorStore struct {
+	LogStore
+	mu      sync.Mutex
+	fail    bool
+	failed  int
+	failMax int
+}
+
+func (e *errorStore) StoreLogs(logs []*Log) error {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	if e.fail {
+		e.failed++
+		if e.failed <= e.failMax {
+			return errors.New("some error")
+		}
+		e.fail = false
+	}
+	return e.LogStore.StoreLogs(logs)
+}
+
+func (e *errorStore) failNext(count int) {
+	e.mu.Lock()
+	e.fail = true
+	e.failMax = count
+	e.mu.Unlock()
+}
+
+func TestLogCacheWithBackendStoreError(t *testing.T) {
+	store := NewInmemStore()
+	errStore := &errorStore{LogStore: store}
+	c, _ := NewLogCache(16, errStore)
+
+	for i := 0; i < 4; i++ {
+		log := &Log{Index: uint64(i) + 1}
+		store.StoreLog(log)
+	}
+	errStore.failNext(1)
+	log := &Log{Index: 5}
+	c.StoreLog(log)
+
+	var out Log
+	for i := 0; i < 4; i++ {
+		if err := c.GetLog(uint64(i+1), &out); err != nil {
+			t.Fatalf("err: %v", err)
+		}
+	}
+	out = Log{}
+	if err := c.GetLog(5, &out); err != ErrLogNotFound {
+		t.Fatalf("Should have returned not found, got err=%v out=%+v", err, out)
 	}
 }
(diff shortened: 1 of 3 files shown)


log_cache_test.go
```diff
@@ -86,3 +88,56 @@ func TestLogCache(t *testing.T) {
...
+		log := &Log{Index: uint64(i) + 1}
+		store.StoreLog(log)
+	}
+	errStore.failNext(1)
+	log := &Log{Index: 5}
+	c.StoreLog(log)
+
+	var out Log
+	for i := 0; i < 4; i++ {
+		if err := c.GetLog(uint64(i+1), &out); err != nil {
```
notnoop, line 135: I got a bit confused following the indexes and seeing the `i+1` here.  How does looping from 1 to 5 instead to avoid the index arithmetic sound?

log_cache_test.go
```diff
@@ -86,3 +88,56 @@ func TestLogCache(t *testing.T) {
...
+			t.Fatalf("err: %v", err)
+		}
+	}
+	out = Log{}
+	if err := c.GetLog(5, &out); err != ErrLogNotFound {
+		t.Fatalf("Should have returned not found, got err=%v out=%+v", err, out)
```
notnoop, line 141: Very informative error message :).

raft_test.go
```diff
@@ -2176,6 +2176,55 @@ func TestRaft_GetConfigurationNoBootstrap(t *testing.T) {
...
+	// Try to restart the follower and make sure it does not fail with a LogNotFound error
+	_, trans := NewInmemTransport(follower.localAddr)
+	follower.logs = es.LogStore
+	conf := follower.config()
+	n, err := NewRaft(&conf, &MockFSM{}, follower.logs, follower.stable, follower.snapshots, trans)
+	if err != nil {
+		c.FailNowf("error restarting follower: %v", err)
+	}
+	n.Shutdown()
+}
```
notnoop, line 2227: white space nitpicks:

```suggestion
}

```

notnoop (approved):
LGTM.  Minor nitpicks. Like the test to code change ratio ;-).

log_cache_test.go
```diff
@@ -86,3 +89,60 @@ func TestLogCache(t *testing.T) {
...
+	for i := 0; i < 4; i++ {
+		log := &Log{Index: uint64(i) + 1}
+		store.StoreLog(log)
+	}
+	errStore.failNext(1)
+	log := &Log{Index: 5}
+	err = c.StoreLog(log)
+	if !strings.Contains(err.Error(), "unable to store logs in log store") {
+		t.Fatalf("Should have returned unable to store logs in log store,  got err=%v", err)
+	}
```
notnoop, line 136: may need to update this after removing the wrapping: it's failing in CI: https://app.circleci.com/pipelines/github/hashicorp/raft/343/workflows/82684bfa-24ff-4bd7-91f2-17afb2efccc1/jobs/5360

PR #461, opened 2021-04-16
make TestRaft_RecoverCluster less flaky

dnephin:
@s-christoff @jsosulska and I paired on this.

The test seems to fail at the end when it's waiting on a leader. 

Increasing the number of peers seemed to make the test fail more often.
Increase the sleep seemed to make it fail much less.

We suspect that we could improve `EnsureLeader` to remove the need for a sleep by polling for a stable leader, but deferred that change for now since it seems this small change is working.

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -204,14 +204,14 @@ func TestRaft_RecoverCluster(t *testing.T) {
 			c.fsms[i] = r2.fsm.(*MockFSM)
 		}
 		c.FullyConnect()
-		time.Sleep(c.propagateTimeout)
+		time.Sleep(c.propagateTimeout * 3)
 
 		// Let things settle and make sure we recovered.
 		c.EnsureLeader(t, c.Leader().localAddr)
 		c.EnsureSame(t)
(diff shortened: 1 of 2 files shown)


raft_test.go
```diff
@@ -204,14 +204,14 @@ func TestRaft_RecoverCluster(t *testing.T) {
...
-		time.Sleep(c.propagateTimeout)
+		time.Sleep(c.propagateTimeout * 3)
 
 		// Let things settle and make sure we recovered.
 		c.EnsureLeader(t, c.Leader().localAddr)
 		c.EnsureSame(t)
 		c.EnsureSamePeers(t)
 	}
-	for applies := 0; applies < 20; applies++ {
+	for applies := 0; applies < 10; applies++ {
```
dnephin, line 214: We thought it would make sense to reduce this number because effectively this is just repeating the same test many times. We really only need to run this test twice, once with 0 applies , and once with n>1 applies, so 10 should be more than sufficient.

s-christoff (approved):

PR #462, opened 2021-04-17
Fix TestRaft_VerifyLeader

s-christoff:
It looks like we are waiting for the leader to step down and verify there is no leader. 
This may be a great case to use the `LeaderCh` !
Opinions welcome :)

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -1622,8 +1622,9 @@ func TestRaft_VerifyLeader_Fail(t *testing.T) {
 	c := MakeCluster(2, t, conf)
 	defer c.Close()
 
-	// Get the leader
 	leader := c.Leader()
+	// Remove the leader election notification from the channel buffer
+	<-leader.LeaderCh()
 
 	// Wait until we have a followers
 	followers := c.Followers()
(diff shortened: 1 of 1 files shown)


dnephin (commented):
Nice!  I think this change makes sense, but I'm not sure it fixes the flake.

I tried running this locally
```
go test -run TestRaft_VerifyLeader_Fail -count=500 -race -failfast .
```

I was able to get it to fail with `main`, but I was also able to get it to fail with the changes in this PR at the same place (`verify.Error()` returning nil`).

When I made the change I suggested below to `t.Fatalf` when v is true, it failed with that new `Fatalf` message. I think this might mean that the flake is being caused when `VerifyLeader` before the cluster is stable.

One thought is maybe we need another one of these `leaderCh` selects before calling `VerifyLeader` ?

s-christoff:
```
go test -run TestRaft_VerifyLeader_Fail -count=500 -race -failfast .

ok      github.com/hashicorp/raft       143.628s
```
:)

dnephin (approved):
LGTM!

PR #463, opened 2021-04-20
testing: remove cluster.FailNowf

dnephin:
This function was nothing but a wrapper around t.Fatalf. Wrapping
t.Fatalf is problematic for a number of reasons:

1. It makes tests less obvious. A reader can easily reason about what
   t.Fatalf is doing, but has to look up what c.FailNowf is supposed to
   be. Since it doesn't do anything other than call t.Fatalf this is an
   unnecessary distraction.
2. Fatalf may only be called from the main goroutine. It is easy to call
   c.FailNowf from some other goroutine without realizing that it is
   incorrect, and there was at least one case of this recently in the
   test suite.
3. The old message that was being printed by FailNowf was going through
   the logger, making it a lot more difficult to find the error that
   caused the test to fail.
4. It is really easy to misuse FailNowf because it uses the testing.T
   embedded in cluster. If this function is called from any subtest, it
   will report the message to the wrong test (the root test instead of
   the sub test).

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -28,31 +28,31 @@ func TestRaft_AfterShutdown(t *testing.T) {
 
 	// Everything should fail now
 	if f := raft.Apply(nil, 0); f.Error() != ErrRaftShutdown {
-		c.FailNowf("should be shutdown: %v", f.Error())
+		t.Fatalf("should be shutdown: %v", f.Error())
 	}
 
 	// TODO (slackpad) - Barrier, VerifyLeader, and GetConfiguration can get
 	// stuck if the buffered channel consumes the future but things are shut
 	// down so they never get processed.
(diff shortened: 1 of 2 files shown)


s-christoff (approved):
This looks good, my only question being is that if this (`FailNowf`) is removed from tests without a `t.Helper()` then are we removing some additional testing data, and if so would it be handy to add that call in the tests themselves in this change?

dnephin:
I don't believe we are losing anything. `t.Helper` ([docs](https://golang.org/pkg/testing/#T.Helper)) is for marking a function as a helper. It's sole purpose is to skip some stack frames when looking for the file and line number to use when reporting a test failure or a `t.Log` message.

By removing the helper function we remove the need to call `t.Helper`.

PR #465, opened 2021-04-30
Handle installSnapshot decodePeers error without a panic

dnephin:
`decodePeers` can be called by the `installSnapshot` RPC handler, which means that a panic can be caused by an external actor by sending malformed input. Instead handle the error and report it back via the RPC response. This prevents a panic and removes the possibility of an external actor being able to cause a raft peer to panic.

The new test would cause a panic before, and now it confirms the expected error is returned.

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -13,6 +13,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hashicorp/go-hclog"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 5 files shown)


raft_test.go
```diff
@@ -2316,3 +2317,24 @@ func TestRaft_ReloadConfigValidates(t *testing.T) {
 //
 // Storage errors handled properly.
 // Commit index updated properly.
+
+func TestRaft_InstallSnapshot(t *testing.T) {
```
banks, line 2321: super nit: this test is specifically only testing the behaviour for invalidly formed peers in a request. Is is worth giving it a more specific name?

banks (approved):
LGTM! Thanks for fixing this up.

PR #466, opened 2021-05-14
ci: update config

dnephin:
Best viewed by individual commit. I can break them up if any of them seem like they need more individual attention.

- test against fewer go versions, only the last 2 major versions of Go are supported so we can run tests against only those two
- only test against a single version with 32bit
- remove the module cache. With the new module proxy it is faster to download from the proxy.
- also fixes new errors reported by `go vet` when run using Go1.16 (previous we used go1.14)

Also refactor the config a bit to:

- remove duplication in the go-test jobs (we can specify GOARCH as a parameter)
- use the same executor for both jobs, so that the docker registry url can be specified in a single place
- remove the need for any yaml anchors and aliases
- run all the jobs at once instead of waiting for the lint job to finish

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -120,13 +120,12 @@ func TestRaft_RecoverCluster_NoState(t *testing.T) {
 }
 
 func TestRaft_RecoverCluster(t *testing.T) {
-	// Run with different number of applies which will cover no snapshot and
-	// snapshot + log scenarios. By sweeping through the trailing logs value
-	// we will also hit the case where we have a snapshot only.
-	var err error
+	snapshotThreshold := 5
 	runRecover := func(t *testing.T, applies int) {
+		var err error
 		conf := inmemConfig(t)
 		conf.TrailingLogs = 10
+		conf.SnapshotThreshold = uint64(snapshotThreshold)
 		c := MakeCluster(3, t, conf)
 		defer c.Close()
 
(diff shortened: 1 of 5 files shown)


raft_test.go
```diff
@@ -212,11 +211,16 @@ func TestRaft_RecoverCluster(t *testing.T) {
...
+
+	t.Run("no snapshot, no trailing logs", func(t *testing.T) {
+		runRecover(t, 0)
+	})
+	t.Run("snapshot", func(t *testing.T) {
+		runRecover(t, snapshotThreshold)
+	})
+	t.Run("snapshot, with trailing logs", func(t *testing.T) {
+		runRecover(t, snapshotThreshold+20)
+	})
```
dnephin, line 223: The old comment said:

>  By sweeping through the trailing logs value we will also hit the case where we have a snapshot only.

But I'm not sure what that means. We never set the snapshot threshold previously. Is it the `TrailingLogs` value that is relevant here? I don't see how. It's only used in the code to prune.

i0rek (approved):
lgtm

raft_test.go
dnephin: I've made one change to this, so that the second case is "some logs with no snapshot", which I think is a more correct description of that case. Also changed the applies value to `snapshotThreshold-1` to make sure we don't have a snapshot yet.

PR #471, opened 2021-08-23
Fix peer access race with raft heartbeat.

SimonRichardson:
This race recently came up when the Juju project moved to the upstream
raft project. The problem was manifested during testing
(unit/integration test) setup. The race happens during a Raft
replicate, a goroutine is spawned to manage the heartbeats against a
peer. The problem occurs when a peer can be changed during a
startStopReplication call. If the server.ID already exists, yet the
server address isn't the same, then the underlying peer is updated and
a data race occurs.

The fix is to wrap all peer access with a RWMutex. As the pattern in the
library uses a lot more reads than writes, it makes much more sense to
use this type of mutex.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -507,9 +507,18 @@ func (r *Raft) startStopReplication() {
 			r.goFunc(func() { r.replicate(s) })
 			asyncNotifyCh(s.triggerCh)
 			r.observe(PeerObservation{Peer: server, Removed: false})
-		} else if ok && s.peer.Address != server.Address {
-			r.logger.Info("updating peer", "peer", server.ID)
-			s.peer = server
+		} else if ok {
+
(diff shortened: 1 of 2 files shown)


raft.go
```diff
@@ -507,9 +507,18 @@ func (r *Raft) startStopReplication() {
...
+		} else if ok {
+
+			s.peerLock.RLock()
+			peer := s.peer
+			s.peerLock.RUnlock()
+
+			if peer.Address != server.Address {
+				r.logger.Info("updating peer", "peer", server.ID)
+				s.peerLock.Lock()
+				s.peer = server
```
SimonRichardson, line 519: This is the underlying issue. 

Setting the server to the underlying peer whilst doing a heartbeat in another goroutine causes a data race.

PR #476, opened 2021-09-22
Forbid removed node from being a `Candidate`

dhiaayachi:
When a follower is removed from the cluster it's removed from the configuration servers list but still run a follower loop (originally a follower node or `ShutdownOnRemove` set to `false` for Leader). When that node timeout (because it's not receiving heartbeat from the other nodes in the cluster), it transition to a Candidate state and try to trigger an election. 2 possible cases:
- the cluster is stable and have a leader: the vote request will be rejected
- the cluster is unstable and don't have a leader: the vote is granted and the removed node will be part of the voting nodes.

This uncover 2 issue:
- a removed node should not transition to a `Candidate state`. This is fixed in this PR
- a node that is not part of the cluster should not be allowed to vote: this a bit trickier to fix because the `RequestVoteRequest` only have the `ServerAddress` which do not represent a unique ID of the node, the one stored in the configuration server list. A test that reproduce the issue is in #477


mkeeler (approved):
LGTM with one minor comment about a warning log.

PR #477, opened 2021-09-23
node not part of the cluster is not allowed to vote

dhiaayachi:
When a follower is removed from the cluster it's removed from the configuration servers list but still run a follower loop (originally a follower node or `ShutdownOnRemove` set to `false` for Leader). When that node timeout (because it's not receiving heartbeat from the other nodes in the cluster), it transition to a Candidate state and try to trigger an election. 2 possible cases:
- the cluster is stable and have a leader: the vote request will be rejected
- the cluster is unstable and don't have a leader: the vote is granted and the removed node will be part of the voting nodes.

This uncover 2 issue:
- a removed node should not transition to a `Candidate state`. This is fixed in #476 
- a node that is not part of the cluster should not be allowed to vote: this a bit trickier to fix because the `RequestVoteRequest` only have the `ServerAddress` which do not represent a unique ID of the node, the one stored in the configuration server list. A test that reproduce the issue is in this PR.

To keep track of things in this PR here is a list of what I intend to achieve:

- [x] Add `ID` and `Addr` field in RPCHeader
- [x] Transition from using `Candidate` and `Leader` fields in Protocol Version 3 to using `Addr` (part of RPC header) field in protocol version 4
- [x] Use `ID` to validate that the node that is triggering an election is part of the latest configuration
- [x] Expose the Leader `ID` field as part of the API to be used in consul
- [x] Mixed `ProtocolVersion` Test


dhiaayachi:
@mkeeler I was thinking about this after our conversation today and I noticed few things:
- Adding the ID as part of `EncodePeer` and `DecodePeer` would break compatibility because in the case of node that is version 4 talking to a node version 3, the node version 3 would not be able to decode the receiver peer correctly and fail, So we will need to make a 2 step upgrade process as discussed
- I was thinking that we can add the ID as a separate field with a separate set of encoders decoders like it's currently the case in this PR, this will make the upgrade seamless and we still can survive with nodes running different versions (3 and 4)
- I also thought that having the ID of the source node as part of any raft request/response is very useful (at least for 2 know cases that we have today) and instead of adding it as part of `RequestVoteRequest` only I can add it in `RPCHeader` so all the request/response benefit from it. We can encode it all the time and only use it with a protocolVersion check in all the needed cases (`RequestVoteRequest` , `AppendEntriesRequest` and `InstallSnapshotRequest`)

Let me know what you think about this?

mkeeler:
So from our original discussions I was thinking the upgrade process would look like:

1. All nodes are on some existing version at protocol 3. DecodePeer & EncodePeer only handle the raw server address.
2. New nodes are added with the new version but set to use protocol 3. EncodePeer sends the server address form. DecodePeer can handle either the old form or some new struct that contains the id and address.
3. Old nodes are removed.
4. New nodes get a rolling restart and get switched to protocol 4. EncodePeer will now send both the id and address as the peer info instead of just the address. DecodePeer can still handle both forms though.

So with that sort of flow it wouldn't break compatibility. However I think I prefer your approach of adding the id to the `RPCHeader` as we are likely to want that information in other places in order to track the leaders id and make it available to the application running Raft.

I wonder whether we should add both the address and id of the peer server to the `RPCHeader` and have Raft proto 4 not encode the `Candidate` field of the `RequestVoteRequest` and the `Leader` field of the `AppendEntriesRequest` / `InstallSnapshotRequest`. What do you think?

dhiaayachi:
Thank you for taking a look at this @mkeeler  

About the upgrade process, I was thinking the same. I just called it a breaking change from the perspective that it need a special upgrade flow (2 restarts).

About removing the `Candidate` and `Leader` fields I need to check if that bring any other implications, but I think the use of it is internal to raft and it's doable, I will update the PR and get back to you for a round of review.

dhiaayachi:
To keep track of things in this PR here is a list of what I intend to achieve:

- [ ] Add `ID` and `Addr` field in RPCHeader
- [ ] Transition from using `Candidate` and `Leader` fields in Protocol Version 3 to using `Addr` (part of RPC header) field in protocol version 4
- [ ] Use `ID` to validate that the node that is triggering an election is part of the latest configuration
- [ ] Expose the Leader `ID` field as part of the API to be used in consul

raft.go
```diff
@@ -1711,11 +1744,13 @@ func (r *Raft) electSelf() <-chan *voteResult {
 	req := &RequestVoteRequest{
 		RPCHeader:          r.getRPCHeader(),
 		Term:               r.getCurrentTerm(),
-		Candidate:          r.trans.EncodePeer(r.localID, r.localAddr),
 		LastLogIndex:       lastIdx,
 		LastLogTerm:        lastTerm,
 		LeadershipTransfer: r.candidateFromLeadershipTransfer,
 	}
+	if req.ProtocolVersion <= 3 {
+		req.Candidate = r.trans.EncodePeer(r.localID, r.localAddr)
```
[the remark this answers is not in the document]
dhiaayachi: I added tests that instantiate protocol 3 nodes with protocol v4 nodes in the same cluster, but this is still use the branch version of the code, so I can't really cover a real upgrade scenario.

raft_test.go
```diff
@@ -1881,6 +1929,108 @@ func TestRaft_ProtocolVersion_Upgrade_2_3(t *testing.T) {
...
+}
+
+func TestRaft_MixedCluster_Stable_v3Leader(t *testing.T) {
+	// Make a cluster back on protocol version 1.
+	conf := inmemConfig(t)
+	conf.ProtocolVersion = 3
+	c := MakeCluster(2, t, conf)
+	defer c.Close()
+
+	// Set up another server speaking protocol version 2.
```
mkeeler, line 1968:
```suggestion
	// Set up another server speaking protocol version 4.
```

raft_test.go
```diff
@@ -1881,6 +1929,108 @@ func TestRaft_ProtocolVersion_Upgrade_2_3(t *testing.T) {
...
+	time.Sleep(c.propagateTimeout)
+
+	// Sanity check the cluster.
+	c.EnsureSame(t)
+	c.EnsureSamePeers(t)
+	c.EnsureLeader(t, c.Leader().localAddr)
+}
+
+func TestRaft_MixedCluster_Stable_v4Leader(t *testing.T) {
+	// Make a cluster back on protocol version 1.
```
mkeeler, line 1999:
```suggestion
	// Make a cluster on protocol version 4.
```

raft_test.go
```diff
@@ -1881,6 +1929,108 @@ func TestRaft_ProtocolVersion_Upgrade_2_3(t *testing.T) {
...
+	conf := inmemConfig(t)
+	conf.ProtocolVersion = 4
+	c := MakeCluster(2, t, conf)
+	defer c.Close()
+
+	// Set up another server speaking protocol version 2.
```
mkeeler, line 2005:
```suggestion
	// Set up another server speaking protocol version 3.
```

raft_test.go
```diff
@@ -1881,6 +1929,108 @@ func TestRaft_ProtocolVersion_Upgrade_2_3(t *testing.T) {
 	}
 }
 
+func TestRaft_LeaderID_Propagated(t *testing.T) {
+	// Make a cluster back on protocol version 2.
```
mkeeler, line 1933:
```suggestion
	// Make a cluster on protocol version 3.
```

mkeeler (approved):
All the main code looks good. There are a handful of comment fixups and two logging fixes suggested but nothing that would require further review of.

api.go
```diff
@@ -716,11 +718,12 @@ func (r *Raft) BootstrapCluster(configuration Configuration) Future {
 // Leader is used to return the current leader of the cluster.
 // It may return empty string if there is no current leader
 // or the leader is unknown.
-func (r *Raft) Leader() ServerAddress {
+func (r *Raft) Leader() (ServerAddress, ServerID) {
```
dnephin, line 721: I believe this would be considered a breaking change of the library interface. I imagine we don't want to release a v2 for this, so I suggest that we keep the old method, add a `Deprecated: ...` line as the last line of the godoc, and add a new `LeaderWithID()` method which returns the extra data.

The deprecated `Leader` method can call `LeaderWithID` and return only the one value.

commands.go
```diff
@@ -8,6 +8,10 @@ type RPCHeader struct {
 	// ProtocolVersion is the version of the protocol the sender is
 	// speaking.
 	ProtocolVersion ProtocolVersion
+	// ID is the ServerID of the node sending the RPC Request or Response
+	ID []byte
+	// Addr is the ServerAddr of the node sending the RPC Request or Response
+	Addr []byte
```
dnephin, line 14: Nice godoc! This helps a lot

fuzzy/verifier.go
```diff
@@ -45,7 +45,13 @@ func (v *appendEntriesVerifier) PreRequestVote(src, target string, rv *raft.Requ
 
 func (v *appendEntriesVerifier) PreAppendEntries(src, target string, req *raft.AppendEntriesRequest) (*raft.AppendEntriesResponse, error) {
 	term := req.Term
-	ldr := string(req.Leader)
+	var ldr string
+	if req.ProtocolVersion > 3 {
+		ldr = string(req.Addr)
```
dnephin, line 50: Since this field is in an embedded structs we can access it this way, but generally I always try to be more explicit. I think including the embedded struct in the field access helps make it clear where the value is coming from:
```suggestion
		ldr = string(req.RPCHeader.Addr)
```

There are a few more places this is accessed, but I'll leave it to you if you think this is better or not in general.

net_transport.go
```diff
@@ -568,10 +568,18 @@ func (n *NetworkTransport) handleCommand(r *bufio.Reader, dec *codec.Decoder, en
...
-			len(req.Entries) == 0 && req.LeaderCommitIndex == 0 {
-			isHeartbeat = true
+		if req.ProtocolVersion > 3 {
+			if req.Term != 0 && req.Addr != nil &&
+				req.PrevLogEntry == 0 && req.PrevLogTerm == 0 &&
+				len(req.Entries) == 0 && req.LeaderCommitIndex == 0 {
+				isHeartbeat = true
+			}
+		} else {
+			if req.Term != 0 && req.Leader != nil &&
```
dnephin, line 578: Do we still populate the `req.Leader` field? I'm wondering if it would be safe to continue to trust the `req.Leader` field in these cases, instead of having a lot of branches that check the protocol version.

There are a few places where we have this check, but it seems like we could continue to use `req.Leader` for most of these. Is there a reason not to do that?

Either way I think we should add to the `Leader` field godoc, to indicate it is ignored in V4+, or that we continue to use it instead of the `RPCHeader.Addr` (if that is an acceptable option).

raft.go
```diff
@@ -31,6 +31,8 @@ var (
 func (r *Raft) getRPCHeader() RPCHeader {
 	return RPCHeader{
 		ProtocolVersion: r.config().ProtocolVersion,
+		ID:              []byte(r.config().LocalID),
+		Addr:            r.trans.EncodePeer(r.config().LocalID, r.localAddr),
```
dnephin, line 35: I'm pretty sure adding these new fields to the `RPCHeader` is backwards compatible, but in case we wanted to verify it, I think something like this works as an "upgrade test" without having to spin up a bunch of processes:

1. create a copy of the old struct definitions, maybe even just in a test package called `AppendEntriesRequestV3` and `RPCHeaderV3`
2. msgpack encode something with the new struct, and show it can be decoded with the old V3 structs
3. msgpack encode something with the V3 structs, and show it can be decoded correctly with the new structs

I did this once in a private repo and it seemed fine, but I thought I'd mention in case we want to be extra sure that it works.

raft.go
```diff
@@ -1739,7 +1774,7 @@ func (r *Raft) electSelf() <-chan *voteResult {
 		if server.Suffrage == Voter {
 			if server.ID == r.localID {
 				// Persist a vote for ourselves
-				if err := r.persistVote(req.Term, req.Candidate); err != nil {
+				if err := r.persistVote(req.Term, req.Addr); err != nil {
```
dnephin, line 1777: `req.Addr` is the new field, and all the other code checks the version. Do we need to have a version check here as well in case the code is still running in V3 mode?

raft.go
```diff
@@ -1479,11 +1488,32 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
+			r.logger.Warn("rejecting vote request since voter not part of the configuration",
+				"from", candidate)
+			return
+		}
+	}
+	if leaderAddr, leaderID := r.Leader(); leaderAddr != "" && leaderAddr != candidate && !req.LeadershipTransfer {
 		r.logger.Warn("rejecting vote request since we have a leader",
 			"from", candidate,
-			"leader", leader)
+			"leader (", leaderAddr, string(leaderID), ")")
```
dnephin, line 1516: This might render ok for the string formatted logs, but I think this is going to look really strange in JSON output.
```suggestion
			"leader", leaderAddr,
			"leaderID", string(leaderID))
```
The varargs is supposed to be a set of key/value pairs, so I think this should be :

net_transport_test.go
```diff
@@ -209,6 +211,7 @@ func TestNetworkTransport_AppendEntries(t *testing.T) {
 			},
 			LeaderCommitIndex: 90,
 		}
+		args.Addr = []byte("cartman")
```
dnephin, line 214: Not necessary, but if you are looking for a way to make this work inline in the struct definition, you can do this:

```go
RPCHeader: RPCHeader{Addr: []byte("cartman")},
```

raft.go
```diff
@@ -1479,11 +1488,32 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
-	candidate := r.trans.DecodePeer(req.Candidate)
-	if leader := r.Leader(); leader != "" && leader != candidate && !req.LeadershipTransfer {
+	var candidate ServerAddress
+	var candidateBytes []byte
+	if req.ProtocolVersion > 3 {
+		candidate = r.trans.DecodePeer(req.Addr)
+		candidateBytes = req.Addr
+	} else {
+		candidate = r.trans.DecodePeer(req.Candidate)
+		candidateBytes = req.Candidate
```
dnephin, line 1498: Similar comment here about if we need to do this. If we are still going to populate `req.Candidate` then maybe we keep using that field here?

Also similar comment about godoc. We should add a godoc to the `Candidate` field explaining whatever we decide (either that it's ignored in V4+, or that we continue to use it instead of the `RPCHeader.Addr`)

dnephin (commented):
Nice! I think this is looking good.

If I understand correctly, this PR does the following:
1. Increments to protocol version to 4, to indicate that we have new data available in the RPCHeader
2. Uses that new `ID` in the header to check if the voter is actually in the configuration before considering the vote.

That makes sense to me, and in general having the ID and addr of the server sending the message seems like it is useful.

A few questions, a few more style suggestions. I think the only blockers are the API backwards compat, and the one place we don't version check `req.Addr`.

But I do think if we can keep using `req.Leader` and `req.Candiate` when available, that might help keep the code easier to follow by reducing the number of places we check the protocol version.

I did not look at the tests much, hopefully Matt reviewed those more closely. Otherwise I can have a more detailed look.

dnephin (commented):
Thinking about the Protocol version increase, are you sure we need to increase the version?

Could we not optionally check if that field is set, and if it is do the new logic? If it's not then nothing we can do about it. That would avoid the difficult upgrade scenario, and allow everyone on V3 to get the benefit of this new check.

api.go
dhiaayachi: Yes we will need to do this or a v2. 

@mkeeler and I had a conversation about it and we where aligning for a v2 but looking at this a bit more I'm not sure it's worth doing a v2 only for this change?
What you think @mkeeler?

commands.go
dhiaayachi: Thanks!!

fuzzy/verifier.go
dhiaayachi: Yes good point, I always struggle to find all the references to an embedded field in my IDE because of that.

net_transport.go
dhiaayachi: My understanding is that we would like to remove this code at some point, I think the premise is that we keep 2 versions compatibility in the code. @mkeeler can give more details on this.
I will add the godoc for it.

raft.go
dhiaayachi: See my comment about removing this at some point. The reason we are keeping it is for backward compatibility but we would like to remove this at some point.

raft.go
dhiaayachi: yes I fixed some of those, this one slipped.

raft.go
dhiaayachi: It's not coming form the wire, we just populated it line 1746. So no need for a check here.

dhiaayachi (commented):
Thank you for reviewing this @dnephin, see my comments in line. 

I'm starting to think more that v4 is not necessary for this. Your point about using the new field and fallback to the old one when the new is empty seem reasonable to me and the fact that we always fill the old fields will ensure the compatibility with the nodes running a code version prior to this PR.

When I first did this with v4 version I was thinking that we will remove the Candidate and Leader field (or stop filling them in v4) but we ended up filling them all the time which ensure retro compatibility.

Testing the wire protocol compatibility is trivial and can easily be done the way you described.

What you think @mkeeler ?

dhiaayachi:
@dnephin, @mkeeler  I did the changes to rely on the `addr` field being empty instead of incrementing the `protocolVersion` to 4. Let me know what you think?

testing.go
```diff
@@ -325,6 +334,7 @@ CHECK:
 				num := len(fsm.logs)
 				fsm.Unlock()
 				if num != fsmLength {
+					fmt.Printf("fsm len : %d\n", num)
```
dnephin, line 337: left over debugging?

replication.go
```diff
@@ -381,8 +382,10 @@ func (r *Raft) heartbeat(s *followerReplication, stopCh chan struct{}) {
 	req := AppendEntriesRequest{
 		RPCHeader: r.getRPCHeader(),
 		Term:      s.currentTerm,
-		Leader:    r.trans.EncodePeer(r.localID, r.localAddr),
+		// this is needed for retro compatibility with protocolVersion = 3
```
dnephin, line 385: Are these comments about compat with version 3 still relevant?

fuzzy/verifier.go
```diff
@@ -45,7 +45,13 @@ func (v *appendEntriesVerifier) PreRequestVote(src, target string, rv *raft.Requ
 
 func (v *appendEntriesVerifier) PreAppendEntries(src, target string, req *raft.AppendEntriesRequest) (*raft.AppendEntriesResponse, error) {
 	term := req.Term
-	ldr := string(req.Leader)
+	var ldr string
+	if len(req.RPCHeader.Addr) > 0 {
+		ldr = string(req.RPCHeader.Addr)
+	} else {
+		ldr = string(req.Leader)
+	}
```
dnephin, line 53: Do we still need to do this? Any reason not to continue to use `req.Leader`? Why do we prefer the new `RPCHeader.Addr` ?

If we are trying to phase it out, we should document that on the `Leader` field.

Same question about the other two places we look at `Leader` or `Candidate`.

raft_test.go
```diff
@@ -1753,6 +1754,111 @@ func TestRaft_Voting(t *testing.T) {
...
+	// a follower that thinks there's a leader should vote for that leader.
+	var resp2 AppendEntriesResponse
+	if err := ldrT.AppendEntries(followers[0].localID, followers[0].localAddr, &reqAppendEntries, &resp2); err != nil {
+		t.Fatalf("RequestVote RPC failed %v", err)
+	}
+
+	require.True(t, resp2.Success)
+}
+
+func TestRaft_Voting_portocolVersion3(t *testing.T) {
```
dnephin, line 1835: Does this test name need to be updated now?

dnephin (commented):
Looking good! I definitely think this works better.

A couple comments about why we prefer the new field, and to document the existing fields accordingly.

I started to read over the tests a bit. I must admit some of the comments in the tests (which I think predate this PR) seem either misleading or confusing. I suspect I won't be able to review the test changes properly, but I trust Matt has done that.

raft.go
dhiaayachi: After discussion with @mkeeler this test is not really needed as we know it's safe to add parameters to msgpack

fuzzy/verifier.go
dhiaayachi: I think we should aim to phaseout Leader and Candidate parameters, it make more sense to infer those values from the RPCHeader now and removing those will reduce the on the wire packet size sensibly. Also if we keep them it mean we need to document/test special cases behaviour (different values, Addr empty and Candidate not...). 
I will add the deprecation comment for both Candidate and Leader

dhiaayachi:
@dnephin I will leave this open for a day or so otherwise I will merge it. I would like to start working on bringing this changes to Consul, I have a couple of fixes blocked on this.

dnephin (approved):
LGTM

PR #478, opened 2021-10-13
Add retract directive for v1.1.3

kisunji:
We had issues `go get`-ing raft@v1.1.3 due to checksum mismatch. It turns out we created tag v1.1.3, fetched it as a module, then deleted the tag and recreated it on GitHub, which violates the assumption that modules are immutable (once fetched from Go's checksum proxies).

This PR ensures that users will not accidentally add v1.1.3 as a dependency.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
   lint:
     executor:
       name: golang
-      version: "1.16"
+      version: "1.17"
     steps:
       - checkout
       - run: go mod download
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,14 +5,19 @@ workflows:
     jobs:
       - lint
       - go-test:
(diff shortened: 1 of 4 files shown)


kisunji:
Hm it looks like `retract` is incompatible with go <1.16. Is it worth bumping up the minimum go version for this fix or should we leave the problem as-is?

FWIW, v1.1.4 was published shortly after v1.1.3 and users fetching @latest won't face any issues

dhiaayachi (commented):
Thanks @kisunji  for adding this!!
Now That we support go `1.16` and `1.17` I think we should add linting to `1.17` as well.
I will remove the retract from the release/1.1.x branch because that one support go 1.14 we can't update it, we will need to upgrade consul 1.8 if we do so.

kisunji:
This is a small change; I hope it can get merged with next tag

PR #483, opened 2021-12-20
a NonVoter node should never be able to transition to a Candidate state

dhiaayachi:
This behaviour was observed in Consul for read-replicas nodes (which are raft nonVoters node). When the cluster is not stable (leader transition) there is a possibility to a nonVoter node to step up as leader (which is not supposed to happen)

@mkeeler and me where able to reproduce the behaviour of having a nonVoter node transition to a Candidate state, which is an indication that this node is possibly able to win an election. The test do the following steps:

1. Create a cluster with 3 Voters node and 1 nonVoter node and wait for stability
2. Remove one follower node to trigger a configuration update
3. after the leader send an `appendEntry` log to the nonVoter to update the latest configuration and before it send the `appendEntry` to commit the new configuration, partition the leader
4. Check that the nonVoter transition to a `Candidate` state

Unfortunately step 3 is very time sensitive and we were not able to come with a test that would consistently reproduce the issue but we were able to confirm that after the fix in this PR the issue could not be reproduced.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -214,15 +214,13 @@ func (r *Raft) runFollower() {
 				}
 			} else {
 				metrics.IncrCounter([]string{"raft", "transition", "heartbeat_timeout"}, 1)
-				if inConfig(r.configurations.latest, r.localID) {
+				if hasVote(r.configurations.latest, r.localID) {
 					r.logger.Warn("heartbeat timeout reached, starting election", "last-leader", lastLeader)
 					r.setState(Candidate)
 					return
-				} else {
-					if !didWarn {
-						r.logger.Warn("heartbeat timeout reached, not part of stable configuration, not triggering a leader election")
-						didWarn = true
-					}
+				} else if !didWarn {
+					r.logger.Warn("heartbeat timeout reached, not part of stable configuration, not triggering a leader election")
+					didWarn = true
 				}
 			}
 
--- a/raft.go
+++ b/raft.go
@@ -219,7 +219,7 @@ func (r *Raft) runFollower() {
 					r.setState(Candidate)
 					return
 				} else if !didWarn {
-					r.logger.Warn("heartbeat timeout reached, not part of stable configuration, not triggering a leader election")
+					r.logger.Warn("heartbeat timeout reached, not part of a stable configuration or a non-voter, not triggering a leader election")
 					didWarn = true
 				}
 			}
--- a/raft_test.go
+++ b/raft_test.go
@@ -2371,19 +2371,12 @@ func TestRaft_runFollower_State_Transition(t *testing.T) {
 			tt.fields.conf.CommitTimeout = 5 * time.Millisecond
 			tt.fields.conf.SnapshotThreshold = 100
 			tt.fields.conf.TrailingLogs = 10
+			tt.fields.conf.skipStartup = true
 
 			// Create a raft instance and set the latest configuration
 			env1 := MakeRaft(t, tt.fields.conf, false)
-			env1.raft.configurations.latest.Servers = tt.fields.servers
-			env1.raft.configurations.latestIndex = 1
-			env1.raft.state = Follower
-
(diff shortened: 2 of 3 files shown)


raft.go
```diff
@@ -214,15 +214,13 @@ func (r *Raft) runFollower() {
...
 					r.setState(Candidate)
 					return
-				} else {
-					if !didWarn {
-						r.logger.Warn("heartbeat timeout reached, not part of stable configuration, not triggering a leader election")
-						didWarn = true
-					}
+				} else if !didWarn {
+					r.logger.Warn("heartbeat timeout reached, not part of stable configuration, not triggering a leader election")
+					didWarn = true
```
banks, line 223: Is this warning message potentially confusing for this case now? We might be part of a stable configuration just as a non voter yet we'll now output this warning.

The message itself is kinda ambiguous so it might be a technicality since probably no-one understood what it meant anyway, but technically it is no longer necessarily true in the same sense as it was original intended (that this node is not in the committed config at all).

One suggestion would be to simplify this to be accurate and easier to understand: `heartbeat timeout reached, not in stable config or a non-voter so skipping leader election`. Or we could make it conditional and make it clear which one it is depending on whether it's a non-voter or not in the config?

dhiaayachi: I clarified the message without making it conditional as it make the code less complex.

banks (approved):
I think this looks good. It's a shame it's so hard to validate deterministically with a unit test. 

This is probably another case for our "one day" refactor of this library to make it easier to mock and drive in a controlled way in unit tests. I don't think we should block this fix on that though as it seems pretty clear that it's never right for a non-voter to enter Candidate state.

dhiaayachi:
> I think this looks good. It's a shame it's so hard to validate deterministically with a unit test.
> 
> This is probably another case for our "one day" refactor of this library to make it easier to mock and drive in a controlled way in unit tests. I don't think we should block this fix on that though as it seems pretty clear that it's never right for a non-voter to enter Candidate state.

@banks reading your comment about testing I had the idea to test this at a more unit level to at least avoid a future regression and have the transition out of follower state defined in a test. Let me know what you think about that?

raft_test.go
```diff
@@ -2345,3 +2345,47 @@ func TestRaft_InstallSnapshot_InvalidPeers(t *testing.T) {
...
+			// Create a raft instance and set the latest configuration
+			env1 := MakeRaft(t, tt.fields.conf, false)
+			env1.raft.setLatestConfiguration(Configuration{Servers: tt.fields.servers}, 1)
+			env1.raft.setState(Follower)
+
+			// run the follower loop exclusively
+			go env1.raft.runFollower()
+
+			// wait enough time to have HeartbeatTimeout
+			time.Sleep(tt.fields.conf.HeartbeatTimeout * 3)
```
banks, line 2385: I like having a test for this. I'm not sure how reliable this will be over time given that it's concurrent and timing dependent. Perhaps worth re-running this in CI a few times and seeing if it fails? We could increase the multiplier here if it does too I guess.

dhiaayachi: I ran it locally few times, I will try on the CI. The wait is deterministic because the follower loop will run after `HeartbeatTimeout` and there is no concurrency other then the test and the `runFollower` loop ( because we avoid starting the raft routines) so the state is in a controlled environment.

banks: Sounds low risk then. In the past we've found even "safe" waits like this can eventually end up causing flakes in highly CPU starved CI environments where `t.Parallel` means that many more tests are running than real CPU cores available and so scheduling delays can last hundreds of milliseconds. If we also avoid running in parallel that probably helps too.

For now it seems OK just wanted to note that based on previous tests that have ended up flaky even 3 * heartbeat might not be enough in the long run!

dhiaayachi: Yes in consul we have a lot of those. I will retry the CI a couple of times before merging to see how it hold.

PR #484, opened 2022-01-07
Tuning performance and prevent leak due to time.After

linxGnu:
First of all, thank you so much for incredible library. 

This PR is an attempt to improve the performance and prevent high memory usage due to timer leak without calling `Stop`.

**Motivation**

Many hot paths in Raft library introduce `time.After` usage. While getting out of scope (but not timeout), those timers are not stopped/recycled by GC yet (until event is fired and pushed to channel); thus pressure to runtime with unnecessary ops.

Moreover, `time.After` actually allocates new `time.Timer`. There is no benefit of using `time.After` over `time.Timer`.

**The patch**

Objectives:
- To make viewer easier to read
- Reduce code changes as much as possible
- Reusable

The patch introduces wrapper for timer which acts similar as `time.After`. Just a note that we need to stop it when getting out of scope (manually or using `defer`)

I would like to avoid `defer timer.Stop()` as much as possible. It's fine for some case with recent optimization introduced in go1.16, but not all.

**Benchmark**

```
go version: 1.17.6
CPU: AMD Ryzen 5850U
goos: linux
goarch: amd64
pkg: github.com/hashicorp/raft
BenchmarkTimeAfter-16                295           3968892 ns/op         2100006 B/op      30000 allocs/op
BenchmarkTimer-16                    451           2738084 ns/op         2000056 B/op      30000 allocs/op
````

```go
const timerBenchAttempt = 10000

func BenchmarkTimeAfter(b *testing.B) {
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for attempt := 0; attempt < timerBenchAttempt; attempt++ {
			select {
			case <-time.After(50 * time.Millisecond):
			default:
			}
		}
	}
}

func BenchmarkTimer(b *testing.B) {
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for attempt := 0; attempt < timerBenchAttempt; attempt++ {
			timer := newTimer(50 * time.Millisecond)
			select {
			case <-timer.C:
			default:
				timer.Stop()
			}
		}
	}
}
```


api.go
```diff
@@ -737,10 +737,8 @@ func (r *Raft) Apply(cmd []byte, timeout time.Duration) ApplyFuture {
 func (r *Raft) ApplyLog(log Log, timeout time.Duration) ApplyFuture {
 	metrics.IncrCounter([]string{"raft", "apply"}, 1)
 
-	var timer <-chan time.Time
-	if timeout > 0 {
-		timer = time.After(timeout)
-	}
+	timer := newTimer(timeout)
+	defer timer.Stop()
```
dnephin, line 741: I think we can accomplish the same thing without wrapping `time.Timer`

```suggestion
	var timer <-chan time.Time
	if timeout > 0 {
		t := time.Newtimer(timeout)
		defer t.Stop()
		timer = t.C
	}
```

This is a few extra lines, but may actually result in a smaller diff. I think in the long term not having a wrapper around `time.Timer` will be easier to maintain.

inmem_transport.go
```diff
@@ -178,9 +180,11 @@ func (i *InmemTransport) makeRPC(target ServerAddress, args interface{}, r io.Re
 		if rpcResp.Error != nil {
 			err = rpcResp.Error
 		}
-	case <-time.After(timeout):
+	case <-timer.C:
 		err = fmt.Errorf("command timed out")
 	}
+
+	timer.Stop()
```
dnephin, line 187: I think it's currently safe to call Stop here, but in the future we could add other returns. It might be good to keep this as a `defer` after creating the timer, both for consistency, and to prevent future regressions. Is there a reason that `defer` won't work in this case?

log_cache_test.go
```diff
@@ -137,8 +137,8 @@ func TestLogCacheWithBackendStoreError(t *testing.T) {
 
 	var out Log
 	for i := 1; i < 5; i++ {
-		if err := c.GetLog(uint64(i), &out); err != nil {
-			t.Fatalf("err: %v", err)
+		if e := c.GetLog(uint64(i), &out); e != nil {
+			t.Fatalf("err: %v", e)
```
dnephin, line 141: What's the reason for this change? Generally `err` is the correct variable name to use for errors. If it's because it's masking `err` from the parent scope, I think that's not really a concern in this test.

net_transport_test.go
```diff
@@ -84,16 +84,16 @@ func TestNetworkTransport_CloseStreams(t *testing.T) {
...
-			}
-
-			// Verify the response
-			if !reflect.DeepEqual(resp, out) {
-				t.Fatalf("command mismatch: %#v %#v", resp, out)
+				atomic.AddInt32(&errCounter, 1)
+				t.Errorf("err: %v", err)
+			} else if !reflect.DeepEqual(resp, out) { // Verify the response
+				atomic.AddInt32(&errCounter, 1)
+				t.Errorf("command mismatch: %#v %#v", resp, out)
```
dnephin, line 96: Thank you for cleaning up the calls to `t.Fatal` in a goroutine! 

While this is good cleanup, I think it would be better to do this cleanup in a separate PR, and at the very least in a separate commit. Keeping commits isolated to one change makes the git history much easier to read, and can make it easier to cherry-pick changes.

I'd also like to suggest a different approach to fixing this problem. Instead of an atomic counter, how about returning an error, and using https://pkg.go.dev/golang.org/x/sync/errgroup to run the goroutines. That is a pattern I've used before, and it works really well. That might look something like this:


```go
	appendFunc := func() error {
		// return errors instead of t.Fatalf
	}

	g := new(errgroup.Group)
	for i = 0; i < 5; i++ {
		g.Go(appendFunc())
	}
	require.NoError(t, g.Wait())
```

This will remove the need for a `sync.WaitGroup` as well.

util.go
```diff
@@ -28,12 +28,12 @@ func newSeed() int64 {
...
 // randomTimeout returns a value that is between the minVal and 2x minVal.
-func randomTimeout(minVal time.Duration) <-chan time.Time {
+func randomTimeout(minVal time.Duration) timerWrapper {
 	if minVal == 0 {
-		return nil
+		return newTimer(0)
 	}
 	extra := (time.Duration(rand.Int63()) % minVal)
-	return time.After(minVal + extra)
+	return newTimer(minVal + extra)
```
dnephin, line 36: I think we can make this function work without the `timerWrapper` by returning a `noopTimer` implementation instead of `nil`. The `noopTimer` can have a method `Stop`  that does nothing, and a field `C` that is a nil channel. Then we can return the `noopTimer` or  a `*time.Timer`

dnephin (changes requested):
Thank you for working on this improvement!

Many of these calls to `time.After` are safe in practice, because the only way for the function to exit without the timer firing is when raft shuts down. I think in general we expect the entire process to exit soon after raft is shutdown, so in most cases that leak is not a concern, and not a risk. But that assumption may not always be true, so it's probably good to fix all of them. I think there is also a good argument to be made for fixing these cases as a way of demonstrating good practices.

There are definitely a couple cases on `time.After` that are concerning, specifically the ones `snapshot.go`, `replication.go`, and `raft.go`.

My main concern with this approach is the new wrapper around `time.Timer`. I think we can fix these calls without introducing the new type, which will be easier to maintain in the long term.

I'd also like to request that you split up all the cleanup changes into at least a separate commit, and possibly a separate PR. Those clean ups are appreciated, but it makes the core changes harder to review when they are mixed into the same commit.

inmem_transport.go
linxGnu: I agree with you. 

Not using `defer` here is just a trade-off between safety (future changes) and performance (tiny improvement). Obviously, not calling `timer.Stop` would result in the same manner as leaking `time.After`.

I would fix that as your suggestion. Thank you so much.

net_transport_test.go
linxGnu: That's great idea to use `ErrGroup`. I would make the fix in another PR as your suggestion.

util.go
linxGnu: I love the idea of `noopTimer`. I would fix that.

dnephin: Thinking about this more, I'm not sure how to return this from `randomTimeout`. We could use an interface, but interfaces don't allow fields.   I'm not sure if there is any way to make this work.

linxGnu: It's possible to use interface and follow SOLID principle here. But the trade-off might look like:
```go
t.C -> t.C() // return channel through func 
```
If it looks okay to you, then I think interface is a good choice rather than concrete type

linxGnu:
@dnephin Follow up your comments, I have reverted changes and keep the PR small as it should be. I believe the main concern left is `timerWrapper`. Could you please take a look.

Thank you very much.

inmem_transport.go
```diff
@@ -287,6 +288,9 @@ func (i *inmemPipeline) decodeResponses() {
 			case <-i.shutdownCh:
 				return
 			}
+
+			timer.Stop()
```
kisunji, line 292: This should be deferred on L267 in case of an early return in the `select` block above

util.go
```diff
@@ -150,3 +150,28 @@ type uint64Slice []uint64
...
+		t.C = t.t.C
+	}
+	return
+}
+
+func (t *timerWrapper) Reset(d time.Duration) {
+	if t.t != nil {
+		t.t.Reset(d)
+	}
+}
```
kisunji, line 171: I think `Reset` will need to handle the case where `t == nil` and call `newTimer`

```go
t := newTimer(0)
t.Reset(1000)
```
Currently, calling `Reset` will not behave like `(*time.Timer).Reset` in this case

api.go
kisunji: When reviewing this PR, I had to pause and locally test the behaviour of `timerWrapper`  to gain confidence that it would behave how I expect (and found a potential bug with `(timerWrapper).Reset`).

From a maintainer's perspective, I think code duplication is an acceptable cost here. The wrapper certainly saves many lines of code but may increase developer time spent reading and understanding the wrapper type.

I'd like to request that we go with @dnephin's suggestion for using the std lib directly

kisunji (changes requested):

inmem_transport.go
linxGnu: @kisunji if we put defer in L276, it will be defer in loop and absolutely bad practice. How do you think?

linxGnu:
@kisunji imho, code duplication might be a lot. We have to change not only one place in my PR but many (and in many files).

I dont think the wrapper breaks any std lib way. And its quite small. Easy to read and understand. Dup code might introduce error prone or miss required specs.

inmem_transport.go
kisunji: Ah, that's a good catch. We may have to call `timer.Stop` in all of the branches that `return` early then

linxGnu: It depends on each logic. As far as I look at the source code. We just need to invoke stop at the end of select or just inside a branch.

linxGnu:
Kindly ping @dnephin @kisunji, there is a fix and merged similar issue at Nomad. Should we continue on this or you wanna applye same mechanism (newSafeTimer)?

imho its no different between newsafetimer and newtimer here. Both of them require stop func invoking.

kisunji:
Hi @linxGnu, while this fix is valuable, more than one maintainer has expressed concerns around `timerWrapper`. We understand Nomad has implemented a wrapper for this use-case, but for Raft (which doesn't change frequently and is maintained by volunteers that come and go) we would like to minimize the introduction of new types to this codebase.

Internally we prefer that this PR uses stdlib directly to make the code obvious at a glance.

inmem_transport.go
```diff
@@ -260,8 +268,10 @@ func (i *inmemPipeline) decodeResponses() {
 		select {
 		case inp := <-i.inprogressCh:
 			var timeoutCh <-chan time.Time
+			var timer *time.Timer
 			if timeout > 0 {
-				timeoutCh = time.After(timeout)
+				timer = time.NewTimer(timeout)
```
linxGnu, line 273: Inside a loop -> won't use `defer` here

replication.go
```diff
@@ -402,9 +401,9 @@ func (r *Raft) heartbeat(s *followerReplication, stopCh chan struct{}) {
 			r.logger.Error("failed to heartbeat to", "peer", peer.Address, "error", err)
 			r.observe(FailedHeartbeatObservation{PeerID: peer.ID, LastContact: s.LastContact()})
 			failures++
-			select {
-			case <-time.After(backoff(failureWait, failures, maxFailureScale)):
-			case <-stopCh:
```
linxGnu, line 407: this is a bug when received from `stopCh` but does not return. This PR also fix this

replication.go
```diff
@@ -206,9 +206,8 @@ func (r *Raft) replicateTo(s *followerReplication, lastIndex uint64) (shouldStop
 START:
 	// Prevent an excessive retry rate on errors
 	if s.failures > 0 {
-		select {
-		case <-time.After(backoff(failureWait, s.failures, maxFailureScale)):
-		case <-r.shutdownCh:
+		if isTerminated(r.shutdownCh, backoff(failureWait, s.failures, maxFailureScale)) {
+			return
```
linxGnu, line 210: this is a bug when received from `shutdownCh` but does not return. This PR also fix this

linxGnu:
@kisunji @dnephin I addressed your comments although I am not convinced by `copy-paste` approach, no-reusing.

My patch also fix some small bugs of missing return when stop/terminate channel is notified. Please take a look.

banks:
Hi 👋 

Just cleaning up some stale issues and PRs. We want to thank you again @linxGnu for the time you spent on this PR and responding to feedback.

At this point it's not very clear to us that these changes fix any real issues in this library - not ones that our usage has ever seen as a problem anyway. So on balance the cost and risk of checking, testing and merging the PR and releasing it into our products seems to outweigh any benefit we might get from it.

It always feels bad saying no to a contribution but we'd rather not leave this open forever if we're not likely to merge it!

Thanks again!

Issue #488, opened 2022-01-31
Add saturation metric for the runFSM and run (main) goroutines.


dnephin:
The two primary goroutines used by Raft (runFSM, and run (main)) are single threaded operations. They can saturate (take more than 100% of the available time to handle the incoming workload) before the CPU of a system reaches 100% utilization. 

When this happens it may be possible to observe the problem using some existing metrics (ex: `fsm.apply` time), but properly interpreting those metrics requires deep knowledge of how raft works. It may also be a challenge to present the data on a dashboard because it requires summing the time, and knowing the aggregation period of the metrics to interpret the summed result.

The existing metrics may also not fully capture the time, because they only measure specific operations done by those goroutines, not the full work vs idle time.

This issue proposes adding two new metrics (one for each goroutine) which measure the amount of time those goroutines spent doing work. When compared to the wall clock time, this gives us a clear signal about the saturation of these operations, and how much buffer there is before the incoming work starts to cause a backlog.

PR #489, opened 2022-02-02
Thread saturation metrics 📈

boxofrad:
Adds metrics suggested in #488, to record the percentage of time the main and FSM goroutines are busy with work vs available to accept new work, to give operators an idea of how close they are to hitting capacity limits.

We update gauges (at most) once a second, possibly less if the goroutines are idle. This should be ok because it's unlikely that a goroutine would go from very high saturation to being completely idle; so at worst we'll leave the gauge on the previous (low) value for a while.

### DIFF
--- a/api.go
+++ b/api.go
@@ -208,9 +208,8 @@ type Raft struct {
 	// followerNotifyCh is used to tell followers that config has changed
 	followerNotifyCh chan struct{}
 
-	// thread saturation metric recorders.
+	// mainThreadSaturation measures the saturation of the main raft goroutine.
 	mainThreadSaturation *saturationMetric
-	fsmThreadSaturation  *saturationMetric
 }
 
 // BootstrapCluster initializes a server's storage with the given cluster
@@ -557,8 +556,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		leadershipTransferCh:  make(chan *leadershipTransferFuture, 1),
 		leaderNotifyCh:        make(chan struct{}, 1),
 		followerNotifyCh:      make(chan struct{}, 1),
-		mainThreadSaturation:  newSaturationMetric([]string{"raft", "mainThreadSaturation"}, 1*time.Second),
-		fsmThreadSaturation:   newSaturationMetric([]string{"raft", "fsm", "threadSaturation"}, 1*time.Second),
+		mainThreadSaturation:  newSaturationMetric([]string{"raft", "thread", "main", "saturation"}, 1*time.Second, 5),
 	}
 
 	r.conf.Store(*conf)
--- a/api.go
+++ b/api.go
@@ -556,7 +556,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		leadershipTransferCh:  make(chan *leadershipTransferFuture, 1),
 		leaderNotifyCh:        make(chan struct{}, 1),
 		followerNotifyCh:      make(chan struct{}, 1),
-		mainThreadSaturation:  newSaturationMetric([]string{"raft", "thread", "main", "saturation"}, 1*time.Second, 5),
+		mainThreadSaturation:  newSaturationMetric([]string{"raft", "thread", "main", "saturation"}, 1*time.Second),
 	}
 
 	r.conf.Store(*conf)
--- a/api.go
+++ b/api.go
@@ -207,6 +207,10 @@ type Raft struct {
 
 	// followerNotifyCh is used to tell followers that config has changed
 	followerNotifyCh chan struct{}
+
+	// thread saturation metric recorders.
+	mainThreadSaturation *saturationMetric
+	fsmThreadSaturation  *saturationMetric
 }
 
 // BootstrapCluster initializes a server's storage with the given cluster
@@ -553,6 +557,8 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		leadershipTransferCh:  make(chan *leadershipTransferFuture, 1),
 		leaderNotifyCh:        make(chan struct{}, 1),
 		followerNotifyCh:      make(chan struct{}, 1),
+		mainThreadSaturation:  newSaturationMetric([]string{"raft", "mainThreadSaturation"}, 1*time.Second),
+		fsmThreadSaturation:   newSaturationMetric([]string{"raft", "fsm", "threadSaturation"}, 1*time.Second),
 	}
 
 	r.conf.Store(*conf)
--- a/saturation.go
+++ b/saturation.go
@@ -1,6 +1,7 @@
 package raft
 
 import (
+	"math"
 	"time"
 
 	"github.com/armon/go-metrics"
@@ -128,11 +129,12 @@ func (s *saturationMetric) report() {
 
 	total := now.Sub(beginning) - lost
 
-	var saturation float32
+	var saturation float64
 	if total != 0 {
-		saturation = float32(total-slept) / float32(total)
+		saturation = float64(total-slept) / float64(total)
+		saturation = math.Round(saturation*100) / 100
 	}
-	s.reportFn(saturation)
+	s.reportFn(float32(saturation))
 
 	s.currentBucket++
 	if s.currentBucket >= len(s.buckets) {
--- a/saturation.go
+++ b/saturation.go
@@ -1,7 +1,6 @@
 package raft
 
 import (
-	"math"
 	"time"
 
 	"github.com/armon/go-metrics"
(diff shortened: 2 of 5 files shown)


saturation.go
```diff
@@ -0,0 +1,139 @@
...
+// sleeping records the time at which the caller began waiting for work. After
+// the initial call it must always be proceeded by a call to working.
+func (s *saturationMetric) sleeping() {
+	s.sleepBegan = s.nowFn()
+
+	// Caller should've called working (we've probably missed a branch of a
+	// select). Reset sleepBegan without recording a sample to "lose" time
+	// instead of recording nonsense data.
+	if s.index%2 == 1 {
+		return
```
boxofrad, line 65: Initially, I had this panicking, which I think is the more correct behavior. But given how easy it'd be to forget to instrument a new branch in `runFollower`, `runLeader`, or `runCandidate`, I think this is the safer bet!

api.go
```diff
@@ -540,6 +544,8 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		bootstrapCh:           make(chan *bootstrapFuture),
 		observers:             make(map[uint64]*Observer),
 		leadershipTransferCh:  make(chan *leadershipTransferFuture, 1),
+		mainThreadSaturation:  newSaturationMetric([]string{"raft", "mainThreadSaturation"}, 1*time.Second),
+		fsmThreadSaturation:   newSaturationMetric([]string{"raft", "fsm", "threadSaturation"}, 1*time.Second),
```
boxofrad, line 548: Not 100% on these names, what do you think?

saturation.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// index of the current sample. We rely on it wrapping-around on overflow and
+	// underflow, to implement a circular buffer (note the matching size of the
+	// samples array).
+	index uint8
+
+	// samples is a fixed-size array of samples (similar to a circular buffer)
+	// where elements at even-numbered indexes contain time spent sleeping, and
+	// elements at odd-numbered indexes contain time spent working.
+	samples [math.MaxUint8 + 1]struct {
```
dnephin, line 32: Nice! Using a fixed size array seems like it really simplifies the tracking. No need to delete stale samples out of a map.

dnephin (commented):
Nice! This is looking good.

Mostly I have questions about the `saturationMetric` implementation. Originally you had a goroutine to report the times. Did you notice any difference in bias toward idle vs work time by removing it and reporting the time synchronously?

I'm wondering if by recording both idle time and work time we reduce the likelihood that the metric values goes the 1.0.  I'm wondering if we could track only idle time, and assume anything not  in idle time is time spent doing work.  I started to experiment with that approach in [this branch](https://github.com/hashicorp/raft/compare/boxofrad/saturation-metrics...dnephin/saturation-metrics?expand=1), but it's definitely not working yet (the tests fail).

Do you think an approach that only tracks idle time could work? My hope is that better represents the full saturation and that the metric will go to 1.0 more easily.

My second question about the implementation is if we are missing some time. We have 256 samples for a 1 second interval. I think that means if the mean time is shorter than 4s we will clobber some of the earlier samples. If we start the interval with lots of work, then a bunch of fast work comes in afterward. I think we would under-report the saturation.

In the branch I linked I continued to use a fixed size array, but use a different value for the index.
Do you think an approach of using  `now.UnixNano() / int64(10*time.Millisecond) % math.MaxUint8` to find the right sample bucket, and appending the idle time to that bucket could work?  I think that by dividing by 10ms we can track up to 2.56s work of times. If we're idle for longer than that then we can still report a low saturation time on the next cycle because values won't have been erased yet.

boxofrad:
Those are great points @dnephin, thanks! 🙌🏻 

I think the approach in your branch makes a lot of sense. It also made me realise that if we kept a simple running accumulator of sleep time (rather than a slice of samples) we could subtract it from the time elapsed since `lastReport`, and presumably wouldn't need to lose any time at all? 

Here's a rough implementation: https://github.com/hashicorp/raft/commit/25ba633b45f1398461cf5d6e83274f1c6599f356, let me know what you think!

dnephin:
Ahh, interesting! I think this approach is a good simplification of what we had, but I wonder if it points out a potential improvement we could make by using a slice of samples.

If we only record elapsed time (no buckets/slice), then every time we report the metric we will always be reporting just the change since last report.  

If we continued to use buckets, we could potentially take the saturation for an arbitrary period that includes time before the last report. That might help stabalize the value a little bit and make it less spiky. I guess with 10ms buckets we don't do much normalization, but if we were to change the buckets to 100ms, then we could easily keep 10s+ of samples. Each time we report the metric the samples from previous buckets would still be included in the idle/total ratio, which I think would help smooth out the values a bit.

boxofrad:
Nice, yeah I think that would help a lot! Do you think we'd need to weight/bias towards more recent values to prevent it from hiding genuine spikes? I guess if the bucket-size is smaller it wouldn't matter as much?

dnephin:
Good question! Ya, it totally depends on bucket-size. As long we the bucket size is small enough we should have a good balance without having to apply weights I think.

saturation.go
```diff
@@ -0,0 +1,148 @@
...
+
+// saturationMetric measures the saturation (percentage of time spent working vs
+// waiting for work) of an event processing loop, such as runFSM. It reports the
+// saturation as a gauge metric (at most) once every reportInterval.
+//
+// Callers must instrument their loop with calls to sleeping and working, starting
+// with a call to sleeping.
+//
+// Note: it is expected that the caller is single-threaded and is not safe for
+// concurrent use by multiple goroutines.
```
kisunji, line 18: This read a bit weird (caller is not safe for concurrent use?) so I attempted to simplify:
```suggestion
// Note: saturationMetric is not thread-safe and should not be used concurrently
// by multiple goroutines
```

kisunji, line 18: This read a bit weird (caller is not safe for concurrent use?) so I attempted to simplify:
```suggestion
// Note: saturationMetric is not thread-safe and should not be used concurrently
// by multiple goroutines
```

saturation.go
```diff
@@ -0,0 +1,148 @@
...
+type saturationMetric struct {
+	reportInterval time.Duration
```
kisunji, line 20: nit: could this be simply `interval`?

kisunji, line 20: nit: could this be simply `interval`?

saturation.go
```diff
@@ -0,0 +1,148 @@
...
+// saturationMetricBucket contains the measurements for a period.
+type saturationMetricBucket struct {
+	// beginning of the period for which this bucket contains measurements.
+	beginning time.Time
+
+	// slept contains time for which the event processing loop was sleeping rather
+	// than working.
+	slept time.Duration
+
+	// lost contains time that was lost due to incorrect instrumentation of the
```
kisunji, line 43:
```suggestion
	// lost contains time that should be considered  lost due to incorrect instrumentation of the
```
```suggestion
	// lost contains time that was lost due to incorrect instrumentation of the
```

saturation.go
```diff
@@ -0,0 +1,148 @@
...
+	// event processing loop (e.g. calling sleeping() or working() multiple times
+	// in succession).
```
kisunji, line 45:
```suggestion
	// lost contains time that is considered lost due to incorrect use
	// of saturationMetricBucket (e.g. calling sleeping() or working() 
	// multiple times in succession).
```

kisunji, line 45:
```suggestion
	// lost contains time that should be considered lost due to incorrect 
	// use of saturationMetricBucket (e.g. calling sleeping() or working() 
	// multiple times in succession).
```

kisunji: My reading of the code is that your functions _account_ for this time lost, but the original wording made it seem like the reporting would be wrong with incorrect use (correct me if I'm wrong!)

kisunji (approved):
This is really neat! LGTM with minor nits about comments

saturation.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// saturationMetric measures the saturation (percentage of time spent working vs
+// waiting for work) of an event processing loop, such as runFSM. It reports the
+// saturation as a gauge metric (at most) once every reportInterval.
+//
+// Callers must instrument their loop with calls to sleeping and working, starting
+// with a call to sleeping.
+//
+// Note: it is expected that the caller is single-threaded and is not safe for
```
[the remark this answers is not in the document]
boxofrad: Much better, thank you! 🙇🏻‍♂️

kisunji:
My github is acting up and left duplicate comments everywhere 😕 Sorry if it confused you as well

saturation.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"math"
+	"time"
+
+	"github.com/armon/go-metrics"
+)
+
+// saturationMetric measures the saturation (percentage of time spent working vs
+// waiting for work) of an event processing loop, such as runFSM. It reports the
+// saturation as a gauge metric (at most) once every reportInterval.
```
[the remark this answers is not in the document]
boxofrad: Oh sure! What I was trying to get at here is that it's possible that the goroutine is asleep or busy for a period longer than `reportInterval` so we actually update the gauge less frequently. It shouldn't really be a problem as far as the metric's accuracy goes, because it's unlikely you'd drop from high saturation to dormant suddenly or vice versa.

Maybe this would be clearer:

> It attempts to report the saturation as a gauge metric once every `reportInterval`, but it may happen less frequently if the event loop is sleeping or busy for a long period.

saturation.go
boxofrad: Sure! I don't feel strongly either way 😄

boxofrad:
No worries, thanks for the review @kisunji 🙌🏻 I really appreciate the clarifications you made to the doc comments.

saturation.go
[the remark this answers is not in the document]
kisunji: Ah that makes sense!

saturation.go
```diff
@@ -0,0 +1,148 @@
...
+
+// newSaturationMetric creates a saturationMetric that will update the gauge
+// with the given name at the given reportInterval. keepPrev determines the
+// number of previous measurements that will be used to smooth out spikes.
+func newSaturationMetric(name []string, reportInterval time.Duration, keepPrev uint) *saturationMetric {
+	m := &saturationMetric{
+		reportInterval: reportInterval,
+		buckets:        make([]saturationMetricBucket, 1+keepPrev),
+		nowFn:          time.Now,
+		reportFn:       func(sat float32) { metrics.SetGauge(name, sat) },
```
FFMMM, line 57: Did we pre-declare this new metric anywhere?

mkeeler, line 57:
```suggestion
		reportFn:       func(sat float32) { metrics.AddSample(name, sat) },
```

Using `AddSample` will allow us to do aggregations over time. So if we are scraping metrics every 10s we don't just get the saturation for the last second but rather get  the min/max/mean/count sorts of metrics.

mkeeler (changes requested):

saturation.go
boxofrad: Nice! Thanks for the recommendation, pushing this into go-metrics would remove a lot of code 🙌🏻

mkeeler (approved):

PR #492, opened 2022-03-01
a NonVoter node should never try to bootstrap

rboyer:
This is a follow-up to #483

TODO:

- [x] add a test (likely inspired by consul's `TestServer_Expect_NonVoters` in some way since that's how i noticed the bug)


mkeeler (approved):
LGTM

PR #493, opened 2022-03-08
Make sure leadership transfer is set back to false in all cases.

VictorLowther:
Release 1.3.4 introduced a bug that would only call setLeadershipTransferInProgress back to false
when leadership transfer to another node succeeded.  In the case where it did not succeed,
leadershipTransferInProgress was left set to true, breaking any callers that relied on it being set back
to false after a failed leadership transfer.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -581,10 +581,6 @@ func (r *Raft) leaderLoop() {
 	// based on the current config value.
 	lease := time.After(r.config().LeaderLeaseTimeout)
 
-	// This would unset leadershipTransferInProgress
-	// in case it was set during the loop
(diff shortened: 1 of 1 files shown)


dhiaayachi (approved):
Thank you for working on this @VictorLowther. 

I think this is a reasonable fix, I won't worry about the gap because it was already present before this.

I will leave this open until early next week to give others the chance to take a look, otherwise I will merge it.

VictorLowther:
@dhiaayachi Just touching base to see if there is an ETA for getting this merged yet.

dhiaayachi:
@VictorLowther sorry for the delay, I will merge this.

PR #494, opened 2022-03-09
Cap maximum grpc wait time when heartbeating to heartbeatTimeout/2

HridoyRoy:
This PR aims to resolve https://hashicorp.atlassian.net/browse/VAULT-5310 , and the associated GH vault issue https://github.com/hashicorp/vault/issues/14153. 

Context: Currently, if the a follower is shut down for some amount of time, the leader cannot heartbeat to it and does an exponential backoff. When the follower restarts, the leader waits for longer than election_timeout to send a heartbeat to it, thus the follower starts an election and increases its term, which causes leadership to change.

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -2639,9 +2639,7 @@ func TestRaft_VoteNotGranted_WhenNodeNotInCluster(t *testing.T) {
 	}
 }
 
-// ROY
-// Verifies that when a follower is taken out of rotation for heartbeat timeout
-// a leader election does not occur.
+// This test is currently a stub
 func TestRaft_FollowerRemovalNoElection(t *testing.T) {
 	// Make a cluster
 	c := MakeCluster(3, t, nil)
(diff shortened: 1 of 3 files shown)


util.go
```diff
@@ -144,6 +144,20 @@ func backoff(base time.Duration, round, limit uint64) time.Duration {
 	return base
 }
 
+// cappedExponentialBackoff computes the exponential backoff with an adjustable
+// cap on the max timeout.
+func cappedExponentialBackoff(base time.Duration, round, limit uint64, cap time.Duration) time.Duration {
+	power := min(round, limit)
+	for power > 2 {
+		if base > cap {
+			return base
```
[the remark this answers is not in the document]
HridoyRoy: Thanks for catching! Updated!

util.go
```diff
@@ -144,6 +144,20 @@ func backoff(base time.Duration, round, limit uint64) time.Duration {
...
+func cappedExponentialBackoff(base time.Duration, round, limit uint64, cap time.Duration) time.Duration {
+	power := min(round, limit)
+	for power > 2 {
+		if base > cap {
+			return cap
+		}
+		base *= 2
+		power--
+	}
+	return base
```
briankassouf, line 158: I think there is still potential for base > cap here. Probably we should move the if condition to right before the return value. 
```suggestion
	if base > cap {
		return cap
	}
	return base
```

replication.go
```diff
@@ -402,12 +402,15 @@ func (r *Raft) heartbeat(s *followerReplication, stopCh chan struct{}) {
 
 		start := time.Now()
 		if err := r.trans.AppendEntries(peer.ID, peer.Address, &req, &resp); err != nil {
-			r.logger.Error("failed to heartbeat to", "peer", peer.Address, "error", err)
+			nextBackoffTime := cappedExponentialBackoff(failureWait, s.failures, maxFailureScale, r.config().HeartbeatTimeout/2)
```
briankassouf, line 405: The code changed from using the local `failures` variable to `s.failures`. What's the difference here?

util.go
HridoyRoy: Makes sense!

replication.go
HridoyRoy: Hm, I'm not sure -- it was an unintentional change, so I'll revert it. It's interesting though that s.failures is used for all the other backoffs, but not for the heartbeat backoffs. `s` is followerReplication, and the comment above the struct says: 

`// followerReplication is in charge of sending snapshots and log entries from
// this leader during this particular term to a remote follower.`

and since heartbeats are empty log entries, it makes sense that s.failures would get incremented and used for backoff on failed heartbeats. On the other hand, s.failures >= failures because failures is local in `heartbeat` and initialized at 0, so it's probably better to revert to failures. 

Thanks for catching!

util.go
```diff
@@ -144,6 +144,23 @@ func backoff(base time.Duration, round, limit uint64) time.Duration {
...
 }
 
+// cappedExponentialBackoff computes the exponential backoff with an adjustable
+// cap on the max timeout.
+func cappedExponentialBackoff(base time.Duration, round, limit uint64, cap time.Duration) time.Duration {
+	power := min(round, limit)
+	for power > 2 {
+		if base > cap {
+			return cap
+		}
```
briankassouf, line 154:
```suggestion
```

raft_test.go
```diff
@@ -2639,6 +2639,66 @@ func TestRaft_VoteNotGranted_WhenNodeNotInCluster(t *testing.T) {
 	}
 }
 
+// TestRaft_FollowerRemovalNoElection ensures that a leader election is not
+// started when a standby is shut down and restarted.
+func TestRaft_FollowerRemovalNoElection(t *testing.T) {
+	// Make a cluster
+	inmemConf := inmemConfig(t)
+	inmemConf.HeartbeatTimeout = 500 * time.Millisecond
```
mkeeler, line 2647: If you reduce these to maybe 100ms each then you could reduce the 3s wait later on and ensure that this test runs much more quickly.

The total time to get to the capped backoff would be no more than 170ms:
* <10ms - standard wait between heartbeats.
* 10ms - first error wait 
* <10ms - standard wait
*  20ms - second error wait
* <10ms - standard wait
*  40ms - third error wait
* <10ms - standard wait
*  50ms - fourth error wait capped due to 50% of 100ms

Then we repeat the same <60ms wait for each subsequent error. In total there is about 160ms of exponential backoff in there so you could sleep 250ms just to be on the safe side. That would reduce this tests runtime by about 83% if my calculations are correct.

mechpen:
hi, I saw this behavior as well.

When a voter stops, the leader's backoff mechanism waits for more than 10 seconds (10ms*1024) to send heartbeats and replicate messages to the voter.  When the voter comes back up, it times out and triggers a new election.  This could cause a leadership change or flap.

Does it make sense to disable backoff for heartbeats?

mkeeler:
@mechpen We normally heartbeat every 1/10th of the heartbeat timeout. When these fail other warnings/errors are emitted in the logs. In these disconnected scenarios we are attempting to perform exponential backoff to ensure we don't needlessly fill up logs and use more network bandwidth than necessary. The bug we have encountered is where we backoff for far too long and get into the situation where a restarted server may hit its heartbeat timeout before the leader attempts it again.

The solution proposed in this PR just drastically lowers the cap on how much we can backoff of the usual rate to ensure that we always send a heartbeat within the timeout value. It reduces the cap enough that practically speaking we only allow backoff by a factor of 5x the original value as opposed to the 100x it was previously. That 5x though will mean there are 1/5 the warning logs which could make figuring out whats going on during an incident a tiny bit easier so I think the solution implemented in this PR is probably the way to go.

raft_test.go
HridoyRoy: Sounds good -- removed the time.Sleep after the node spun down, and changed the time.Sleep to 250 ms during which we're waiting for a re-election.

PR #496, opened 2022-03-16
Add HeartbeatTimeout and ElectionTimeout to reloadable config.

ncabatoff:


api.go
```diff
@@ -696,6 +700,16 @@ func (r *Raft) ReloadConfig(rc ReloadableConfig) error {
 		return err
 	}
 	r.conf.Store(newCfg)
+
+	if rc.HeartbeatTimeout < oldCfg.HeartbeatTimeout {
+		// On leader, ensure replication loops running with a longer
+		// timeout than what we want now discover the change.
+		for _, repl := range r.leaderState.replState {
+			asyncNotifyCh(repl.notifyCh)
+		}
```
banks, line 709: Is it safe to access `leaderState` here outside the leader loop? I assume this read could potentially race with an update to that state from the leader loop couldn't it?

ncabatoff: Ah, good catch, probably not.  Let me see if I can re-work.

ncabatoff: Pushed a fix.

raft.go
```diff
@@ -190,6 +190,12 @@ func (r *Raft) runFollower() {
 		case b := <-r.bootstrapCh:
 			b.respond(r.liveBootstrap(b.configuration))
 
+		case <-r.leaderNotifyCh:
```
HridoyRoy, line 193: Why do we need to explicitly ignore a notification from this channel now?

HridoyRoy, line 193: Why do we need to explicitly ignore a notification from this channel now?

raft.go
```diff
@@ -819,6 +832,14 @@ func (r *Raft) leaderLoop() {
 			// Renew the lease timer
 			lease = time.After(checkInterval)
 
+		case <-r.leaderNotifyCh:
+			for _, repl := range r.leaderState.replState {
+				asyncNotifyCh(repl.notifyCh)
+			}
+
+		case <-r.followerNotifyCh:
```
HridoyRoy, line 840: Why do we need to explicitly ignore a notification from this channel now?

HridoyRoy (approved):
Lgtm!

raft.go
ncabatoff: I'm just being defensive, since the channels are buffered, and thinking about the case where we've transitioned from leader to follower and there might be a notification in the channel.  Probably not needed, but also doesn't hurt.

HridoyRoy: Definitely, makes sense!

raft.go
```diff
@@ -337,6 +344,12 @@ func (r *Raft) runCandidate() {
 		case b := <-r.bootstrapCh:
 			b.respond(ErrCantBootstrap)
 
+		case <-r.followerNotifyCh:
```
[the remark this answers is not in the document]
ncabatoff: Yes, fixed.

mkeeler (approved):
LGTM

Issue #508, opened 2022-06-11
How do I know that FSM.Apply has completed on a majority of nodes?


bootjp:
Thank you for the wonderful library.
I really like hashicorp's OSS because it's simple and robust.

By the way, I want to know that FSM.Apply has completed on the majority of nodes after FSM.Apply has been called.
Is there this way?

Specific cases where this might be necessary include.

Assume that immediately after FSM.Apply writes data, the leader node crashes and the leader is transferred to another no.
In this case, we assume that to avoid losing data written by FSM.Apply, we might wait for FSM.Apply to be called on a large number of nodes.

banks:
> Thank you for the wonderful library.
> I really like hashicorp's OSS because it's simple and robust.

Thank you ❤️ 

>  I want to know that FSM.Apply has completed on the majority of nodes after FSM.Apply has been called.
Is there this way?

There is no way to do exactly this in the library... but read on!

> Assume that immediately after FSM.Apply writes data, the leader node crashes and the leader is transferred to another no.
In this case, we assume that to avoid losing data written by FSM.Apply, we might wait for FSM.Apply to be called on a large number of nodes.

This is true - good catch. The way we deal with this though is not by reading from a quorum to check they have all applied. Instead the library provides two things:

 1. Thanks to the raft specification, a newly elected leader must have the most up to date logs - i.e. it has every comitted write in the cluster in _logs_ already (though not necessarily applied to it's FSM).
 2. We have a method called `Barrier` which is designed to be called by the leader: https://github.com/hashicorp/raft/blob/91745625f50efb7098f6a87a6c485cf54c435c1b/api.go#L822-L846

    Barrier will effectively push a non-op write through Raft right through to the FSM apply. Most users of this library should call this as soon as a new leader is established in the cluster, before that leader starts accepting new writes (or consistent reds) to ensure that not only does it have all the committed logs, but that they have actually been applied already. You can see this in Consul right here at the start of our leader loop: https://github.com/hashicorp/consul/blob/a02e9abcc17ec4cea09e8ee4dec300e36bd8ac7b/agent/consul/leader.go#L158-L165

Does that help?

bootjp:
I see.
Please allow me to continue asking questions because I still don't understand some of them.

1. New leaders always have up-to-date logs.
2. Raft Log is stored in stable storage when Barrier is called

[...]

PR #511, opened 2022-06-21
doc: add textual description of the raft.Apply process

huikang:


docs/apply.md
```diff
@@ -43,3 +44,45 @@ sequenceDiagram
...
+to the `applyCh` channel.
+
+2. If the node is not a leader, the method will return an error of `ErrNotLeader`. Otherwise,
+the main loop of the leader node calls `raft.dispatchLogs` to write the log entry locally.
+
+3. `raft.dispatchLogs` also sends a notification to the `f.triggerCh` of each follower (`map[ServerID]*followerReplication`) to start replicating log entries to the followers.
+
+4. For each follower, the leader has started a long running routine (`replicate`) to
+replicates log entries. On receiving a log entry to the `triggerCh`, the `replicate`
+routine makes the `Transport.AppendEntries` RPC call to do the replicatioin.
```
mkeeler, line 60:
```suggestion
routine makes the `Transport.AppendEntries` RPC call to do the replication.
```

docs/apply.md
```diff
@@ -43,3 +44,45 @@ sequenceDiagram
...
+
+10. The leader receives the notification on the  `r.leaderState.commitCh` channel and starts
+grouping the entries that can be applied to the fsm.
+
+11. `processLogs` applies all the committed entries that haven't been applied by batching the log entries and forwarding them through the `fsmMutateCh` channel to fsm.
+
+12. The actual place applying the commited log entries is in the main loop of `runFSM()`.
+
+13. After the log entries that contains the client req are applied to the fsm, the fsm
+module will set the reponses to the client request (`req.future.respond(nil)`).
```
mkeeler, line 88: Just to close the loop on the whole process it might be worth noting that the future returned by raft.Apply should now be unblocked and calls to `Error` or `Response` should return the data at this point.

docs/apply.md
```diff
@@ -43,3 +44,45 @@ sequenceDiagram
...
+the request. It appends any new entries to the local log store.
+
+6. In the same method on the follower as step 5, if the LeaderCommitIndex > this follower's
+commitIndex, start `processLogs` to send all the committed entries that haven't been applied
+to fsm (`fsmMutateCh <- batch`).
+
+7. The peer applies the commited entries to the FSM.
+
+8. If all went well, the follower responds success (`resp.Success = true`) to the 
+`appendEntries` RPC call.
```
mkeeler, line 72: I am not exactly sure where in this flow but do you think it would be good to point out the stepping nature of the processLogs within the append entries RPC.

One RPC will contain new logs entries but the LeaderCommitIndex will lag behind those log indexes. For example the following flow is likely

AppenEntries(Log: 1..5, LeaderCommitIndex: 0)
AppendEntries(Log: 6..8, LeaderCommitIndex: 4)
AppendEntries(Log: 9, LeaderCommitIndex: 8)
AppendEntries(Log: <none>, LeaderCommitIndex: 9)

Sometimes append entries rpcs wont contain new logs but only be bumping the commit index. Only if a server is far enough behind the majority of all servers will it ever see logs with a leader commit index which is greater or equal to the logs index within the rpc.

This subtlety is important because when the first AppendEntries RPC from my example above returns, the follower will not have queued up any applies to the FSM.

Due to this behavior its possible that a very small window of time exists when all followers have committed the log to disk, the write has been realized in the FSM of the leader but the followers have not yet applied the log to their FSM.

Paul's replies in this issue are relevant: https://github.com/hashicorp/raft/issues/508

huikang: @mkeeler , thanks for your suggestions. Totally agreed that it is worth to point out the details. Will send a new commit to update.

huikang: Hi, @mkeeler , could you please take another look at the updated PR when you have a chance? Thanks.

mkeeler (approved):

liyuan35023:
I deployed a consul cluster of 3 server nodes.I stoped all servers and start again, its possible that a very small window of time exists when consul returns stale data. The reason is that consul restores from the snapshot and needs to re-apply the committed but unapplied logs to fsm. In this case, the server may return stale data. 
I wonder if this is a bug or a scenario that could be allowed. And any advice if I want to avoid this situation?
Thanks. @mkeeler

mkeeler:
If you perform a consistent query in Consul it should ensure all old logs have been applied to the FSM before reading/returning the data. (`consistent=true` query param or you can make this the default

The way that is implemented is that the Raft leader (where the RPC for the query will be forwarded) will perform a Raft [`Barrier`](https://pkg.go.dev/github.com/hashicorp/raft#Raft.Barrier) operation. Calling that `Barrier` method on Raft will send a new log through Raft and only unblock once that log and all previous logs have been applied. This ensures that the data read will not be more stale than when the request was made.

PR #516, opened 2022-07-21
Improved election debug logs

benbuzbee:
When debugging elections in production you want to know what happened in
a particular term in order to identify what went wrong.

Add term= to all of the relevent debug logs

You also want to know which peers were contacted (their ID and their
address) and what the result from them was so you know why an election
was not won in a timely manner

Finally the log that just says "votes" in the message field seems
strange; improve the message.


ncabatoff (approved):

ncabatoff:
Thanks @benbuzbee, this looks helpful.

PR #519, opened 2022-08-19
file_snapshot: do defer after os.Open succeed

dbadoy:
os.Open always returns `nil` if an error occurs. 
So, it's seems better do defer after os.Open succeed.


ncabatoff (approved):

ncabatoff:
Thanks @dbadoy , nice catch!

PR #520, opened 2022-08-19
progress, state: use atomic method instead of mutex

dbadoy:
`atomic` performs better than mutex. So I think it's good to use `atomic`.

This is my sanity test codes not used goroutine for check the performance.
If there's many goroutines, the race condition may happens frequently, and the performance difference will be even greater. 
```go
package test

import (
	"sync"
	"sync/atomic"
	"testing"
)

type Test struct {
	data uint32
	mu   sync.Mutex
}

// Set
func BenchmarkMutexSet(b *testing.B) {
	t := Test{}
	for i := 0; i < b.N; i++ {
		t.mu.Lock()
		t.data = uint32(i)
		t.mu.Unlock()
	}
}

func BenchmarkAtomicSet(b *testing.B) {
	t := Test{}
	for i := 0; i < b.N; i++ {
		atomic.StoreUint32(&t.data, uint32(i))
	}
}

// Load
func BenchmarkMutexLoad(b *testing.B) {
	t := Test{
		data: 1,
	}
	for i := 0; i < b.N; i++ {
		t.mu.Lock()
		_ = t.data
		t.mu.Unlock()
	}
}

func BenchmarkAtomicLoad(b *testing.B) {
	t := Test{
		data: 1,
	}
	for i := 0; i < b.N; i++ {
		_ = atomic.LoadUint32(&t.data)
	}
}
```

```
# Set
BenchmarkMutexSet-10             1000000                26.86 ns/op
BenchmarkAtomicSet-10            1000000                 0.4546 ns/op
# Load
BenchmarkMutexLoad-10            1000000                18.83 ns/op
BenchmarkAtomicLoad-10           1000000                 0.6662 ns/op
```

### DIFF
--- a/progress.go
+++ b/progress.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	hclog "github.com/hashicorp/go-hclog"
@@ -94,23 +95,17 @@ type CountingReader interface {
 
 type countingReader struct {
 	reader io.Reader
-
-	mu    sync.Mutex
-	bytes int64
+	bytes  int64
 }
 
 func (r *countingReader) Read(p []byte) (n int, err error) {
 	n, err = r.reader.Read(p)
-	r.mu.Lock()
-	r.bytes += int64(n)
-	r.mu.Unlock()
+	atomic.StoreInt64(&r.bytes, atomic.AddInt64(&r.bytes, int64(n)))
 	return n, err
 }
 
 func (r *countingReader) Count() int64 {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-	return r.bytes
+	return atomic.LoadInt64(&r.bytes)
 }
 
 func newCountingReader(r io.Reader) *countingReader {
--- a/progress.go
+++ b/progress.go
@@ -100,7 +100,7 @@ type countingReader struct {
 
 func (r *countingReader) Read(p []byte) (n int, err error) {
 	n, err = r.reader.Read(p)
-	atomic.StoreInt64(&r.bytes, atomic.AddInt64(&r.bytes, int64(n)))
+	atomic.AddInt64(&r.bytes, int64(n))
 	return n, err
 }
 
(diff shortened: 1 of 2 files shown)


dhiaayachi:
Thank you for working on this and sharing those benchmark results.

In a use case as the one you benchmarked (where we need to access only 1 value atomically) atomics offer better performance then locks. That said, in the use case that we have, specifically for the term and the index, reading/writing both parameters need to happen atomically.
For example:

```golang
func (r *raftState) getLastSnapshot() (index, term uint64) {
	r.lastLock.Lock()
	index = r.lastSnapshotIndex
	term = r.lastSnapshotTerm
	r.lastLock.Unlock()
	return
}
```

is not equivalent to

```golang
func (r *raftState) getLastSnapshot() (uint64, uint64) {
	return atomic.LoadUint64(&r.lastSnapshotIndex), atomic.LoadUint64(&r.lastSnapshotTerm)
}
```

and would introduce a race condition where we end up with `lastSnapshotIndex` and `lastSnapshotTerm` diverging.

To achieve that using `atomics`, it would require creating a structure for both fields and perform a pointer swap/copy atomically. That become a more complex use case of atomics and do not necessarily offer the same performance gain (because it could prevent some other optimization from happening) and would require a specific benchmark for the use case.

Atomics can be more performant but are harder to reason about once you need to have consistent updates across more than on integer. This is pointed at by the Go authors in the [atomic package go docs](https://pkg.go.dev/sync/atomic):
>These functions require great care to be used correctly. Except for special, low-level applications, synchronization is better done with channels or the facilities of the sync package. Share memory by communicating; don’t communicate by sharing memory.

We prefer to not make performance optimizations prematurely when they increase code complexity and as far as we know the lock in question here has never been a performance bottleneck in this code since snapshots are not a frequent occurrence and is much simpler to reason about so we’d prefer to keep it.

Based on this I will close this PR, please feel free to reach out if you have questions.

dbadoy:
> and would introduce a race condition where we end up with lastSnapshotIndex and lastSnapshotTerm diverging.

Oops, you right. Thanks the kindly review!

PR #530, opened 2022-11-09
Raft pre-vote extension implementation

dhiaayachi:
This is an idea to implement pre-vote without breaking backward compatibility I'm exploring.

Pre-vote, based on the original raft thesis, is an extension to raft that help reduce term inflation and cluster instability when a node is partitioned from the cluster for a while and then is back online.

In normal election when a node have no leader, it transition into candidate state. In that state the node will try to run an election by incrementing its term and sending a vote-request to the other nodes in the cluster.

If the node win the election, the node become a leader and the cluster is stable again. But in the event that this node don't win the election, let's say because it's not connected to a quorum of nodes in the cluster, it will retry by incrementing it's term again and running another round of election. That could happen multiple times, which create a term inflation.

Once the node is connected back to quorum of nodes it's almost certain that its term will be higher than the other nodes in the cluster, as it was running rounds of elections non stop, while most likely the remaining of the cluster was stable. This will lead to the node making the current raft leader to step-down and another round of election is run. That's not a desired behaviour because the cluster was already stable and that node most likely don't have the latest log, so it won't be able to win the election anyway.

In this PR, a new `bool` called `preVote` is introduced to both the `RequestVoteRequest` and `RequestVoteResponse`. When a node transition to a candidate state and pre-vote is activated in the config, it will run a round of pre-election by sending RequestVoteRequest with `preVote` set to `true` and a term equal its current term + 1 but without incrementing its current term.

If the node that receive the request support pre-vote and have pre-vote activated, it will respond by sending a `RequestVoteResponse` with `preVote` set to true and grant its vote as if it's a normal election, but without incrementing its vote.

Otherwise the node will respond by sending a `RequestVoteResponse` with `preVote` set to false and grant its vote as it does for normal election, including incrementing its term.

The candidate node will count all the votes with `preVote` set to true, `grantedPrevotes` and all the votes with `preVote` set to false, `grantedVotes`.

If `grantedPrevotes` is bigger than the needed votes the candidate will run a normal election.

If `grantedVotes` is bigger than the needed votes the candidate will consider it self winning the election and transition to leader.

### DIFF
--- a/api.go
+++ b/api.go
@@ -210,6 +210,9 @@ type Raft struct {
 
 	// mainThreadSaturation measures the saturation of the main raft goroutine.
 	mainThreadSaturation *saturationMetric
+
+	// preVote control if the pre-vote feature is activated
+	preVote bool
 }
 
 // BootstrapCluster initializes a server's storage with the given cluster
@@ -557,6 +560,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		leaderNotifyCh:        make(chan struct{}, 1),
 		followerNotifyCh:      make(chan struct{}, 1),
 		mainThreadSaturation:  newSaturationMetric([]string{"raft", "thread", "main", "saturation"}, 1*time.Second),
+		preVote:               conf.PreVote,
 	}
 
 	r.conf.Store(*conf)
--- a/commands.go
+++ b/commands.go
@@ -89,6 +89,7 @@ type RequestVoteRequest struct {
 	// transfer. It is required for leadership transfer to work, because servers
 	// wouldn't vote otherwise if they are aware of an existing leader.
 	LeadershipTransfer bool
+	PreVote            bool
 }
 
 // GetRPCHeader - See WithRPCHeader.
@@ -110,6 +111,9 @@ type RequestVoteResponse struct {
 
 	// Is the vote granted.
 	Granted bool
+
+	// Is it a preVote response
+	PreVote bool
 }
 
 // GetRPCHeader - See WithRPCHeader.
--- a/config.go
+++ b/config.go
@@ -25,67 +25,74 @@ import (
 // The version details are complicated, but here's a summary of what's required
 // to get from a version 0 cluster to version 3:
 //
-// 1. In version N of your app that starts using the new Raft library with
-//    versioning, set ProtocolVersion to 1.
-// 2. Make version N+1 of your app require version N as a prerequisite (all
-//    servers must be upgraded). For version N+1 of your app set ProtocolVersion
-//    to 2.
-// 3. Similarly, make version N+2 of your app require version N+1 as a
-//    prerequisite. For version N+2 of your app, set ProtocolVersion to 3.
+//  1. In version N of your app that starts using the new Raft library with
+//     versioning, set ProtocolVersion to 1.
+//  2. Make version N+1 of your app require version N as a prerequisite (all
+//     servers must be upgraded). For version N+1 of your app set ProtocolVersion
+//     to 2.
+//  3. Similarly, make version N+2 of your app require version N+1 as a
+//     prerequisite. For version N+2 of your app, set ProtocolVersion to 3.
 //
 // During this upgrade, older cluster members will still have Server IDs equal
 // to their network addresses. To upgrade an older member and give it an ID, it
 // needs to leave the cluster and re-enter:
 //
-// 1. Remove the server from the cluster with RemoveServer, using its network
-//    address as its ServerID.
-// 2. Update the server's config to use a UUID or something else that is
-//	  not tied to the machine as the ServerID (restarting the server).
-// 3. Add the server back to the cluster with AddVoter, using its new ID.
+//  1. Remove the server from the cluster with RemoveServer, using its network
+//     address as its ServerID.
+//  2. Update the server's config to use a UUID or something else that is
+//     not tied to the machine as the ServerID (restarting the server).
+//  3. Add the server back to the cluster with AddVoter, using its new ID.
 //
 // You can do this during the rolling upgrade from N+1 to N+2 of your app, or
 // as a rolling change at any time after the upgrade.
 //
-// Version History
+// # Version History
 //
 // 0: Original Raft library before versioning was added. Servers running this
-//    version of the Raft library use AddPeerDeprecated/RemovePeerDeprecated
-//    for all configuration changes, and have no support for LogConfiguration.
+//
+//	version of the Raft library use AddPeerDeprecated/RemovePeerDeprecated
+//	for all configuration changes, and have no support for LogConfiguration.
+//
 // 1: First versioned protocol, used to interoperate with old servers, and begin
-//    the migration path to newer versions of the protocol. Under this version
-//    all configuration changes are propagated using the now-deprecated
-//    RemovePeerDeprecated Raft log entry. This means that server IDs are always
-//    set to be the same as the server addresses (since the old log entry type
-//    cannot transmit an ID), and only AddPeer/RemovePeer APIs are supported.
-//    Servers running this version of the protocol can understand the new
-//    LogConfiguration Raft log entry but will never generate one so they can
-//    remain compatible with version 0 Raft servers in the cluster.
+//
+//	the migration path to newer versions of the protocol. Under this version
+//	all configuration changes are propagated using the now-deprecated
+//	RemovePeerDeprecated Raft log entry. This means that server IDs are always
+//	set to be the same as the server addresses (since the old log entry type
+//	cannot transmit an ID), and only AddPeer/RemovePeer APIs are supported.
+//	Servers running this version of the protocol can understand the new
+//	LogConfiguration Raft log entry but will never generate one so they can
+//	remain compatible with version 0 Raft servers in the cluster.
+//
 // 2: Transitional protocol used when migrating an existing cluster to the new
-//    server ID system. Server IDs are still set to be the same as server
-//    addresses, but all configuration changes are propagated using the new
-//    LogConfiguration Raft log entry type, which can carry full ID information.
-//    This version supports the old AddPeer/RemovePeer APIs as well as the new
-//    ID-based AddVoter/RemoveServer APIs which should be used when adding
-//    version 3 servers to the cluster later. This version sheds all
-//    interoperability with version 0 servers, but can interoperate with newer
-//    Raft servers running with protocol version 1 since they can understand the
-//    new LogConfiguration Raft log entry, and this version can still understand
-//    their RemovePeerDeprecated Raft log entries. We need this protocol version
-//    as an intermediate step between 1 and 3 so that servers will propagate the
-//    ID information that will come from newly-added (or -rolled) servers using
-//    protocol version 3, but since they are still using their address-based IDs
-//    from the previous step they will still be able to track commitments and
-//    their own voting status properly. If we skipped this step, servers would
-//    be started with their new IDs, but they wouldn't see themselves in the old
-//    address-based configuration, so none of the servers would think they had a
-//    vote.
+//
+//	server ID system. Server IDs are still set to be the same as server
+//	addresses, but all configuration changes are propagated using the new
+//	LogConfiguration Raft log entry type, which can carry full ID information.
+//	This version supports the old AddPeer/RemovePeer APIs as well as the new
+//	ID-based AddVoter/RemoveServer APIs which should be used when adding
+//	version 3 servers to the cluster later. This version sheds all
+//	interoperability with version 0 servers, but can interoperate with newer
+//	Raft servers running with protocol version 1 since they can understand the
+//	new LogConfiguration Raft log entry, and this version can still understand
+//	their RemovePeerDeprecated Raft log entries. We need this protocol version
+//	as an intermediate step between 1 and 3 so that servers will propagate the
+//	ID information that will come from newly-added (or -rolled) servers using
+//	protocol version 3, but since they are still using their address-based IDs
+//	from the previous step they will still be able to track commitments and
+//	their own voting status properly. If we skipped this step, servers would
+//	be started with their new IDs, but they wouldn't see themselves in the old
+//	address-based configuration, so none of the servers would think they had a
+//	vote.
+//
 // 3: Protocol adding full support for server IDs and new ID-based server APIs
-//    (AddVoter, AddNonvoter, etc.), old AddPeer/RemovePeer APIs are no longer
-//    supported. Version 2 servers should be swapped out by removing them from
-//    the cluster one-by-one and re-adding them with updated configuration for
-//    this protocol version, along with their server ID. The remove/add cycle
-//    is required to populate their server ID. Note that removing must be done
-//    by ID, which will be the old server's address.
+//
+//	(AddVoter, AddNonvoter, etc.), old AddPeer/RemovePeer APIs are no longer
+//	supported. Version 2 servers should be swapped out by removing them from
+//	the cluster one-by-one and re-adding them with updated configuration for
+//	this protocol version, along with their server ID. The remove/add cycle
+//	is required to populate their server ID. Note that removing must be done
+//	by ID, which will be the old server's address.
 type ProtocolVersion int
 
 const (
@@ -99,19 +106,22 @@ const (
 // Currently, it is always assumed that the server generates the latest version,
 // though this may be changed in the future to include a configurable version.
 //
-// Version History
+// # Version History
 //
 // 0: Original Raft library before versioning was added. The peers portion of
-//    these snapshots is encoded in the legacy format which requires decodePeers
-//    to parse. This version of snapshots should only be produced by the
-//    unversioned Raft library.
+//
+//	these snapshots is encoded in the legacy format which requires decodePeers
+//	to parse. This version of snapshots should only be produced by the
+//	unversioned Raft library.
+//
 // 1: New format which adds support for a full configuration structure and its
-//    associated log index, with support for server IDs and non-voting server
-//    modes. To ease upgrades, this also includes the legacy peers structure but
-//    that will never be used by servers that understand version 1 snapshots.
-//    Since the original Raft library didn't enforce any versioning, we must
-//    include the legacy peers structure for this version, but we can deprecate
-//    it in the next snapshot version.
+//
+//	associated log index, with support for server IDs and non-voting server
+//	modes. To ease upgrades, this also includes the legacy peers structure but
+//	that will never be used by servers that understand version 1 snapshots.
+//	Since the original Raft library didn't enforce any versioning, we must
+//	include the legacy peers structure for this version, but we can deprecate
+//	it in the next snapshot version.
 type SnapshotVersion int
 
 const (
@@ -221,6 +231,9 @@ type Config struct {
 
 	// skipStartup allows NewRaft() to bypass all background work goroutines
 	skipStartup bool
+
+	// PreVote activate the pre-vote feature
+	PreVote bool
 }
 
 func (conf *Config) getOrCreateLogger() hclog.Logger {
--- a/raft.go
+++ b/raft.go
@@ -284,7 +284,13 @@ func (r *Raft) runCandidate() {
 	metrics.IncrCounter([]string{"raft", "state", "candidate"}, 1)
 
 	// Start vote for us, and set a timeout
-	prevoteCh := r.electSelf(true)
+	var voteCh <-chan *voteResult
+	var prevoteCh <-chan *voteResult
+	if r.preVote {
+		prevoteCh = r.electSelf(true)
+	} else {
+		voteCh = r.electSelf(false)
+	}
 
 	// Make sure the leadership transfer flag is reset after each run. Having this
 	// flag will set the field LeadershipTransfer in a RequestVoteRequst to true,
@@ -301,7 +307,7 @@ func (r *Raft) runCandidate() {
 	preVoteGrantedVotes := 0
 	votesNeeded := r.quorumSize()
 	r.logger.Debug("calculated votes needed", "needed", votesNeeded, "term", term)
-	var voteCh <-chan *voteResult
+
 	for r.getState() == Candidate {
 		r.mainThreadSaturation.sleeping()
 
@@ -1565,13 +1571,16 @@ func (r *Raft) processConfigurationLogEntry(entry *Log) error {
 func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 	defer metrics.MeasureSince([]string{"raft", "rpc", "requestVote"}, time.Now())
 	r.observe(*req)
-
+	preVote := req.PreVote
+	if !r.preVote {
+		preVote = false
+	}
 	// Setup a response
 	resp := &RequestVoteResponse{
 		RPCHeader: r.getRPCHeader(),
 		Term:      r.getCurrentTerm(),
 		Granted:   false,
-		PreVote:   req.PreVote,
+		PreVote:   preVote,
 	}
 	var rpcErr error
 	defer func() {
@@ -1627,7 +1636,7 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 	if req.Term > r.getCurrentTerm() {
 		// Ensure transition to follower
 		r.logger.Debug("lost leadership because received a requestVote with a newer term")
-		if !req.PreVote {
+		if !preVote {
 			r.setState(Follower)
 			r.setCurrentTerm(req.Term)
 		}
@@ -1687,7 +1696,7 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
 	}
 
 	// Persist a vote for safety
-	if !req.PreVote {
+	if !preVote {
 		if err := r.persistVote(req.Term, candidateBytes); err != nil {
 			r.logger.Error("failed to persist vote", "error", err)
 			return
--- a/raft.go
+++ b/raft.go
@@ -284,7 +284,7 @@ func (r *Raft) runCandidate() {
 	metrics.IncrCounter([]string{"raft", "state", "candidate"}, 1)
 
 	// Start vote for us, and set a timeout
-	voteCh := r.electSelf()
+	prevoteCh := r.electSelf(true)
 
 	// Make sure the leadership transfer flag is reset after each run. Having this
 	// flag will set the field LeadershipTransfer in a RequestVoteRequst to true,
@@ -298,9 +298,10 @@ func (r *Raft) runCandidate() {
 
 	// Tally the votes, need a simple majority
 	grantedVotes := 0
+	preVoteGrantedVotes := 0
 	votesNeeded := r.quorumSize()
 	r.logger.Debug("calculated votes needed", "needed", votesNeeded, "term", term)
-
+	var voteCh <-chan *voteResult
 	for r.getState() == Candidate {
 		r.mainThreadSaturation.sleeping()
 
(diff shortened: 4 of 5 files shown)


raft.go
```diff
@@ -308,7 +315,54 @@ func (r *Raft) runCandidate() {
...
+
+			// Check if we've become the leader
+			// If we won the election because the majority of nodes don't support pre-vote (or pre-vote is deactivated)
+			// set our state to leader and our term to the pre-vote term.
+			if grantedVotes >= votesNeeded {
+				r.logger.Info("election won", "term", vote.Term, "tally", grantedVotes)
+				r.setState(Leader)
+				r.setLeader(r.localAddr, r.localID)
 
+				r.setCurrentTerm(term)
```
banks, line 356: This is the part that I think we should probably also `persistVote` for ourselves even though in this code you didn't count our own preVote in the quorum. I don't think this is necessarily incorrect as it is, but I wonder if we'll violate assumptions later that we are leader in a term but we never voted in that term? I think it's safe in all cases to decide to "upgrade" our own preVote to a real vote here anyway and avoids ever needing to think about that case.

Technically if we keep this approach we could get to this point even faster by checking for `grantedVotes >= (votesNeeded-1)` since we can always upgrade our own vote to make up the quorum if enough other peers voted for real.

raft.go
```diff
@@ -308,7 +315,54 @@ func (r *Raft) runCandidate() {
...
+				return
+			}
+			// Check if we've won the pre-vote and proceed to election if so
+			if preVoteGrantedVotes >= votesNeeded {
+				r.logger.Info("pre election won", "term", vote.Term, "tally", preVoteGrantedVotes)
+				preVoteGrantedVotes = 0
+				grantedVotes = 0
```
banks, line 363: Do we need to reset this? If some of the peers already voted "for real" then we already have their vote as far as they are concerned. If we really wanted to, we could modify `electSelf` to skip sending a requests to any nodes that already granted a "real" vote! (Not saying we should though, just an observation). As it is this would trigger our "info" log about duplicate vote requests on those old nodes, but that's not a huge deal.

banks: Actually I came up with a scenario where the cluster could get stuck with one node running PreVote campaigns that never get enough support to complete but prevent other nodes getting elected despite a quorum of healthy nodes with this logic. If we change it to only need `quorum - 1` and upgrade our own vote then those cases can't happen...

raft.go
```diff
@@ -308,7 +315,54 @@ func (r *Raft) runCandidate() {
...
+				r.setCurrentTerm(vote.Term)
+				return
+			}
+			if vote.Term > term && vote.PreVote {
+				r.logger.Debug("newer term discovered on pre-vote, fallback to follower", "term", vote.Term)
+				r.setState(Follower)
+				r.setCurrentTerm(vote.Term)
+				return
+			}
+
```
banks, line 336: Do we need different logic for PreVote as well as this?

For example if the peer we sent it to denied us despite a higher term because our logs are behind, or because there is already a stable leader, we should treat that the same way and switch to Follower to try and wait for the leader to catch us up rather than just running pre vote over and over.

Actually I feel like that's not special to PreVote 🤔. Maybe we should do that for real elections too when we are rejected without a higher election term being present 🤔 .

Ideally, as a candidate, especially one having prevotes rejected, we need a path to transition back to follower so we can continue without disrupting the cluster. The only way out of candidate state right now I think is to either win an election in a new term ourselves, or observe another election happeneing - either way we can't become a healthy follower for the stable cluster leader without the cluster being disrupted by an election unless there is some other path to follower that _doesn't_ involve observing a term higher that our prevote one!

raft.go
```diff
@@ -308,7 +315,54 @@ func (r *Raft) runCandidate() {
...
 
+				r.setCurrentTerm(term)
+				return
+			}
+			// Check if we've won the pre-vote and proceed to election if so
+			if preVoteGrantedVotes >= votesNeeded {
+				r.logger.Info("pre election won", "term", vote.Term, "tally", preVoteGrantedVotes)
+				preVoteGrantedVotes = 0
+				grantedVotes = 0
+				voteCh = r.electSelf(false)
```
banks, line 364: Do we also need to reset the electionTimer? I'm not sure. If we don't it means that we have to get PreVote _and_ Vote rounds within one election timeout. If we are close to timing out when we get PreVotes through then it seems wasteful to start a real election that's really likely to fail due to timing out before responses are gathered and then restart prevoting all over again!

raft.go
```diff
@@ -284,7 +284,13 @@ func (r *Raft) runCandidate() {
...
 
 	// Start vote for us, and set a timeout
-	voteCh := r.electSelf()
+	var voteCh <-chan *voteResult
+	var prevoteCh <-chan *voteResult
+	if r.preVote {
+		prevoteCh = r.electSelf(true)
+	} else {
+		voteCh = r.electSelf(false)
+	}
```
banks, line 293: At first I felt like we need to somehow keep the state around so that if we timeout a PreVote campaign we fall back to a real one next time... but then I realise that would defeat the purpose of a prevote because a partitioned server would only prevote for one round before it resumed incrementing it's term...

I do wonder a bit about whether it's possible to get "stuck" never quite winning a pre vote in time but also not making progress.

Here's the scenario I came up with.

 * Three servers A, B, C
 * Only A is upgraded to have PreVote
 * B is currently leader
 * C is down for a bit
 * A disconnects temporarily
 * B looses leadership and becomes a follower waiting for an election (or to start one)
 * A starts election first and use PreVote
 * B treats it like RealVote and so grants it an resets its contact time (which prevents it from timing out and becoming a candidate
 * A sees the request granted by B but it's not enough for a quorum (using the code here) so keeps waiting
 * A's prevote campaign times out and it starts again... electionTimeout is always less than or equal to heartbeat timeout so this could continue forever even though there are a quorum of healthy servers.

Note that this scenario is "fixed" if we use the suggestion I made below about only requiring `quorum - 1` real vote grants and then upgrading our own. But it's an example of something non-obvious that could cause the cluster to get stuck in a whole new way that isn't possible today which makes me 🤔.

banks (commented):
A whole bunch of thoughts for now.

There is a lot here - mostly because I'm trying to reason about it for myself rather than because this is necessarily a bad approach. I could also be wrong in any of my thought processes so have a think and we can maybe discuss some more which overall approach we think will be simplest to reason about and implement correctly!

raft.go
banks: Actually I came up with [a scenario](https://github.com/hashicorp/raft/pull/530#discussion_r1019159636) where the cluster could get stuck with one node running PreVote campaigns that never get enough support to complete but prevent other nodes getting elected despite a quorum of healthy nodes with this logic. If we change it to only need quorum - 1 and upgrade our own vote then those cases can't happen...

raft.go
dhiaayachi: What about B in that case, at some point B will start an election too and because B don't have prevote A will vote for B and we will be out of the loop. 
The idea, is that:
- if we have majority of nodes with prevote, pre vote will be guaranteed to go through
- if we have majority of nodes without prevote, the prevote will become a vote
- if we have same number, the prevote will continue until: a) we have another node that support prevote added, b) a node that don't support prevote take over
I think that the quorum - 1 idea would work but it will be an optimization.

raft.go
dhiaayachi: yes that's probably needed and explain some of the timing issue I was having in tests.

raft.go
dhiaayachi: I don't think this is needed in the election code, if we are part of the cluster already. The leader will eventually send an AppendEntry that will transition us to followers. Doing so in here have the potential, for example, of making a node not part of the cluster think it's a follower.

raft.go
dhiaayachi: You just solved a bug I was chasing with one of the tests, when I set the number of nodes to 2 prevotes,1 no prevote 😄

dhiaayachi:
I think that the different RPC idea is simpler, the main reason I steered away from it for the following reasons:

- pre-vote need to always stay aligned with real vote (except for state persistence) for it to work but I guess we can make sure that underneath both RPCs call the same code.
- from optimization perspective the current solution will stabilize in 1 round in mixed server scenario while the different RPC will need 2 rounds

That said, it's easier to reason about special cases like the bug you just found out because we avoid changing the state of the no pre-vote nodes while doing pre-vote.

raft.go
dhiaayachi: > If we really wanted to, we could modify electSelf to skip sending a requests to any nodes that already granted a "real" vote! (Not saying we should though, just an observation). 

I think it's safer if we don't, at least from code complexity perspective. Also I think we should recommend to only have mixed clusters as a transient state, so I don't think the logs are a big issue in that case.

Also about not resetting this, if we don't implement skipping the nodes that already voted, it's incorrect as there is a possibility we have the same node vote double.

dhiaayachi:
Thinking more about implementing this using a different RPC command, there is few use cases here, but the most interesting use case is when some nodes support prevote, some don't and a minority of nodes is down.

In that case if neither the nodes which support prevote or the nodes which don't can form a quorum we will be stuck in pre-election forever. One way of solving this would be to say if we get enough answers (error + granted-prevote) to form a quorum we transition to full election.

The issue with that solution is that it create another gap where, we have this condition true but without receiving all the nodes answer which can make the transition to full election premature.

I don't think this is a deal breaker as with the solution in here we will be able to progress even if the pre-vote is not optimal, but it show that the different RPC alternative will have a similar complexity.

raft.go
dhiaayachi: While going through this use case to fix it I think it's part of a more general use case. Where `grantedVotes + grantedPreVotes >= quorum` but `grantedVotes < quorum` and `grantedPreVotes < quorum`.
I think the best way to solve this would be to transition to election as soon as `grantedVotes + grantedPreVotes >= quorum`.

mkeeler (commented):
Overall I think the pre-vote implementation looks fine. However testing seems to focus on the happy path where not enabling pre-vote would have worked without many subsequent leader elections.

I think before merging I would want to see some test either in the raft-compat package or the unit tests which forces pre-voting to kick in to prevent tons of leader elections.

raft-compat/prevote_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/hashicorp/raft"
+	raftprevious "github.com/hashicorp/raft-previous-version"
+	"github.com/hashicorp/raft/compat/testcluster"
+	"github.com/hashicorp/raft/compat/utils"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestRaft_PreVote_BootStrap_PreVote(t *testing.T) {
```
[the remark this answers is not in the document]
dhiaayachi: I added the node removal in the leave hook. Also those tests were focusing on the upgrade scenario as that's the biggest risk we identified. I added a new test that cover the "term inflation" scenario to verify that the term don't inflate when a node is partitioned in a while and then comeback to the cluster.

raft_test.go
```diff
@@ -2061,6 +2061,95 @@ func TestRaft_AppendEntry(t *testing.T) {
...
+			c.Partition([]ServerAddress{leaderOld.localAddr})
+			time.Sleep(c.propagateTimeout * 3)
+			leader := c.Leader()
+			require.NotEqual(t, leader.leaderID, leaderOld.leaderID)
+		})
+	}
+
+}
+
+func TestRaft_PreVoteAvoidElectionWithPartition(t *testing.T) {
```
dhiaayachi, line 2122: @mkeeler this is the test that verify that prevote behave as untended.

raft_test.go
```diff
@@ -2061,6 +2061,95 @@ func TestRaft_AppendEntry(t *testing.T) {
...
+	// Make a prevote cluster.
+	conf := inmemConfig(t)
+	conf.PreVote = true
```
dhiaayachi, line 2125: If this is set to false the test will fail

raft_test.go
```diff
@@ -2061,6 +2061,64 @@ func TestRaft_AppendEntry(t *testing.T) {
 	require.True(t, resp2.Success)
 }
 
+func TestRaft_PreVoteMixedCluster(t *testing.T) {
```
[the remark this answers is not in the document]
dhiaayachi: The focus in this test is to verify that leader election still work when we have multiple mixes of pre-vote/no-pre-vote servers. So it should exercise pre-vote is some cases but not all the time. The most important here, is that leader election still happen and succeed in all those scenario, regardless of if pre-vote was used or not.

raft.go
```diff
@@ -1680,6 +1732,102 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
+		return
+	}
+
+	if lastTerm == req.LastLogTerm && lastIdx > req.LastLogIndex {
+		r.logger.Warn("rejecting vote request since our last index is greater",
+			"candidate", candidate,
+			"last-index", lastIdx,
+			"last-candidate-index", req.LastLogIndex)
+		return
+	}
```
mkeeler, line 1825: Similar to the previous comment, these returns will inadvertently send a response that grants the vote instead of rejecting it.

raft_test.go
```diff
@@ -2061,6 +2061,95 @@ func TestRaft_AppendEntry(t *testing.T) {
...
+	// Check the leader is stable and the followers are as expected
+	leaderTerm := c.Leader().getCurrentTerm()
+	require.Equal(t, leaderTerm, oldLeaderTerm)
+	require.Len(t, c.WaitForFollowers(3), 3)
+
+	// reconnect the partitioned node
+	c.FullyConnect()
+	time.Sleep(3 * c.propagateTimeout)
+
+	// Check that the number of followers increase and
```
mkeeler, line 2146: I think the rest of this comment would be useful to readers even though I think I can piece together what you would want to convey.

testing.go
```diff
@@ -21,13 +21,13 @@ import (
 var userSnapshotErrorsOnNoData = true
 
 // Return configurations optimized for in-memory
-func inmemConfig(tb testing.TB) *Config {
+func inmemConfig(t *testing.T) *Config {
```
mkeeler, line 24: Any particular reason for this change?

mkeeler (changes requested):

raft-compat/prevote_test.go
```diff
@@ -0,0 +1,294 @@
...
+	leaveTransfer := func(t *testing.T, cluster testcluster.RaftCluster, id string) {
+		if cluster.GetLeader().GetLocalID() == id {
+			transfer := cluster.Raft(id).(*raft.Raft).LeadershipTransfer()
+			utils.WaitFuture(t, transfer)
+		}
+		f := cluster.Raft(id).(*raft.Raft).Shutdown()
+		utils.WaitFuture(t, f)
+	}
+	leaveNoTransfer := func(t *testing.T, cluster testcluster.RaftCluster, id string) {
+		fr := cluster.GetLeader().GetRaft().(*raft.Raft).RemoveServer(raft.ServerID(id), 0, 0)
```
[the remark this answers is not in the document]
dhiaayachi: I'm not sure I understand 🤔. This will trigger prevote, specifically when we get to upgrade the leader but it won't test term inflation. Prevote happen in all leader election scenarios before a real vote, so in this case, when we upgrade the leader node (the last node to be upgraded), a leader election will happen and it will involve a pre-vote.

raft.go
```diff
@@ -1680,6 +1732,102 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
+	defer func() {
+		rpc.Respond(resp, rpcErr)
+	}()
+
+	// Version 0 servers will panic unless the peers is present. It's only
+	// used on them to produce a warning message.
+	if r.protocolVersion < 2 {
+		resp.Peers = encodePeers(r.configurations.latest, r.trans)
+	}
+
```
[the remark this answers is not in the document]
dhiaayachi: This is unlikely to happen, but I think for consistency with the vote request it would make sense to have the same behaviour.

raft.go
```diff
@@ -1680,6 +1732,102 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
+		resp.Term = req.Term
+		resp.Granted = true
+	}
+
+	// if we get a request for vote from a nonVoter  and the request term is higher,
+	// step down and update term, but reject the vote request
+	// This could happen when a node, previously voter, is converted to non-voter
+	// The reason we need to step in is to permit to the cluster to make progress in such a scenario
+	// More details about that in https://github.com/hashicorp/raft/pull/526
+	if len(req.ID) > 0 {
```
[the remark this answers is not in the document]
dhiaayachi: I'm not sure 🤔, this will happen when a node think of it self as a voter but it's not from the perspective of the other nodes. If we remove this check, the prevote will be granted and it will proceed to a vote which will fail. I think it's better if we shut it down from the prevote phase.

raft.go
```diff
@@ -1680,6 +1732,102 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
+		candidateID := ServerID(req.ID)
+		if len(r.configurations.latest.Servers) > 0 && !hasVote(r.configurations.latest, candidateID) {
+			r.logger.Warn("rejecting pre-vote request since node is not a voter", "from", candidate)
+			return
```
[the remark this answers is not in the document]
dhiaayachi: I agree on the fact that it would be easier to follow if refactored, but I think it's better if we do this in a second phase and refactor both vote and prevote. I would like at least for the time being to keep pre-vote similar to vote.

raft.go
dhiaayachi: Fixed this by avoiding to set granted to true above and keep processing further.

commands.go
```diff
@@ -113,13 +114,63 @@ type RequestVoteResponse struct {
...
+	Candidate []byte
+
+	// Used to ensure safety
+	LastLogIndex uint64
+	LastLogTerm  uint64
+
+	// Used to indicate to peers if this vote was triggered by a leadership
+	// transfer. It is required for leadership transfer to work, because servers
+	// wouldn't vote otherwise if they are aware of an existing leader.
+	LeadershipTransfer bool
```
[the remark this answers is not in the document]
dhiaayachi: yes it does. I traced leadership transfer and it work as follow:
- When a node receive a leadership transfer command, it sends a TimeoutNow RPC to the chosen new leader and step down.
- Upon receiving the TimeoutNow RPC, the chosen leader will set it's state to leadershipTransfer and transition to Candidate

From this point normal leadership election will happen (including pre-vote) the only exception is that we send the LeadershipTransfer as part of the vote (and pre-vote) to avoid not granting the vote/pre-vote when the voting node think it still have a leader.

I'm mitigated about if we should keep it as is or completely remove pre-vote from the equation when performing leadership transfer I think it won't change anything about the behaviour in leadership transfer as in all cases we will end up in leader-election.

I would be interested in hearing your thoughts and @banks about this.

banks (commented):
Amazing Dhia.

I'm much happier with this approach overall and the tests with mixed versions and upgrades are great. I have a bunch of things inline which are mostly minor but probably worthwhile.

The biggest thing I think is the leadership transfer which seems to me like it really shouldn't ever involve a pre-vote but maybe I'm not thinking about it right!

config.go
```diff
@@ -234,6 +234,9 @@ type Config struct {
 
 	// skipStartup allows NewRaft() to bypass all background work goroutines
 	skipStartup bool
+
+	// PreVote activate the pre-vote feature
+	PreVote bool
```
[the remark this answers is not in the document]
dhiaayachi: I agree about making this an opt-out flag, I'm not sure about how useful will it be to be a dynamic flag. Especially considering the impact related to toggling it in the middle of operations or while the cluster is unstable 🤔

commands.go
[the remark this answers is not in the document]
dhiaayachi: > Leaders only try to transfer to nodes that are healthy and connected anyway (should check that) so by definition it won't have an elevated term

This is not necessarily true as we allow providing the server we would like to transfer leadership to and it's a best effort transfer, so if the server we specify is not able to become a leader the transfer won't happen.

That said, I agree on the fact that leadership transfer is probably not something that we will benefit from optimizing the election for and removing it from the pre-vote flow remove some of the complexity.

dhiaayachi:
@banks @mkeeler I updated based on all the review, I think this is getting close now 😅

raft-compat/prevote_test.go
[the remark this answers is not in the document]
mkeeler: Right it will trigger the pre-vote when the leader gets rolled but not when followers are being shut down.

commands.go
```diff
@@ -120,6 +120,43 @@ func (r *RequestVoteResponse) GetRPCHeader() RPCHeader {
...
+// RequestPreVoteRequest is the command used by a candidate to ask a Raft peer
+// for a vote in an election.
+type RequestPreVoteRequest struct {
+	RPCHeader
+
+	// Provide the term and our id
+	Term uint64
+
+	// Deprecated: use RPCHeader.Addr instead
+	Candidate []byte
```
mkeeler, line 132:
```suggestion
```

Could we remove this field. Seeing as its been deprecated for the regular RequestVoteRequest it seems reasonable for pre-vote to just never support this way of specifying the candidate server. Also in processing pre vote requests we never use this field.

raft.go
```diff
@@ -1907,6 +2044,84 @@ func (r *Raft) electSelf() <-chan *voteResult {
...
+	// Increment the term
+	newTerm := r.getCurrentTerm() + 1
+
+	// Construct the request
+	lastIdx, lastTerm := r.getLastEntry()
+	req := &RequestPreVoteRequest{
+		RPCHeader: r.getRPCHeader(),
+		Term:      newTerm,
+		// this is needed for retro compatibility, before RPCHeader.Addr was added
+		Candidate:    r.trans.EncodePeer(r.localID, r.localAddr),
```
mkeeler, line 2065:
```suggestion
```

This change goes in tandem with my other proposal to remove the Candidate field from the struct definition.

mkeeler (changes requested):
Two last minor requests to remove the deprecated Candidate field from the RequestPreVoteRequest

api.go
```diff
@@ -560,6 +565,7 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 		leaderNotifyCh:        make(chan struct{}, 1),
 		followerNotifyCh:      make(chan struct{}, 1),
 		mainThreadSaturation:  newSaturationMetric([]string{"raft", "thread", "main", "saturation"}, 1*time.Second),
+		preVoteDisabled:       conf.PreVoteDisabled || !transportSupportPreVote,
```
banks, line 568: perhaps we should also log a warning here? Something like `WARN: raft: pre-vote is disabled because it is not supported by the Transport`.

raft.go
```diff
@@ -310,7 +323,41 @@ func (r *Raft) runCandidate() {
...
+				preVoteGrantedVotes++
+				r.logger.Debug("pre-vote granted", "from", preVote.voterID, "term", preVote.Term, "tally", preVoteGrantedVotes)
+			} else {
+				preVoteRefusedVotes++
+				r.logger.Debug("pre-vote denied", "from", preVote.voterID, "term", preVote.Term, "tally", preVoteGrantedVotes)
+			}
+
+			// Check if we've won the pre-vote and proceed to election if so
+			if preVoteGrantedVotes >= votesNeeded {
+				r.logger.Info("pre-vote successful, starting election", "term", preVote.Term, "tally", preVoteGrantedVotes, "votesNeeded", votesNeeded-1)
```
banks, line 349: 🤔 I'm trying to work out why we `-1` the votesNeeded here. I think it's because we auto-vote for ourselves in `votesNeeded` and so it might be confusing to operators whether `votesNeeded` includes that self-vote or not.

I.e. in a cluster of 3 `votesNeeded` would actually be 2, but we effectively only need one vote from another server in addition to our own so you've chosen to show it as 1 instead here.

I initially thought this was potentially more surprising because for those who know about consensus, they'd maybe expect this number to match the quorum number of the cluster. I changed my mind when I considered the UX and assumed we'd only see "vote granted" logs for the remote peers so having this number correlate to that might be more intuitive. But then I went and looked at `preElectSelf` again and we do Debug log "voting for self".

I don't think ultimately it's a big deal but on balance it seems slightly less confusing to not subtract 1 here both to match any intuitions about quorum size, and to remain consistent with debug log output where we make explicit that we vote for ourselves too?

raft.go
```diff
@@ -1680,6 +1730,85 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
+		Granted:   false,
+	}
+	var rpcErr error
+	defer func() {
+		rpc.Respond(resp, rpcErr)
+	}()
+
+	// Check if we have an existing leader [who's not the candidate] and also
+	// check the LeadershipTransfer flag is set. Usually votes are rejected if
+	// there is a known leader. But if the leader initiated a leadership transfer,
```
banks, line 1751: I think the comments about LeadershipTransfer are out of sync with the code now right?

raft.go
```diff
@@ -1907,6 +2044,82 @@ func (r *Raft) electSelf() <-chan *voteResult {
...
+		RPCHeader:    r.getRPCHeader(),
+		Term:         newTerm,
+		LastLogIndex: lastIdx,
+		LastLogTerm:  lastTerm,
+	}
+
+	// Construct a function to ask for a vote
+	askPeer := func(peer Server) {
+		r.goFunc(func() {
+			defer metrics.MeasureSince([]string{"raft", "candidate", "electSelf"}, time.Now())
```
banks, line 2071: Should we make this a separate metric `preElectSelf`?

raft.go
```diff
@@ -1907,6 +2044,82 @@ func (r *Raft) electSelf() <-chan *voteResult {
...
+			resp := &preVoteResult{voterID: peer.ID}
+			if prevoteTrans, ok := r.trans.(WithPreVote); ok {
```
banks, line 2073: nit: would it be cleaner to do this type assertion right at the top of `preElectSelf` and bail out immediately if the transport doesn't support it (which _shouldn't_ be possible due to the check in NewRaft but defensive against future change...).

Main rationale is that it allows un-indenting and simplifying the body of this callback which makes it easier to follow the important parts!

banks (approved):
Amazing @dhiaayachi. 🎉 

One minor thing. The PR description still says things like:

> In this PR, a new bool called preVote is introduced to both the RequestVoteRequest and RequestVoteResponse. When a node transition to a candidate state and pre-vote is activated in the config, it will run a round of pre-election by sending RequestVoteRequest with preVote set to true and a term equal its current term + 1 but without incrementing its current term.

Which was the original implementation we moved away from. Since this is a significant change, what do you think about editing the PR description with the same great detail but about the actual implementation that merges (and why we took this path vs the original one)? We don't have an RFC for this but the detail we've thought through in the PR and offline would be great to capture somewhere coherent for future reference and I think this PR is the most natural home for that?

Overall I think this is ready to ship. There are a couple of minor suggestions inline but all are non-blocking potential code clarity improvements that I'll leave you to choose if they are worthwhile!

raft.go
dhiaayachi: At this level our own vote is completely transparent to us. It come through the `preVoteCh` like any other vote. see: https://github.com/hashicorp/raft/blob/1912201252579c050610c4a93ccfc05356647872/raft.go#L2105-L2105

raft.go
dhiaayachi: I changed that and added a defensive panic. Please check if this is what you have in mind.

dhiaayachi:
Thank you @banks!
I updated based on your comments, I also rewrote the PR description based on the current implementation. Can you please give it another 👀 ?

raft.go
banks: huh... then that makes me wonder even more why we have `votesNeeded-1` in the log here?

raft.go
```diff
@@ -1680,6 +1730,82 @@ func (r *Raft) requestVote(rpc RPC, req *RequestVoteRequest) {
...
+	if lastTerm == req.LastLogTerm && lastIdx > req.LastLogIndex {
+		r.logger.Warn("rejecting pre-vote request since our last index is greater",
+			"candidate", candidate,
+			"last-index", lastIdx,
+			"last-candidate-index", req.LastLogIndex)
+		return
+	}
+
+	resp.Granted = true
+	r.setLastContact()
```
banks, line 1806: Hmmm I didn't spot this before... but should we be setting this? I _think_ `lastContact` is supposed to indicate when we last heard from the _leader_ of the cluster. (Yep, I checked the doc comments for it) In a real vote, it seems reasonable to expect that the server we just gave our vote to is about to become the leader. In this case, that's not quite as clear (thought still likely I suppose).

Did you consider whether it's more like to be better or worse to set this? I think I'd lean towards not calling this because it changes state that at least some other part of raft relies on and so makes it just a tiny bit more risky than a pre-vote campaign could cause new behaviour in the cluster. I think semantically, the node that spoke to us is _not_ the leader right now either so it seems like we shouldn't treat it like it's already won an election. I guess that logic also applies to a real election though.

I don't think it is likely to be a problem either way. Just curious to think it through and see if we can imagine any impact or increased risk of impact either way. 🤔 

OK so it is used in a follower when heartbeat timeout is called when in follower state: https://github.com/hashicorp/raft/blob/cc2bb080d64f172ff2fcad13289e7ec07dd2e6c6/raft.go#L219

I think if we get here (granting the pre vote) then this node must already not have a leader (i.e. be in candidate state)... So I don't think it really is a _problem_, but I'm still not sure it makes sense to pretend we "heard from a leader" based on a prevote?

raft.go
```diff
@@ -1907,6 +2041,90 @@ func (r *Raft) electSelf() <-chan *voteResult {
...
+	// but check just in case
+	prevoteTrans, prevoteTransSupported := r.trans.(WithPreVote)
+	if !prevoteTransSupported {
+		panic("preElection is not possible if the transport don't support pre-vote")
+	}
+
+	// Create a response channel
+	respCh := make(chan *preVoteResult, len(r.configurations.latest.Servers))
+
+	// Increment the term
```
banks, line 2061:
```suggestion
	// Propose the next term without actually changing our state
```

raft.go
```diff
@@ -1907,6 +2041,90 @@ func (r *Raft) electSelf() <-chan *voteResult {
...
+		r.goFunc(func() {
+			defer metrics.MeasureSince([]string{"raft", "candidate", "preElectSelf"}, time.Now())
+			resp := &preVoteResult{voterID: peer.ID}
+
+			err := prevoteTrans.RequestPreVote(peer.ID, peer.Address, req, &resp.RequestPreVoteResponse)
+
+			// If the target server do not support Pre-vote RPC we count this as a granted vote to allow
+			// the cluster to progress.
+			if err != nil && strings.Contains(err.Error(), rpcUnexpectedCommandError) {
+				r.logger.Error("target do not support pre-vote RPC",
```
banks, line 2084:
```suggestion
				r.logger.Error("target does not support pre-vote RPC, treating as granted",
```

banks (approved):
Looks so awesome Dhia.

Some super minor things I spotted inline but see what you think. I don't think they are blocking.

We did discuss offline verifying this change end-to-end in Consul and Vault before we actually merge out of an abundance of caution. Lets report back here when that testing is done!

raft.go
dhiaayachi: yes I think you're right last contact should not be set in here, although I don't think of any scenario where it will create a breakage.

I will remove it to be on the safe side.

raft.go
dhiaayachi: yes you are right, I misread your comment and was thinking that it's not related to the log 😅. I fixed that and added the number of refused prevotes to the log.

PR #537, opened 2023-01-16
feat: allow logger node name to be edited

### LINKED ISSUE #536: How to costumize the Logger Name
Is there an option that I may have not seen to customize the Logger name?

If this is not yet possible, I think I could make a PR to enable this, would this be interesting to the community?

I am doing some tests and demos where Raft Nodes are run in different processes in the same terminal and I have no way from the logs to tell which Node is giving which log message.

Thanks for everything.

JoanFM:
Allow to edit the name used by the Logger via configuration.

Fixes #536

### DIFF
--- a/config.go
+++ b/config.go
@@ -213,6 +213,9 @@ type Config struct {
 	// LogOutput with LogLevel is used.
 	Logger hclog.Logger
 
+	// Name shown by the Logger messages.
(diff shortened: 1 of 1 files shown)


JoanFM:
It seems it should work, I will confirm soon

PR #544, opened 2023-03-08
[NET-3024] Configure GitHub Actions

loshz:
Signed-off-by: Dan Bond <danbond@protonmail.com>

## Description
This PR adds CI workflows for running Go tests in GitHub Actions. It uses the same commands as the existing CircleCI workflow and builds in both x86 and x86-64 architectures.

Other changes:
- Add `CODEOWNERS`
- Configure dependabot
- `go fmt ./...`
- Fix issues in tests related to `go vet`
- Remove `./circleci` config.

Note: we don't need to upload the test output xml as this was a CircleCI specific task https://github.com/hashicorp/raft/blob/main/.circleci/config.yml#L91-L94


loshz:
Potentially blocked on this bug: https://github.com/actions/setup-go/issues/347

net_transport_test.go
```diff
@@ -62,7 +67,7 @@ func TestNetworkTransport_CloseStreams(t *testing.T) {
 				// Verify the command
 				req := rpc.Command.(*AppendEntriesRequest)
 				if !reflect.DeepEqual(req, &args) {
```
loshz, line 69: I'm not sure why we need to do this if we explicitly pass the `&args` below.

modrake (changes requested):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hashicorp/raft
 
-go 1.16
+go 1.20
```
banks, line 3: Is this required? I don't think we want to force all downstream users of the library to only be able to build on the latest go release.

This number in go.mod is the "minimum supported version" not "they current version" and we should only update it if we are actually using a new feature of the language that requires at least this version.

If we made this change, we'd could have issues with our product builds if they are not yet ready for 1.20 or even have trouble pulling fixes into backports of our old releases that have to use older go versions etc.

modrake (approved):

go.mod
banks: I think last time we bumped it we decided "last two" on the basis that that covered most of our supported product releases. But IMO we should leave this as 1.16 until we actually have a specific reason to need language features that are worth breaking compatibility for.

Might be worth checking what Go versions last supported Consul/Nomad/Vault branches built from of you want to be sure but IMO risk is low for this library that we'd introduce backward incompatible changes given how little we commit here so last 2 versions is probably fine?

loshz: Look like Consul/Nomad is on 1.20 and Vault is on 1.19. I will continue building on latest 2 Go versions, but will revert the changes to `go.mod` even though it's "probably" safe.

banks (approved):
LGTM thanks!

PR #545, opened 2023-03-10
Post restore reset

mpalmi:
This PR introduces an interface which acts as a handler for a leaky
abstraction in the structure of underlying log stores. In order to
properly handle post-snapshot-restore cleanup for log stores
generically, we need some awareness of whether the underlying store
permits gaps.

Boltdb allows for gaps in log store indexes, but to handle them it
requires a freelist, which is written on every commit. This is costly,
particularly when the freelist is large. By completely resetting the
LogStore after snapshot restore, we grow the size of the freelist, which would result in performance degradation.

The MonotonicLogStore interface is implemented by LogStores with
guarantees of sequential/monotonic indexes, like raft-wal, but reverts
to the old behavior for for LogStores with index holes, like Boltdb.

The interface also requires special handling within LogStore wrappers (like LogCache), to ensure that the type assertion is passed to the underlying store.

We then use the MonotonicLogStore type assertion to delete all
entries from the LogStore after snapshot restore.

### DIFF
--- a/api.go
+++ b/api.go
@@ -645,6 +645,14 @@ func (r *Raft) restoreSnapshot() error {
 		r.setCommittedConfiguration(conf, index)
 		r.setLatestConfiguration(conf, index)
 
+		// Remove old logs if r.logs is a MonotonicLogStore. Log any errors and
+		// continue.
+		if logs, ok := r.logs.(MonotonicLogStore); ok && logs.IsMonotonic() {
+			if err := r.removeOldLogs(); err != nil {
+				r.logger.Error("failed to reset logs", "error", err)
+			}
+		}
+
 		// Success!
 		return nil
 	}
--- a/api.go
+++ b/api.go
@@ -645,14 +645,6 @@ func (r *Raft) restoreSnapshot() error {
 		r.setCommittedConfiguration(conf, index)
 		r.setLatestConfiguration(conf, index)
 
-		// Remove old logs if r.logs is a MonotonicLogStore. Log any errors and
-		// continue.
-		if logs, ok := r.logs.(MonotonicLogStore); ok && logs.IsMonotonic() {
-			if err := r.removeOldLogs(); err != nil {
-				r.logger.Error("failed to reset logs", "error", err)
-			}
-		}
-
 		// Success!
 		return nil
 	}
--- a/log.go
+++ b/log.go
@@ -122,11 +122,24 @@ type LogStore interface {
 	// StoreLog stores a log entry.
 	StoreLog(log *Log) error
 
-	// StoreLogs stores multiple log entries.
+	// StoreLogs stores multiple log entries. Implementers of StoreLogs with
+	// guarantees of monotonically increasing sequential indexes should make use
+	// of the MonotonicLogStore interface.
 	StoreLogs(logs []*Log) error
 
 	// DeleteRange deletes a range of log entries. The range is inclusive.
 	DeleteRange(min, max uint64) error
+}
+
+// MonotonicLogStore is an optional interface for LogStore implementations with
+// gapless index requirements. If they return true, the LogStore must have an
+// efficient implementation of DeleteLogs, as this called after every snapshot
+// restore when gaps are not allowed. We avoid deleting all records for
+// LogStores that do not implement MonotonicLogStore because this has a major
+// negative performance impact on the BoltDB store that is currently the most
+// widely used.
+type MonotonicLogStore interface {
+	IsMonotonic() bool
 }
 
 func oldestLog(s LogStore) (Log, error) {
--- a/log.go
+++ b/log.go
@@ -122,22 +122,22 @@ type LogStore interface {
 	// StoreLog stores a log entry.
 	StoreLog(log *Log) error
 
-	// StoreLogs stores multiple log entries. Implementers of StoreLogs with
-	// guarantees of monotonically increasing sequential indexes should make use
-	// of the MonotonicLogStore interface.
+	// StoreLogs stores multiple log entries. By default the logs stored may not be contiguous with previous logs (i.e. may have a gap in Index since the last log written). If an implementation can't tolerate this it may optionally implement `MonotonicLogStore` to indicate that this is not allowed. This changes Raft's behaviour after restoring a user snapshot to remove all previous logs instead of relying on a "gap" to signal the discontinuity between logs before the snapshot and logs after.
 	StoreLogs(logs []*Log) error
 
 	// DeleteRange deletes a range of log entries. The range is inclusive.
 	DeleteRange(min, max uint64) error
 }
 
-// MonotonicLogStore is an optional interface for LogStore implementations with
-// gapless index requirements. If they return true, the LogStore must have an
-// efficient implementation of DeleteLogs, as this called after every snapshot
-// restore when gaps are not allowed. We avoid deleting all records for
-// LogStores that do not implement MonotonicLogStore because this has a major
-// negative performance impact on the BoltDB store that is currently the most
-// widely used.
+// MonotonicLogStore is an optional interface for LogStore implementations that
+// cannot tolerate gaps in between the Index values of consecutive log entries. For example, 
+// this may allow more efficient indexing because the Index values are densely populated. If true is 
+// returned, Raft will avoid relying on gaps to trigger re-synching logs on followers after a 
+// snapshot is restored. The LogStore must have an efficient implementation of 
+// DeleteLogs, as this called after every snapshot restore when gaps are not allowed. 
+// We avoid deleting all records for LogStores that do not implement MonotonicLogStore 
+// because this has a major negative performance impact on the BoltDB store that is currently 
+// the most widely used.
 type MonotonicLogStore interface {
 	IsMonotonic() bool
 }
--- a/log.go
+++ b/log.go
@@ -130,13 +130,13 @@ type LogStore interface {
 }
 
 // MonotonicLogStore is an optional interface for LogStore implementations that
-// cannot tolerate gaps in between the Index values of consecutive log entries. For example, 
-// this may allow more efficient indexing because the Index values are densely populated. If true is 
-// returned, Raft will avoid relying on gaps to trigger re-synching logs on followers after a 
-// snapshot is restored. The LogStore must have an efficient implementation of 
-// DeleteLogs, as this called after every snapshot restore when gaps are not allowed. 
-// We avoid deleting all records for LogStores that do not implement MonotonicLogStore 
-// because this has a major negative performance impact on the BoltDB store that is currently 
+// cannot tolerate gaps in between the Index values of consecutive log entries. For example,
+// this may allow more efficient indexing because the Index values are densely populated. If true is
+// returned, Raft will avoid relying on gaps to trigger re-synching logs on followers after a
+// snapshot is restored. The LogStore must have an efficient implementation of
+// DeleteLogs, as this called after every snapshot restore when gaps are not allowed.
+// We avoid deleting all records for LogStores that do not implement MonotonicLogStore
+// because this has a major negative performance impact on the BoltDB store that is currently
 // the most widely used.
 type MonotonicLogStore interface {
 	IsMonotonic() bool
--- a/log.go
+++ b/log.go
@@ -134,9 +134,9 @@ type LogStore interface {
 // this may allow more efficient indexing because the Index values are densely populated. If true is
 // returned, Raft will avoid relying on gaps to trigger re-synching logs on followers after a
 // snapshot is restored. The LogStore must have an efficient implementation of
-// DeleteLogs, as this called after every snapshot restore when gaps are not allowed.
+// DeleteLogs for the case where all logs are removed, as this must be called after snapshot restore when gaps are not allowed.
 // We avoid deleting all records for LogStores that do not implement MonotonicLogStore
-// because this has a major negative performance impact on the BoltDB store that is currently
+// because although it's always correct to do so, it has a major negative performance impact on the BoltDB store that is currently
 // the most widely used.
 type MonotonicLogStore interface {
 	IsMonotonic() bool
--- a/log_cache.go
+++ b/log_cache.go
@@ -33,6 +33,16 @@ func NewLogCache(capacity int, store LogStore) (*LogCache, error) {
 	return c, nil
 }
 
+// IsMonotonic implements the MonotonicLogStore interface. This is a shim to
+// expose the underyling store as monotonically indexed or not.
+func (c *LogCache) IsMonotonic() bool {
+	if store, ok := c.store.(MonotonicLogStore); ok {
+		return store.IsMonotonic()
+	}
+
+	return false
+}
+
 func (c *LogCache) GetLog(idx uint64, log *Log) error {
 	// Check the buffer for an entry
 	c.l.RLock()
--- a/raft.go
+++ b/raft.go
@@ -1129,7 +1129,7 @@ func (r *Raft) restoreUserSnapshot(meta *SnapshotMeta, reader io.Reader) error {
 		}
 	}
 
-	r.logger.Info("restored user snapshot", "index", latestIndex)
+	r.logger.Info("restored user snapshot", "index", lastIndex)
 	return nil
 }
 
--- a/raft.go
+++ b/raft.go
@@ -1122,6 +1122,13 @@ func (r *Raft) restoreUserSnapshot(meta *SnapshotMeta, reader io.Reader) error {
 	r.setLastApplied(lastIndex)
 	r.setLastSnapshot(lastIndex, term)
 
+	// Remove old logs if r.logs is a MonotonicLogStore. Log any errors and continue.
+	if logs, ok := r.logs.(MonotonicLogStore); ok && logs.IsMonotonic() {
+		if err := r.removeOldLogs(); err != nil {
+			r.logger.Error("failed to remove old logs", "error", err)
+		}
+	}
+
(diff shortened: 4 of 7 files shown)


log_cache.go
```diff
@@ -33,6 +33,17 @@ func NewLogCache(capacity int, store LogStore) (*LogCache, error) {
 	return c, nil
 }
 
+// IsMonotonic implements the MonotonicLogStore interface. This is a shim to
+// expose the underyling store as monotonically indexed or not.
+func (c *LogCache) IsMonotonic() bool {
+	store, ok := c.store.(MonotonicLogStore)
+	if ok {
```
banks, line 40: nit: I don't have a strong preference but some Gophers prefer to restrict the scope and do this in one line like:

```suggestion
	if store, ok := c.store.(MonotonicLogStore); ok {
```

testing.go
```diff
@@ -133,6 +133,47 @@ func (m *MockSnapshot) Persist(sink SnapshotSink) error {
 func (m *MockSnapshot) Release() {
 }
 
+// MockMonotonicLogStore is a stubbed LogStore wrapper for testing the
```
mpalmi, line 136:
```suggestion
// MockMonotonicLogStore is a LogStore wrapper for testing the
```

raft.go
```diff
@@ -28,6 +29,12 @@ var (
 	keyLastVoteCand = []byte("LastVoteCand")
 )
 
+// MonotonicLogStore is a simple interface which acts as a compatibility shim
+// for log stores with gapless index requirements.
+type MonotonicLogStore interface {
+	IsMonotonic() bool
+}
+
```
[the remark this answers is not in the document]
mpalmi: This is great feedback, thanks!

raft.go
```diff
@@ -1790,15 +1800,20 @@ func (r *Raft) installSnapshot(rpc RPC, req *InstallSnapshotRequest) {
 	r.setLatestConfiguration(reqConfiguration, reqConfigurationIndex)
 	r.setCommittedConfiguration(reqConfiguration, reqConfigurationIndex)
 
-	// Compact logs, continue even if this fails
-	if err := r.compactLogs(req.LastLogIndex); err != nil {
-		r.logger.Error("failed to compact logs", "error", err)
+	// Fully truncate logs for stores with monotonically increasing log indexes.
+	// Otherwise, compact the logs. Log any errors and continue.
+	err = r.resetMonotonicLogs()
```
banks, line 1805: Using an `error` as a signal for a non-error case isn't ideal. I wonder if it's cleaner just to do the type assertion here?

If this saves a significant amount of duplication a more idiomatic way would be to change `resetMonotonicLogs` to return a bool that indicates whether it did anything or not rather than an error. If there are legit reasons that this could error too then return both, but in that case we should also handle the error up on line 1133 (even if we choose to swallow it or just log it etc.)

snapshot.go
```diff
@@ -249,3 +252,35 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
...
 }
+
+// resetMonotonicLogs truncates all logs for MonotonicLogStores after a snapshot
+// restore.
+func (r *Raft) resetMonotonicLogs() error {
+	// Only truncate if the underlying store has monotonic log index
+	// requirements.
+	logs, ok := r.logs.(MonotonicLogStore)
+	if ok && logs.IsMonotonic() {
+		defer metrics.MeasureSince([]string{"raft", "resetMonotonicLogs"}, time.Now())
```
banks, line 263: nit: maybe we shouldn't leak this complicated internal name to operators to whom it might not make sense?

How about calling this "resetLogs". In Consul we'd document it as only applying to certain log stores or something.

snapshot.go
```diff
@@ -249,3 +252,35 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
...
+	} else {
+		r.logger.Info("skipping truncate for non-monotonic log store")
+		return errNonMonotonicLogStore
+	}
+
+	// Determine log ranges to truncate
+	minLogIdx, err := r.logs.FirstIndex()
```
banks, line 270: MASSIVE nitpick 😄: I'd stick with either `first/last` or `min/max` rather than a mixture.

snapshot.go
```diff
@@ -249,3 +252,35 @@ func (r *Raft) compactLogs(snapIdx uint64) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get first log index: %w", err)
+	}
+
+	lastLogIdx, err := r.logs.LastIndex()
+	if err != nil {
+		return fmt.Errorf("failed to get last log index: %w", err)
+	}
+
+	r.logger.Info("truncating monotonic log store", "from", minLogIdx, "to", lastLogIdx)
```
banks, line 280:
```suggestion
	r.logger.Info("resetting log store", "from", minLogIdx, "to", lastLogIdx)
```
Just to make it more meaningful for operators who don't know about this one weird abstraction detail...

raft_test.go
```diff
@@ -1019,6 +1020,79 @@ func TestRaft_SnapshotRestore(t *testing.T) {
...
+	// We should have restored from the snapshot!
+	if last := r.getLastApplied(); last != snap.Index {
+		t.Fatalf("bad last index: %d, expecting %d", last, snap.Index)
+	}
+
+	// Verify that logs have been reset
+	first, _ := r.logs.FirstIndex()
+	last, _ := r.logs.LastIndex()
+	assert.Zero(t, first)
+	assert.Zero(t, last)
```
briankassouf, line 1093: Does this imply that we are reset the WAL Index back to 0 after a successful snapshot restore? What will be the index of the next value i write through raft?

mpalmi: I believe the next index would just be written as usual. A log entry can be appended to any index in an empty log. We just have to clear the stale entries from the log (post-restore) so that we don't violate raft-wal's assertion of a gapless log.

banks: @briankassouf great question but no it doesn't imply that. As Mike said, this is just validating that the LogStore now contains zero logs (and therefore first and last are no longer meaningful concepts).

The first write to an empty LogStore doesn't necessarily have to be index 1 since a follower that starts after the leader has already taken at least one snapshot will never experience that.

In the case of snapshots, the current behaviour of this library is to artificially set the "index" of the snapshot it's restoring to the current `LastIndex + 1` which ensures it's larger than any index every actually written to this cluster (and keeps the term). So the first Log written after the restore (the NoOp log used to force snapshot replication to followres) will always be at `LastIndex + 1` of the current log, and that will be the new first log stored in each node's LogStore with this change.

log.go
```diff
@@ -122,13 +122,26 @@ type LogStore interface {
 	// StoreLog stores a log entry.
 	StoreLog(log *Log) error
 
-	// StoreLogs stores multiple log entries.
+	// StoreLogs stores multiple log entries. Implementers of StoreLogs with
+	// guarantees of monotonically increasing sequential indexes should make use
+	// of the MonotonicLogStore interface.
```
banks, line 127:
```suggestion
	// StoreLogs stores multiple log entries. By default the logs stored may not be contiguous with previous logs (i.e. may have a gap in Index since the last log written). If an implementation can't tolerate this it may optionally implement `MonotonicLogStore` to indicate that this is not allowed. This changes Raft's behaviour after restoring a user snapshot to remove all previous logs instead of relying on a "gap" to signal the discontinuity between logs before the snapshot and logs after.
```

log.go
```diff
@@ -122,13 +122,26 @@ type LogStore interface {
...
 	DeleteRange(min, max uint64) error
 }
 
+// MonotonicLogStore is an optional interface for LogStore implementations with
+// gapless index requirements. If they return true, the LogStore must have an
+// efficient implementation of DeleteLogs, as this called after every snapshot
+// restore when gaps are not allowed. We avoid deleting all records for
+// LogStores that do not implement MonotonicLogStore because this has a major
+// negative performance impact on the BoltDB store that is currently the most
+// widely used.
```
banks, line 140:
```suggestion
// MonotonicLogStore is an optional interface for LogStore implementations that
// cannot tolerate gaps in between the Index values of consecutive log entries. For example, 
// this may allow more efficient indexing because the Index values are densely populated. If true is 
// returned, Raft will avoid relying on gaps to trigger re-synching logs on followers after a 
// snapshot is restored. The LogStore must have an efficient implementation of 
// DeleteLogs, as this called after every snapshot restore when gaps are not allowed. 
// We avoid deleting all records for LogStores that do not implement MonotonicLogStore 
// because this has a major negative performance impact on the BoltDB store that is currently 
// the most widely used.
```

raft_test.go
banks: Good catch Brian. This actually indicates one place where we should _not_ change the behaviour. Will add comments above.

api.go
```diff
@@ -645,6 +645,14 @@ func (r *Raft) restoreSnapshot() error {
...
 		r.setLatestConfiguration(conf, index)
 
+		// Remove old logs if r.logs is a MonotonicLogStore. Log any errors and
+		// continue.
+		if logs, ok := r.logs.(MonotonicLogStore); ok && logs.IsMonotonic() {
+			if err := r.removeOldLogs(); err != nil {
+				r.logger.Error("failed to reset logs", "error", err)
+			}
+		}
+
```
banks, line 655: This change is not necessary and in fact will _loose data_ in raft.

This method is called on startup to restore from the snapshot on disk which is very different from restoring a user snapshot (which invalidates our current state).

As written, this would delete all the recently written things from disk on startup - if we happen to restart or crash again before the next snapshot we would _loose data_.

So we should remove this from here!

banks (changes requested):
@mpalmi Nice. This is also super close, but as it stands it has a huge issue we should fix!

In the last changes we added the log clearing code into the startup restore which is wrong and will delete loose committed data!

Let me know if it's not clear why that is or if I've misunderstood something here - restoring the on-disk snapshot is very different and not at all problematic. We are only trying to fix the problem of restoring from an _external_ snapshot which has the effect of making all the current state invalid and useless!

raft_test.go
banks: So this test should _not_ touch the logs at all - we are just restoring from local disk not replacing the local state with external snapshot state.

We would however need a test that exercises the path that the Vault integration tests do:
 - setup  a cluster with at least one follower (probably 3 nodes)
 - write some stuff
 - take a snapshot and store it somewhere
 - write some more stuff
 - restore the snapshot on leader
 - validate that the leader and followers have only the restored data and not the extra stuff written in between...

That is the test that exercises the issues that this PR is designed to fix I think?

mpalmi: Makes sense. Should be easy enough to add. I will have a bit of time to work on this later today. I'll ping you for re-review then. Sorry for the churn on this. Hopefully the next pass is good to go!

api.go
jmurret: Paired with @banks today to get context.  The correction to the test and the code is contained in this commit: https://github.com/hashicorp/raft/pull/545/commits/dd083107e93e2ab7c7b0679d38bd225e0fc6720c

jmurret (approved):
Great work here by all!

raft_test.go
```diff
@@ -1019,6 +1020,88 @@ func TestRaft_SnapshotRestore(t *testing.T) {
 	}
 }
 
+func TestRaft_SnapshotRestore_Monotonic(t *testing.T) {
```
mpalmi, line 1023: Do we need this test anymore? I thought this was testing that a restore during NewRaft reset the log indexes, but we're no longer modifying that code path. I may be mistaken, but I think we might want to add [a check ](https://github.com/hashicorp/raft/pull/545#discussion_r1134154086) in the `snapShotAndRestore` helper, rather than here.

banks: IMO it's worth keeping if only to ensure we don't regress and accidentally re-add any trimming of logs to this code path in the future!

The fact that we could have added that code that would loose committed data but no existing tests would have caught it makes me feel like this is worth keeping for future!

I think we learned and made the tests better from this PR! Thanks Mike!

jmurret: I added https://github.com/hashicorp/raft/pull/545/commits/fc09cbd0a778414d990a44d3ed3aad45a0313a72 to verify whether logs are removed or not during user restores (and based off @mpalmi 's suggestion).  
I also added a check in here for when restoring to a new cluster vs the existing cluster.  I think it may also serve as the verification in raft of this integration test that is added to Consul: https://github.com/hashicorp/consul/pull/16647/files

Feel free to comment, adjust, or remove.

cc: @dhiaayachi @banks

raft_test.go
```diff
@@ -1382,6 +1478,23 @@ func snapshotAndRestore(t *testing.T, offset uint64) {
 	// Get the last index before the restore.
 	preIndex := leader.getLastIndex()
 
+	if restoreNewCluster {
+		var c2 *cluster
+		if monotonicLogStore {
+			opts := &MakeClusterOpts{
```
mpalmi, line 1484: We can probably just re-use the cluster opts from above if you scope the opts var to the function. All of the vars are the same.

raft_test.go
```diff
@@ -1397,17 +1510,36 @@ func snapshotAndRestore(t *testing.T, offset uint64) {
...
 		t.Fatalf("Index was not updated correctly: %d vs. %d", lastIndex, expected)
 	}
 
-	// Ensure all the logs are the same and that we have everything that was
+	// Ensure raft logs are removed for monotonic log stores but remain
+	// untouched for non-monotic (BoltDB) logstores.
+	// When first index = 1, then logs have remained untouched.
+	// When first indext is set to the next commit index / last index, then
+	// it means logs have been removed.
+	firstLogIndex, err := leader.logs.FirstIndex()
```
mpalmi, line 1531: I think @banks mentioned that we should check the follower nodes as well.

mpalmi (commented):
Looks great! Just a couple of questions that are not showstoppers, but one is related to a comment from @banks, so might be worth addressing.

raft_test.go
jmurret: Good call.  Added in https://github.com/hashicorp/raft/pull/545/commits/7d73c137a93834c6d84107c2f14eb332020b809f

log.go
```diff
@@ -122,13 +122,26 @@ type LogStore interface {
...
 	// DeleteRange deletes a range of log entries. The range is inclusive.
 	DeleteRange(min, max uint64) error
 }
 
+// MonotonicLogStore is an optional interface for LogStore implementations that
+// cannot tolerate gaps in between the Index values of consecutive log entries. For example,
+// this may allow more efficient indexing because the Index values are densely populated. If true is
+// returned, Raft will avoid relying on gaps to trigger re-synching logs on followers after a
+// snapshot is restored. The LogStore must have an efficient implementation of
+// DeleteLogs, as this called after every snapshot restore when gaps are not allowed.
```
banks, line 137:
```suggestion
// DeleteLogs for the case where all logs are removed, as this must be called after snapshot restore when gaps are not allowed.
```

log.go
```diff
@@ -122,13 +122,26 @@ type LogStore interface {
...
+// We avoid deleting all records for LogStores that do not implement MonotonicLogStore
+// because this has a major negative performance impact on the BoltDB store that is currently
```
banks, line 139:
```suggestion
// because although it's always correct to do so, it has a major negative performance impact on the BoltDB store that is currently
```

raft_test.go
banks: Great job @jmurret your new test that exercises all the combinations covers the actual snapshot Restore cases that we _did_ change.

I'd vote to keep this test for now since it validates that the "restore on startup" case does _not_ delete logs. Perhaps the one suggestion would be that we rename it to something like `TestRaft_RestoreSnapshotOnStartup_Monotonic` or something? TO make it clear that it's different from testing the `raft.RestoreSnapshot()` cases in the other test?

banks (approved):
LGTM. Thanks @jmurret.

I had some minor comment suggestions and a test naming suggestion in line but nothing blocking.

@dhiaayachi do you want to do a final pass at this?

I had this branch running against Consul's restore integration test yesterday so I feel happy we resolved the issues here (in combination with https://github.com/hashicorp/raft-wal/pull/24),

dhiaayachi (approved):
Great Job All!!
I added a nit but I'm ok with merging this as is.

raft_test.go
jmurret: Renamed this test.  Resolving.

raft.go
```diff
@@ -1790,15 +1797,19 @@ func (r *Raft) installSnapshot(rpc RPC, req *InstallSnapshotRequest) {
...
 
-	// Compact logs, continue even if this fails
-	if err := r.compactLogs(req.LastLogIndex); err != nil {
+	// Clear old logs if r.logs is a MonotonicLogStore. Otherwise compact the
+	// logs. In both cases, log any errors and continue.
+	if mlogs, ok := r.logs.(MonotonicLogStore); ok && mlogs.IsMonotonic() {
+		if err := r.removeOldLogs(); err != nil {
+			r.logger.Error("failed to reset logs", "error", err)
+		}
+	} else if err := r.compactLogs(req.LastLogIndex); err != nil {
```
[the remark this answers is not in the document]
banks: Heh I'll take the blame for that. Mike originally thought the same and I argued that "compact" was more subtle and almost all the logic in there was around "how many logs to keep" so it was maybe better to just have an explicit separate "delete everything".

But honestly it was a quick comment not a strongly held conviction so if someone else feels it is cleaner to combine I won't disagree!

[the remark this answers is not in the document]
jmurret: Addressed this in https://github.com/hashicorp/raft/pull/545/commits/fdcb70b2f362b397ac6f1e4950b03231dd5de3fc

PR #550, opened 2023-03-17
Adding changelog for 1.4.0

jmurret:


dhiaayachi:
🤔 This changlog seem to be missing all the changes in minor version. Are they supposed to be in it?

jmurret:
I'm not sure of the history with that, but if we have to backfill it, I would like to add an issue for that and address it after this is released.

banks (approved):
Nice. Thanks!

PR #562, opened 2023-07-03
gofumpt

mpalmi:
This PR resolves some issues in CI when modifying `api.go`.

### DIFF
--- a/fuzzy/fsm_batch.go
+++ b/fuzzy/fsm_batch.go
@@ -1,6 +1,7 @@
 // Copyright (c) HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
+//go:build batchtest
 // +build batchtest
 
 package fuzzy
--- a/fuzzy/transport.go
+++ b/fuzzy/transport.go
@@ -8,20 +8,19 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"github.com/hashicorp/go-hclog"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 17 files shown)


fuzzy/fsm_batch.go
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) HashiCorp, Inc.
 // SPDX-License-Identifier: MPL-2.0
 
+//go:build batchtest
```
banks, line 4: What is the minimum version of Go that supports this syntax? I suggest we bump go.mod to that version to avoid surprises.

fuzzy/transport.go
```diff
@@ -8,20 +8,19 @@ import (
...
 	"fmt"
-	"github.com/hashicorp/go-hclog"
 	"io"
 	"os"
 	"path/filepath"
 	"sync"
 	"time"
 
+	"github.com/hashicorp/go-hclog"
+
```
banks, line 18:
```suggestion
	"github.com/hashicorp/go-hclog"
```

banks: ^ removing the extra newline

fuzzy/fsm_batch.go
mpalmi: Addressed here: https://github.com/hashicorp/raft/pull/562/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6

Good call!

banks (approved):
LGTM!

Issue #568, opened 2023-09-08
Panic brain-split in appendEntries call


nopnoping:
Hello team. 
I'm confused about this: https://github.com/nopnoping/raft/blob/main/raft.go#L1393.
It will be follower when the state is leader and the term equals requester term, but this situation is a brain-split which should panic an error. Although it may never happen, explicitly panic brain-split error may better.
Like this: https://github.com/nopnoping/raft/blob/opt_appendEntries/raft.go#L1391-L1402

banks:
Hi @nopnoping, I can see your logic there but that's not actually how the raft protocol (or any consensus protocol I know of works).

It's in general impossible to perfectly solve the distributed leader election problem in a real network (this is called the FLP result in literature) and neither Raft nor Paxos makes the guarantee that "at most one node will consider itself to be leader at any time". To put this another way: "split brain" suituations are **enevitable** in real world situations and the correctness of a consensus protocol is not in avoiding that happening entirely, but rather ensuring that _when it does, nothing bad happens in the system_.

In other words if we panic in the case your PR does **we will break raft!**. That's because systems _can and do_ get into a situation where for a short time, two nodes think they are the leader. I've been testing this exact scenario in Vault for the last few months - it's really easy to get into and we see it in production too. In those cases, your PR would cause one or both of the nodes to crash, making the whole system unavailable. The whole point of raft is correctly tolerate this situation and having one node crash is not that! At best, your added panic will make it take significantly longer for the cluster to get back to a healthy state (and cause scary logs that will cause operators great concern too). At worst, it could destroy the availability of the cluster and potentially force a manual recovery (though that would be more rare).

Here's an example of how this can happen:

 1. imagine a 3 node cluster with nodes A, B, C
 2. node A is leader and all is healthy
 3. at time t1 node A is suspended - perhaps someone paused the container to migrate it, hit ctrl-z or just the CPU is suddenly hammered by some other process and so it gets no CPU time for a few seconds all of these are plausible!
 4. node B hits it's Heartbeat timeout first and becomes a candidate and gets elected by a vote from C
 5. node A un-pauses (gets scheduled on the CPU again).

At this point, both nodes A and B think they are the leader although B has a higher term. (Notice since A was suspended and not just partitioned it never stepped down since it's own LeaderLease timer has not fired yet).

The current behaviour, following the Raft paper, is that on Node A one of these things will happen first:
 1. It's leader lease timeout fires and it's not had heartbeats while it was suspended so "steps down"
[...]

PR #569, opened 2023-09-08
Optimize appendEntries to panic error in brain-split situation

nopnoping:
This commit panic brain-split error in appendEntries call when state is leader and term equals requester term.
https://github.com/hashicorp/raft/issues/568

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1388,9 +1388,13 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


banks:
Thanks for the contribution @nopnoping. 

I've explained in some detail in https://github.com/hashicorp/raft/issues/568#issuecomment-1711693475 why I think this change is strictly worse for raft and the issue it's trying to fix is actually not a problem but part of how the raft algorithm is expected to work.

I'll close this for now.

Issue #473, opened 2021-09-13
msgpack dependencies 


SimonRichardson:
This is more of a potential problem that's up and coming, rather than a problem right now (msgpack v1.1.5 tag and mspack v0.5.5 as seen https://github.com/hashicorp/go-msgpack/tags). The upstream msgpack has changed how time was encoded/decoded and the workaround was [provided](https://github.com/ugorji/go/issues/269).

I suspect either two options, decode the data with TimeNotBuiltin set and then encode it without, allowing a stable migration or it might be prudent to expose `TimeNotBuiltin` in the raft configuration (under another name `LegacyEncoding` or similar), so this can be changed for users with existing data.

ncabatoff:
Hi @SimonRichardson,

It looks like there's been a new release of hashicorp/go-msgpack, and presumably that maintains our fork's absention from the new time serialization.  We'll update this library to make use of that 1.1.5 release.

SimonRichardson:
If you update raft, then existing deployments that utilize the 0.5.5 release that upgrades to 1.5.5 will hit the following error, unless a migration is applied.

> failed to get last log at index 52: msgpack decode error [pos 13]: invalid length of bytes for decoding time - expecting 4 or 8 or 12, got 15

ncabatoff:
Ah, I think I misunderstood you.  I thought you were talking about upgrading to newer hashicorp/go-msgpack versions that already exist.  But your actual concern was: if hashicorp/go-msgpack were to pull in the upstream changes from ugorji/go, in particular https://github.com/ugorji/go/commit/debb8e2d2e8bd8cf1e9b6d806cf5e58df86b970c, then any time.TIme values we serialized with older msgpack versions would be a problem.  Is that right?

SimonRichardson:
Indeed.

wallyworld:
This is still an issue that it would be good to see a fix for.

ncabatoff:
This has been addressed, see https://github.com/hashicorp/raft/pull/577#issuecomment-1810897224.

PR #577, opened 2023-10-18
Upgrade go-msgpack to 2.1.1

swenson:
We add a parameter that allows the behavior of the time encoding to be changed by `NetworkTransportConfig.MsgpackUseNewTimeFormat`. The default value is `false`, which keeps the current behavior of time encoding is compatible with with `go-msgpack` v0.5.5 (the version in `go.mod` prior to this change).

However, users who upgrade to this version of Raft will no longer have their own version of pre-v2 `go-msgpack` honored, which could cause unexpected binary compatibility problems.

This will be a problem for HashiCorp products who might blindly upgrade:
* [Vault](https://github.com/hashicorp/vault/blob/f2f532ec802d356346c2302e9c38734f2a594f3f/go.mod#L98) (`go-msgpack` v1.1.5)
* [Nomad](https://github.com/hashicorp/nomad/blob/cb2363f2fbbb771af9c8a62ed09a60250df647a2/go.mod#L64) (`go-msgpack` v1.1.5).

I don't believe this will be a problem for [Consul](https://github.com/hashicorp/consul/blob/8eb074e7c17a0a2077436637c06f521b036407f5/go.mod#L199) (`go-msgpack` v0.5.5).

For the sake of non-HashiCorp products using this library, I'd recommend merging this and making a release with a full `v2` so that people are not surprised.

I'll follow-up this PR and release with PRs to upgrade Vault, Consul, and Nomad.

I tested this branch with Vault using
`NetworkTransportConfig.MsgpackUseNewTimeFormat` set to `true` and verified it was able to replicate to Vault 1.15.0 (which uses `go-msgpack` v1.1.5). (I also tested that the same setting of `false` would fail to replicate to Vault 1.15.0 with a time decoding error, as expected).

I also added a quick benchmark in `bench_test.go`, and did not notice any significant performance differences before and after this change (which is not surprising, since I don't assume msgpack is a significant cost for snapshots).


bench_test.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+	}
+
+	for i := 0; i < b.N; i++ {
+		// Do some commits
+		applyAndWait(raft, 100, 10)
+		// Do a snapshot
+		NoErr(WaitFuture(raft.raft.Snapshot()), b)
+	}
+}
```
banks, line 43: Was this benchmark used to check before and after performance of the msgpack encoding? What were the results?

util_test.go
```diff
@@ -4,11 +4,29 @@
...
+	}
+	buf, err := encodeMsgPack(tm)
+
+	expected := []byte{175, 1, 0, 0, 0, 14, 187, 75, 55, 229, 0, 0, 0, 0, 255, 255}
+
+	if !bytes.Equal(buf.Bytes(), expected) {
+		t.Errorf("Expected time %s to encode as %+v but got %+v", stamp, expected, buf.Bytes())
+	}
+}
+
```
banks, line 29: Would it be worth having a similar test that encodes an entire `Log` populated with some arbitrary data? My thinking is that that struct is really the main thing msgpack is used for other than some minor RPC framing.

As far as I know the only part of the encoding that caused issues due to changes upstream was the time.Time in the `AppendedAt` field which this covers, but it might provided added confidence that there are no other unknown incompatibilities too?

Not essential, just a thought.

banks (approved):
Thanks @swenson.

I thought through whether there is anything else that will help us validate that this doesn't break things but I'm don't think there is. I had one suggestion inline that I don't expect to change anything but would give maybe a tiny bit of extra confidence.

bench_test.go
swenson: Yes. Differences were basically non-existent, as I suspect the disk I/O / sync was the dominant cost.

util_test.go
swenson: So, there are some tests like that in `go-msgpack` already: https://github.com/hashicorp/go-msgpack/blob/main/codec/raft_test.go which links to a bunch of data I captured from running Raft with various versions of `go-msgpack` https://github.com/hashicorp/go-msgpack/tree/main/codec/internal/testdata

banks:
> However, users who upgrade to this version of Raft will no longer have their own version of pre-v2 go-msgpack honored, which could cause unexpected binary compatibility problems.

🤔 It took me a while to parse this and think through the ramifications. Is this correct @swenson ?


1. If you currently build against go-msgpack 1.1.5 then the problem is that your raft messages are currently encoded on the wire with new time format while this PR will default to old format. While the new node can accept messages from the old leader, if the new node becomes leader while non-upgraded servers are still present it will cause them to crash or error. **Fix: ensure that you set the new `NetworkTransportConfig.MsgpackUseNewTimeFormat` to `true` and test mixed scenarios**
 2. If you currently build against go-msgpack 0.5.5 or earlier and you use go-msgpack in your app outside of raft, then you should be OK because raft will now import the new msgpack version as `v2` and leave your app importing v0.5.5. If you don't use it outside of raft you are also OK.

> For the sake of non-HashiCorp products using this library, I'd recommend merging this and making a release with a full v2 so that people are not surprised.

I'm not sure who this advice would be aimed at though? Is that if you import raft but are also a library with downstream consumers?

swenson:
> > However, users who upgrade to this version of Raft will no longer have their own version of pre-v2 go-msgpack honored, which could cause unexpected binary compatibility problems.
> 
> 🤔 It took me a while to parse this and think through the ramifications. Is this correct @swenson ?
> 
> 1. If you currently build against go-msgpack 1.1.5 then the problem is that your raft messages are currently encoded on the wire with new time format while this PR will default to old format. While the new node can accept messages from the old leader, if the new node becomes leader while non-upgraded servers are still present it will cause them to crash or error. **Fix: ensure that you set the new `NetworkTransportConfig.MsgpackUseNewTimeFormat` to `true` and test mixed scenarios**
> 2. If you currently build against go-msgpack 0.5.5 or earlier and you use go-msgpack in your app outside of raft, then you should be OK because raft will now import the new msgpack version as `v2` and leave your app importing v0.5.5. If you don't use it outside of raft you are also OK.

Exactly!

> > For the sake of non-HashiCorp products using this library, I'd recommend merging this and making a release with a full v2 so that people are not surprised.
> 
> I'm not sure who this advice would be aimed at though? Is that if you import raft but are also a library with downstream consumers?

This isn't strictly necessary. But, I assume there might be non-HashiCorp libraries or servers using this library? And if so, the next release breaking compatibility unless you specify a certain flag, depending on what version you were using before and what version of go-msgpack you might have in your `go.mod`, might be surprising. So, I thought doing a `v2` release might help alleviate problems.

(With HashiCorp stuff, I think we have it mostly handled.)

swenson:
(If we are doing a release, we could also put the upgrade notes in the release notes)

HouseK:
Note: This might handle https://github.com/hashicorp/raft/issues/473

swenson:
@banks pinging this again

swenson:
(Oops, I missed that you approved this @banks. Thanks! I'll wait to merge in case there is something else you wanted to bring up.)

banks:
@swenson thanks. Still OK with merging this, my only pause is about how and where we document the nuances I tried to spell out above - do you have a plan for that? I guess in the raft library changelog `CHANGELOG.md` (although I'm not sure how consistent we've been at maintaining that...) but also maybe on the releases page when we draft that? Maybe an internal email to other teams within HC who use this? Do you think it's worth drafting that and getting feedback as part of the PR?

swenson:
@banks will do. I'll draft something in a comment later today.

swenson:
@banks how about:

## Email

Hi developers,

If you aren't a developer who uses the https://github.com/hashicorp/raft library, you can stop reading now.

tl;dr The next Raft release (1.6.0) will switch the underlying hashicorp/go-msgpack library from v0.5.5 to v2.1.1. On its own, v2.1.1 should have the same *default* encoding as v0.5.5.

However, some users of the library have overridden the go-msgpack library with other versions that may be use an incompatible binary encoding.

It is possible in go-msgpack v2.1.1 and Raft 1.6.0 to specify which binary encoding is preferred at runtime.

We have PRs out for Consul, Nomad, and Vault to facilitate these upgrades and set the options appropriately:

- https://github.com/hashicorp/vault/pull/23750
- https://github.com/hashicorp/nomad/pull/18812
- https://github.com/hashicorp/consul/pull/19313

More details are available in the PR https://github.com/hashicorp/raft/pull/577 and in the release notes (TBD).

If you have further questions, feel free to reach out in #tech-raft.

Thanks,
--The Raft maintainers


## Release notes for 1.6.0

Upgrade hashicorp/go-msgpack to v2, with go.mod upgraded from v0.5.5 to v2.1.1.

go-msgpack v2.1.1 is by default binary compatible with v0.5.5 ("non-builtin" encoding of `time.Time`), but can decode messages produced by v1.1.5 as well ("builtin" encoding of `time.Time`).

However, if users of this libary overrode the version of go-msgpack (especially to v1), this **could break** compatibility if raft nodes are running a mix of versions.

This compatibility can be configured at runtime in Raft using `NetworkTransportConfig.MsgpackUseNewTimeFormat` -- the default is `false`, which maintains compatibility with `go-msgpack` v0.5.5, but if set to `true`,
will be compatible with `go-msgpack` v1.1.5.

banks:
Those look awesome @swenson thanks. I think we can maybe duplicate the Release Notes on both the Releases page in GH and in CHANGELOG.md - we normally update that by hand it seems and on main. If you remember to do that before cutting the release so it's also in the tag then even better but I don't think we are very consistent about that!

swenson:
Thanks!

PR #579, opened 2023-10-30
X/async log

banks:


log_cache_async.go
```diff
@@ -0,0 +1,405 @@
...
+	// It's very unlikely there won't be, but if we are writing really fast into a
+	// small cache and the flusher is blocked on IO for a while then we need to
+	// ensure we don't overwrite cache entries that are not persistent yet!
+	for !hasSpaceFor(len(logs), lastIdx, persistedIdx, c.size) {
+		// We need to block and wait for they sync thread to persist some more
+		// stuff! We do that by sending a sync request even though it's already busy
+		// this lets us get notified about when it's free. Note that even though we
+		// unlock and it's _possible_ for another StoreLogsAsync call to be made,
+		doneCh := make(chan struct{})
+		c.state.triggerChan <- syncRequest{startTime: start, doneCh: doneCh}
```
banks, line 350: This causes a deadlock right now since the flusher thread is likely flushing right now and then will need to take state lock before it can update the cache state and continue it's loop to receive this trigger.

We need to unlock the state lock before this send not just before the `doneCh` read

banks:
We found at least 3 bugs in this implementation during performance testing this week!

1. DeleteRange doesn’t work while in async mode. I thought this wasn’t needed for some reason but it is and it results in leader unable to truncate logs after snapshot.
2. Deadlock! If the cache fills up there is a deadlock. I’ve commented above and it’s easy to fix, I didn’t expect to hit that code path though, it’s possible I only did because of #3.
3. It seems from the graphs we saw that something else is very wrong - we saw disk throughput skyrocket, way more than without this until it hits limit of disk  and everything falls apart. It would seem like maybe the logic to update the cursors after flush is wrong and so the cache is never "emptying" and so the next flush is flushing the same data again etc. That should be easy enough to verify!

log_cache_async.go
```diff
@@ -0,0 +1,405 @@
...
+	err := c.store.StoreLogs(logs)
+
+	lwc := LogWriteCompletion{
+		PersistentIndex: logs[0].Index - 1,
+		Error:           err,
+		Duration:        time.Since(start),
+	}
+	if err == nil {
+		lwc.PersistentIndex = logs[len(logs)-1].Index
+	}
```
banks, line 298: Yep, here is bug #3 as I guessed. We are not updating `c.persistentIndex` here so the next flush is going to flush all the same logs again and never "empty" the cache.

banks:
TODO:
* [ ] Test/verify that this works correctly in all cases now that the leader's FSM snapshot _could possibly_ be at a later index than it's logs after a restart. Do we still restore correctly in that case?

banks:
Another extremely subtle issue that we'd have to resolve before merging this if we choose to exists with the assumptions made during LeadershipTransfer.

Right now there is an implicit assumption that by [setting the `leadershipTransferInProgress` flag during the leader loop](https://github.com/hashicorp/raft/blob/1462fd5e80ad0eb38748f68198505025cb2c96d8/raft.go#L721) before initiating a transfer, we guarantee:
 1. All logs the leader has accepted up to this point are already written to it's own log (and therefore will be accounted for when "catching up" the target new leader just by reading `getLastLog`.
 2. All logs accepted by the leader loop after this point (up until the transfer completes and we either step down or resume normal operation) will be immediately failed with `ErrLeadershipTransferInProgress`.

At first, I thought this PR would violate that assumption because the leader loop might move on before logs were persisted and start a transfer assuming that no concurrent log appends are happening.

After careful thought, I think this is actually _not_ an issue. Here's my reasoning:

 1. Regardless of whether they fsynced we still call `setLastLog` with the highest log index before we return from `dispatchLogs`
 2. So a leadership transfer might be concurrent with the actual disk sync (i.e. the leader might not have finished committing the logs yet), but it can't start persisting more because of the existing check that will cause further log appends to error with `ErrLeadershipTransferInProgress`
 3. `leadershipTransfer` will replicate up to the last log index that was set, regardless of whether that log is durable on disk or not - the async log store guarantees to return it again to a `Get` even before it's durable.

So the transfer will behave the same way as now, even though it's possible that the follower will end up with a _more_ up to date log (at least durably on disk) than the current leader by the time we issue the `TimeoutNow` RPC to force it to hold an election, it can't be less up-to-date.

banks:
I think I found the bug! Although we were now correctly updating the persistent index after my last fix, we still weren't clearing the `batch` slice so were re-writing the same logs over and over.

A `monotonic` log store would have errored but I testing in Vault with BoltDB which is not monotonic and so will happily store the same logs over and over again with each write.

I've updated the tests to prove this:

 1. The in-mem store is now (optionally) monotonic to catch this
 2. I realised the test was not correctly exercising completions to detect errors!
 3. I realised the test is non-deterministic and timing sensitive as to how many flushes occur so now it allows for any valid combination of flushes

banks:
I think this is OK for perf testing. One TODO if we choose to take it forward:
 - [ ] Properly test the buffer overflow logic and that we don't loose unpersisted writes if we wrap around.

PR #581, opened 2023-11-23
Reproduce leadership transfer failures when writes happen during transfer

ncabatoff:
Also fix the problem partially: it can still occur if we dispatchLogs after the transfer message has been sent to the target, before it wins the election.

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -6,7 +6,9 @@ package raft
 import (
 	"bytes"
 	"container/list"
+	"context"
 	"fmt"
+	"golang.org/x/sync/semaphore"
 	"io"
 	"sync/atomic"
 	"time"
@@ -92,6 +94,7 @@ type leaderState struct {
 	replState                    map[ServerID]*followerReplication
 	notify                       map[*verifyFuture]struct{}
 	stepDown                     chan struct{}
+	applyable                    *semaphore.Weighted
 }
 
 // setLeader is used to modify the current leader Address and ID of the cluster
@@ -401,7 +404,7 @@ func (r *Raft) setLeadershipTransferInProgress(v bool) {
 
 func (r *Raft) getLeadershipTransferInProgress() bool {
 	v := atomic.LoadInt32(&r.leaderState.leadershipTransferInProgress)
-	return v == 1
+	return v > 0
 }
 
 func (r *Raft) setupLeaderState() {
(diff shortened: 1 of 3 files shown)


raft.go
```diff
@@ -692,8 +692,21 @@ func (r *Raft) leaderLoop() {
...
+						// Wait for up to ElectionTimeout before flagging the
+						// leadership transfer as done and unblocking applies in
+						// the leaderLoop.
+						select {
+						case <-time.After(r.config().ElectionTimeout):
+							err := fmt.Errorf("leadership transfer timeout")
+							r.logger.Debug(err.Error())
+							future.respond(err)
+						case <-leftLeaderLoop:
+							r.logger.Debug("lost leadership during transfer (expected)")
```
banks, line 706: What do you think about just staying in the loop here instead of starting a _new_ `ElectionTimeout` ticker and having to duplicate the leftLeaderLoop code path too?

I was originally expecting that we'd not wait for _another_ `ElectionTimeout` but rather keep waiting for _up to_ the original one we started for the transfer to complete. I don't think it's terrible to do it this way but it strikes me that it's less code and _maybe_ simpler to reason able if the behaviour is just: block until leadership transfer works or one election timeout from handling the request...

The rationale for using ElectionTimeout here presumably was that, LeadershipTransfer is only an optimization to avoid just ungracefully stopping and having to wait for an election timeout... so letting leader transfer take longer than a whole election timeout is a little bit defeating the point and we should probably return fast and let the old leader just shut down if it wants to.

I think that rationale gets a bit more subtle in cases like Autopilot upgrade where we _rely_ on Leadership Transfer working before we move on to the next phase rather than just shutting down anyway.

tl;dr, I don't think the behaviour here is bad or wrong especially and it beats the bug without this wait. If it turns out easier to  follow to duplicate the code and make the behavior more explicit like you have here I'm OK with it. Just curious if you tried simply replacing this `else` branch with `continue` which I _think_ is also a correct fix?

banks (approved):
I think this looks great and don't see any issues.

I did have one question in line about whether or not an alternative was easier to read/reason about but I'm not really convinced either way to be honest and I think both are effectively evuivalent so not blocking!

raft.go
ncabatoff: Yes, I tried that.  It got messy due to the other cases wanting to read from doneCh.  In the end I decided this was clearer.

raft_test.go
```diff
@@ -2337,17 +2337,71 @@ func TestRaft_LeadershipTransferWithOneNode(t *testing.T) {
 	}
 }
 
+func TestRaft_LeadershipTransferWithWrites(t *testing.T) {
```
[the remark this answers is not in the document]
ncabatoff: I couldn't give you numbers, but it didn't take many tries to provoke a failure.  It can still fail if I up the concurrency enough, e.g. if I run 16 parallel -race instances.  But I suspect the same is true for many other tests, as the failures are things like timeouts due to heartbeats not being processed quickly enough.

PR #584, opened 2024-01-04
Add reference use of Hashicorp Raft

otoolep:
Many folks have told me they find this example useful, perhaps you'd like to include it here.


freddygv (approved):
Thank you!

PR #591, opened 2024-04-02
fix: serialize NetworkTransport data race on ServerAddr()

moogacs:
`ServerAddr()` is called concurrently through `AppendEntries`, `RequestVote`. 
for example `getConn()` is thread safe [here](https://github.com/hashicorp/raft/blob/cc2bb080d64f172ff2fcad13289e7ec07dd2e6c6/net_transport.go#L383) by calling `connPoolLock`, however `getProviderAddressOrFallback()` is not thread safe.
 
given that the lib doesn't control which data type is used by the consumer and there is no documentation for how `ServerAddr()` could be used by RAFT lib, so this PR protecting reading `ServerAddr()` by adding `RWMutex` to the `NetworkTransport` and serialize access to `serverAddressLock`

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -91,6 +91,7 @@ type NetworkTransport struct {
 	maxPool     int
 	maxInFlight int
 
+	serverAddressLock     sync.RWMutex
 	serverAddressProvider ServerAddressProvider
 
 	shutdown     bool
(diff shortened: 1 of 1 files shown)


jmurret (approved):
Thank you @moogacs !

gavraz:
Hey,
I came across this PR and noticed and noticed there is no invocation of serverAddressLock.Lock which suggests this has no effect since we only handle reads.

PR #594, opened 2024-05-11
Delete .gitignore

erosmary:


erosmary:
what CLA you mean your automated licences and it's distribution of massive
copy you've got to be kidding me
otherwise you should re analyse that creepy license with multiple owner

PR #597, opened 2024-06-06
Fix flaky raft tests, checking for candidate state

dhiaayachi:
In some tests in `raft_test.go` we have the following pattern:

```golang
// wait for the remaining follower to trigger an election
waitForState(follower, Candidate)
require.Equal(t, Candidate, follower.getState())
```
The issue with this pattern is that candidate state is a transient state and a node is not expected to stay in it for a long time, so by the time we finish `waitForState(follower, Candidate)` and get the state again the new node state could already have changed.


jrasell (approved):
LGTM!

PR #605, opened 2024-08-17
Fix pre-vote implementation where leader's pre-vote is rejected

### LINKED ISSUE #606: Follower rejecting leader's pre-vote
Hello, thanks for the library! We were trying out raft v1.7.0, and found that the pre-vote implementation has a bug where a follower rejects the pre-vote req from node X even when when the follower thinks that X is the leader.

This is the offending line. https://github.com/hashicorp/raft/blob/main/raft.go#L1752

```go
// Check if we have an existing leader [who's not the candidate] and also
var candidate ServerAddress // ISSUE: this is not set at all
candidateID := ServerID(req.ID)

// if the Servers list is empty that mean the cluster is very likely trying to bootstrap,
// Grant the vote
if len(r.configurations.latest.Servers) > 0 && !inConfiguration(r.configurations.latest, candidateID) {
	r.logger.Warn("rejecting pre-vote request since node is not in configuration",
		"from", candidate)
	return
}

if leaderAddr, leaderID := r.LeaderWithID(); leaderAddr != "" && leaderAddr != candidate { // ISSUE: so this will always return true as long as the follower has a leader
	r.logger.Warn("rejecting pre-vote request since we have a leader",
		"from", candidate,
		"leader", leaderAddr,
		"leader-id", string(leaderID))
	return
}
```

k-jingyang:
Fixes #606

### DIFF
--- a/raft.go
+++ b/raft.go
@@ -1749,7 +1749,7 @@ func (r *Raft) requestPreVote(rpc RPC, req *RequestPreVoteRequest) {
 	}()
 
 	// Check if we have an existing leader [who's not the candidate] and also
-	var candidate ServerAddress
+	candidate := r.trans.DecodePeer(req.GetRPCHeader().Addr)
 	candidateID := ServerID(req.ID)
 
 	// if the Servers list is empty that mean the cluster is very likely trying to bootstrap,
--- a/raft_test.go
+++ b/raft_test.go
@@ -3218,3 +3218,45 @@ func TestRaft_runFollower_ReloadTimeoutConfigs(t *testing.T) {
 	// Check the follower loop set the right state
 	require.Equal(t, Candidate, env.raft.getState())
 }
+
+func TestRaft_PreVote_ShouldNotRejectLeader(t *testing.T) {
+	// Make a cluster
+	c := MakeCluster(3, t, nil)
+	defer c.Close()
+	err := waitForLeader(c)
+	require.NoError(t, err)
+	leader := c.Leader()
+
+	// Wait until we have 2 followers
+	limit := time.Now().Add(c.longstopTimeout)
+	var followers []*Raft
+	for time.Now().Before(limit) && len(followers) != 2 {
+		c.WaitEvent(nil, c.conf.CommitTimeout)
+		followers = c.GetInState(Follower)
+	}
+	if len(followers) != 2 {
+		t.Fatalf("expected two followers: %v", followers)
+	}
+
+	// A follower who thinks that x is the leader should not reject x's pre-vote
+	follower := followers[0]
+	require.Equal(t, leader.localAddr, follower.leaderAddr)
+
+	reqPrevote := RequestPreVoteRequest{
+		RPCHeader:    leader.getRPCHeader(),
+		Term:         leader.getCurrentTerm() + 1,
+		LastLogIndex: leader.lastLogIndex,
+		LastLogTerm:  leader.getCurrentTerm(),
+	}
+
+	var resp RequestPreVoteResponse
+	leaderT := c.trans[c.IndexOf(leader)]
+	if err := leaderT.RequestPreVote(follower.localID, follower.localAddr, &reqPrevote, &resp); err != nil {
+		t.Fatalf("RequestPreVote RPC failed %v", err)
+	}
+
+	// the pre-vote should be granted
+	if !resp.Granted {
+		t.Fatalf("expected pre-vote to be granted, but it wasn't %+v", resp)
+	}
+}
--- a/raft_test.go
+++ b/raft_test.go
@@ -3242,7 +3242,7 @@ func TestRaft_PreVote_ShouldNotRejectLeader(t *testing.T) {
 	follower := followers[0]
 	require.Equal(t, leader.localAddr, follower.leaderAddr)
 
-	reqPrevote := RequestPreVoteRequest{
+	reqPreVote := RequestPreVoteRequest{
 		RPCHeader:    leader.getRPCHeader(),
 		Term:         leader.getCurrentTerm() + 1,
 		LastLogIndex: leader.lastLogIndex,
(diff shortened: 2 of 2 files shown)


dhiaayachi:
Hey @k-jingyang 
Thank you for reporting this and providing a fix. I will go ahead and merge this once the CI pass.

Just making sure I fully grasp the impact of this bug. Before this fix, a leader who loses leadership, while pre-vote is on, will not be able to be reelected and leadership will likely transfer to another node but this won't introduce any extra scenarios where leadership become unstable or no leader is able to be elected.

k-jingyang:
> Hey @k-jingyang Thank you for reporting this and providing a fix. I will go ahead and merge this once the CI pass.
> 
> Just making sure I fully grasp the impact of this bug. Before this fix, a leader who loses leadership, while pre-vote is on, will not be able to be reelected and leadership will likely transfer to another node but this won't introduce any extra scenarios where leadership become unstable or no leader is able to be elected.

Thanks! I realised that the CI pipeline is failing, I tried to run `make test` and it failed even on the current master. Not sure if I'm missing something here.

Yes, you are correct. However, the situation becomes more serious in the following scenario:
- 3 node cluster
- 1 node is down, 2 nodes are still up (node A leader, node B)
- node A somehow think that it lost leadership 
- node B can still heartbeat to node A (so it doesn't call for an election)
- node A tries to call for an election again (but pre-vote prevents it from doing so, as node B keeps rejecting it as per the bug)

This was what we encountered. I didn't manage to dig into how node A thought that it lost leadership.

So even though a majority in the cluster (2 out of 3) is still up, there are no actual leaders

dhiaayachi:
@k-jingyang 
I'm looking into the tests, it seems that tests are broken for some reason not related to this PR.
> Yes, you are correct. However, the situation becomes more serious in the following scenario:
> 
> * 3 node cluster
> * 1 node is down, 2 nodes are still up (node A leader, node B)
> * node A somehow think that it lost leadership
> * node B can still heartbeat to node A, so it doesn't call for an election)
> * node A tries to call for an election again, but pre-vote prevents it from doing so, as node B keeps rejecting it as per the bug
> 
> This was what we encountered. I didn't manage to dig into how node A thought that it lost leadership.
> 
> So even though a majority in the cluster (2 out of 3) is still up, there are no actual leaders

I'm not sure I understand? If `node A` think it's not a leader anymore it should stop sending heartbeats to all the other nodes which will lead to `node B` timing out the heartbeat and becoming a Candidate and eventually become a leader. `Node B` being a follower should not be sending heartbeats to `node A`,  Am I missing something?

k-jingyang:
> @k-jingyang I'm looking into the tests, it seems that tests are broken for some reason not related to this PR.
> 
> > Yes, you are correct. However, the situation becomes more serious in the following scenario:
> > 
> > * 3 node cluster
> > * 1 node is down, 2 nodes are still up (node A leader, node B)
> > * node A somehow think that it lost leadership
> > * node B can still heartbeat to node A, so it doesn't call for an election)
> > * node A tries to call for an election again, but pre-vote prevents it from doing so, as node B keeps rejecting it as per the bug
> > 
> > This was what we encountered. I didn't manage to dig into how node A thought that it lost leadership.
> > So even though a majority in the cluster (2 out of 3) is still up, there are no actual leaders
> 
> I'm not sure I understand? If `node A` think it's not a leader anymore it should stop sending heartbeats to all the other nodes which will lead to `node B` timing out the heartbeat and becoming a Candidate and eventually become a leader. `Node B` being a follower should not be sending heartbeats to `node A`, Am I missing something?

I believe you're right. My understanding of the issue could be slightly wrong. Let me dig into the logs to see why node B still thinks node A was the leader even though node A is a Candidate.

k-jingyang:
@dhiaayachi,

After digging through the logs, I don't have a definitive conclusion. You're right that only the leader should be heartbeat-ing to the follower, and once node A lost leadership, it should not heartbeat anymore.

What I can be sure of:
1. node B rejected node A's pre-vote because it still thinks node A was the leader (the bug)
```bash
# 1 of our log line
raft: rejecting pre-vote request since we have a leader from:"" leader:"NODE_A_IP" leader-id:NODE_A
```
2. Even as a Candidate, node A has logs regarding failed heartbeat to the downed node for a while, then it stopped
- Since we don't seem to log successful heartbeats and heartbeat metrics has expired from our telemetry systems. I'm not sure if this meant that successful heartbeats were still ongoing to node B or not (even when node A is a Candidate). If so, this would explain why node B did not trigger an election. And if so, this would also indicate some form of bug or edge case

raft.go
```diff
@@ -1749,7 +1749,7 @@ func (r *Raft) requestPreVote(rpc RPC, req *RequestPreVoteRequest) {
 	}()
 
 	// Check if we have an existing leader [who's not the candidate] and also
-	var candidate ServerAddress
+	candidate := r.trans.DecodePeer(req.GetRPCHeader().Addr)
```
banks, line 1752: 🤔 In the `requestVote` method above we do this:

```
	var candidate ServerAddress
	var candidateBytes []byte
	if len(req.RPCHeader.Addr) > 0 {
		candidate = r.trans.DecodePeer(req.RPCHeader.Addr)
		candidateBytes = req.RPCHeader.Addr
	} else {
		candidate = r.trans.DecodePeer(req.Candidate)
		candidateBytes = req.Candidate
	}
```

As far as I can tell that's the same except that we store the raw bytes for some use that we don't need here, and that for some reason we account for the case where the RPCHeader address is empty and fall back to the candidate in the message.

TBH I have so many questions about this bit of logic that I don't recall answers to if I ever did know:
 * Why do we prefer the candidate in the RPC header to the one in the message if there is always one in the message?
 * In which cases may the RPCHeader not contain an address? Maybe in-memory type transports that don't have real underlying network connections don't set that or something?

I've not had time to chase down the history on those though. If anyone else does please leave your findings here and we can understand better. But my hunch is that either the above code is redundant now and we should change both, or we should keep the same logic here as in `requestVote` (minus the unused `candidateBytes` parts).

banks (commented):
@k-jingyang thanks for helping us understand this. I also think this fix is worth merging and should be possible soon, but it's really useful to us to understand the impact of the issue it might cause to help us decide how urgently we need to get the fix tested and merged into our products etc!

In your logs above how far in wall-clock time apart are the t=1 and t=2 logs? The old leader when it switches to candidate state would shut down replication goroutines which are the things sending the heartbeats and logging that error. If they are milliseconds apart I suspect that this is just a timing thing and the replication threads just happened to try sending a heartbeat around the time the leader changed, failed but the error didn't get logged until after the switch to candidate. If there are many seconds passing while the old node is a candidate but still attempting multiple times to send heartbeats then yep that sounds more concerning. Would you be able to post your whole logs either here or privately if we provide an email address?

raft.go
k-jingyang: To add on, while implementing this change. I saw that Candidate in RequestVoteRequest is deprecated.

```go
// RequestVoteRequest is the command used by a candidate to ask a Raft peer
// for a vote in an election.
type RequestVoteRequest struct {
	RPCHeader

	// Provide the term and our id
	Term uint64

	// Deprecated: use RPCHeader.Addr instead
	Candidate []byte
```

k-jingyang:
@banks 

The time between when the node became a candidate and when it failed to heartbeat were more than 10s apart.

raft.go
dhiaayachi: @k-jingyang they are deprecated but we nee do keep them for the time being to ensure retro compatibility with older versions of raft and allow a smooth upgrade from a software using an older version of raft to the new version.
I think we will need to update the fix to follow the same thing implemented in `requestVote`

```go
        var candidate ServerAddress
	if len(req.RPCHeader.Addr) > 0 {
		candidate = r.trans.DecodePeer(req.RPCHeader.Addr)
	} else {
		candidate = r.trans.DecodePeer(req.Candidate)
	}
```

raft_test.go
```diff
@@ -3270,3 +3270,45 @@ func TestRaft_runFollower_ReloadTimeoutConfigs(t *testing.T) {
 	// Check the follower loop set the right state
 	require.Equal(t, Candidate, env.raft.getState())
 }
+
+func TestRaft_PreVote_ShouldNotRejectLeader(t *testing.T) {
```
dhiaayachi, line 3274: maybe we can add a similar scenario in here that test the opposite: `TestRaft_PreVote_ShouldRejectNonLeader`

dhiaayachi (changes requested):
Thank you again @k-jingyang! The tests should be better now if you rebase the PR on top of the latest changes from `main`. I left few suggestions.

raft.go
k-jingyang: @dhiaayachi RequestPreVoteRequest does not contain any other field to indicate the addr of the peer (e.g. Candidate). I don't think we need anything other than the Addr from the RPCHeader.

```go
// RequestPreVoteRequest is the command used by a candidate to ask a Raft peer
// for a vote in an election.
type RequestPreVoteRequest struct {
	RPCHeader

	// Provide the term and our id
	Term uint64

	// Used to ensure safety
	LastLogIndex uint64
	LastLogTerm  uint64
}

```

k-jingyang:
@dhiaayachi, you're welcome! I've addressed your comments! :smile:

raft.go
dhiaayachi: yes that's right! Sorry for the confusion 🤦 

@banks I think we are safe here because prevote request type only exist in the new version and an older version of raft (prior to prevote) won't process it anyway. The sender would count that as a prevote granted and carry on with the actual election.

dhiaayachi (approved):
LGTM! Thank you for the contribution @k-jingyang. 

I will wait for @banks to resolve the remaining conversation and merge this!

k-jingyang:
Hey @banks, would it be possible to provide an email address so that I can share an investigation doc?

banks:
Hey, You can use `consul-oss-debug@` our corporate domain. Let me know if
you have any issues I checked that is still a thing but it's not been used
for a little while here. In case anyone else comes across this message,
this email alias is not monitored generally and is only used when Hashicorp
Engineers specifically have requested something that is to private to share
in a public GH issue or PR.

PR #608, opened 2024-08-22
use current github.com/hashicorp/go-metrics

nlacey:
github.com/armon/go-metrics no longer exists.

When you utilize nats-streaming-server, you can get an error 

github.com/nats-io/nats-streaming-server/server imports
        github.com/hashicorp/raft imports
        github.com/armon/go-metrics: github.com/armon/go-metrics@v0.5.3: parsing go.mod:
        module declares its path as: github.com/hashicorp/go-metrics
                but was required as: github.com/armon/go-metrics

There is a chain of dependencies to be fixed, but I believe this is the first required fix.

steps to re-produce
replace github.com/armon/go-metrics  with github.com/hashicorp/go-metrics within the source code.
than do a 
go mod tidy within top level and fuzzy packages,   
didn't touch raft-compat (unsure of it's purpose)

thanks for consideration. 

----
FYI for everyone else in case this PR doesn't get excepted.

I was able to fix my go mod tidy problem by adding to the bottom of my go.mod file

replace github.com/armon/go-metrics => github.com/hashicorp/go-metrics v0.5.3


dhiaayachi:
👋 @nlacey 
Thank you for the contribution, unfortunately this is not a change we can accept for the time being as it will have downfalls for other Hashicorp libraries/products, probably the same issue you are facing with nats. `go-metrics` uses a global state and init() which make it impossible to mix `github.com/armon/go-metrics` and `github.com/hashicorp/go-metrics` as dependencies in the same dependencies tree of a library/product. 

We need to internally synchronize effort to be able to perform that migration safely, and until that is done, we have decided to keep using `github.com/armon/go-metrics` for the time being.

I will close this, but feel free to comment or open a new issue if you have more questions about this.

Issue #549, opened 2023-03-15
Enhancement: persist commit index in LogStore to accelerate recovery


banks:
Today I realised my mental model of how our Raft library works is subtly wrong.

I assumed that when a node restarts, it loads the most recent snapshot into the FSM _and replays committed logs_ from disk into the FSM before rejoining the cluster.

In fact, we only load from snapshot since currently we don't persist any information about which logs are known to be committed which means that replaying them all might apply a few uncommitted ones from the end of the log.

In practice this works OK in most cases because if the node is elected leader then it will write a new log which as soon as it's committed will trigger applying all the other committed logs in to the FSM too. If it's a follower it doesn't catch up it's FSM until the first `AppendEntries` RPC from the leader tells it the current commit index in the cluster.

As we consider reducing snapshot frequency with WAL LogStore being able to store lots more logs performantly, this will become more of an issue because it will mean followers are much more "stale" when they start serving requests.

To fix this correctly, we'd need to persist the last known commit index so that we know which prefix of the logs it's safe to apply on startup. We could periodically persist that to the `StableStore` but that increases amount of disk transactions and trades that off against how recent the index stays. Currently StableStore performance is not an issue because it's only updated infrequently during elections. This would make it much more critical.

An alternative that would be more efficient but a little more complicate would be:
 1. Introduce a new optional interface called something like:
     ```go
     type CommitTrackingLogStore interface {
         SetCommitIndex(idx uint64) error
         ReadCommitIndex() (uint64, error)
     }
     ```
 2. Have Raft detect that the `LogStore` also implements this just before calling `StoreLogs` during replication or leader's `dispatchLogs`, if it does, call `SetCommitIndex` first with the most recent commitIndex we know
 3. Implementations may store that index in memory until the next `StoreLogs` call and then write it out on disk along with the log data.
 4. When the `LogStore` recovers it's state on open, it would recover the last persisted commit index to and load it into memory
 5. When Raft starts up it can restore snapshot like now then...
 6. Check if the `LogStore` implements `CommitTrackingLogStore`. If it does, load the commit index from the store and replay all logs from `lastSnapshotIndex` up to and including the commit index that was persisted, before completing startup.

otoolep:
This would be very useful, as rqlite (which uses this library) can also encounter long start up times. Knowing which logs, which are not contained in the latest snapshot, are actually committed, would solve this problem. I could optimize replay at start-up.

[...]

PR #613, opened 2024-09-01
Enhancement: persist commit index in LogStore to accelerate recovery

lalalalatt:
https://github.com/hashicorp/raft/issues/549


lalalalatt:
@banks This is first PR for https://github.com/hashicorp/raft/issues/549.
Thanks for spending your time to review this~

lalalalatt:
Purpose for next PR:
**Persist commit index every time `processLogs` is called.**
https://github.com/hashicorp/raft/blob/42d34464b2d203e389e11ed6d43db698792c0604/raft.go#L1292-L1359
```diff
func (r *Raft) processLogs(index uint64, futures map[uint64]*logFuture) { 
 	// Reject logs we've applied already 
 	lastApplied := r.getLastApplied() 
 	if index <= lastApplied { 
 		r.logger.Warn("skipping application of old log", "index", index) 
 		return 
 	} 

+       if r.fastRecovery && isCommitTrackingLogStore(r.logs) {
+               store := r.logs.(CommitTrackingLogStore)
+               if err = store.SetCommitIndex(index) {
+                       // show some error msg
+               }
+       }
  
 	....
  
 	// Update the lastApplied index and term 
 	r.setLastApplied(index) 
 }
 ```

**Comment by otoolep** [comment]
As a long-term user of this library, this could be useful. However I would strongly recommend that this functionality be wrapped in a flag, which is settable in the Raft Config object (similar to `NoSnapshotRestoreOnStart`), and be disabled by default. Systems built on Raft need to be solid, so taking a conservative approach is warranted.

**Comment by lalalalatt** [comment]
> As a long-term user of this library, this could be useful. However I would strongly recommend that this functionality be wrapped in a flag, which is settable in the Raft Config object (similar to `NoSnapshotRestoreOnStart`), and be disabled by default. Systems built on Raft need to be solid, so taking a conservative approach is warranted.

Thanks for the suggestion, I would add `r.fastRecovery` as the feature flag~

**Review by banks — COMMENTED** [review]
Thanks for starting on this @lalalalatt.

The interface looks good. I have a comment inline and given that I'm suggesting removing half the lines in this PR, it might make sense to also add the change you proposed for your "next PR" into this one?

I also think it might be worth rebasing this PR on to a feature branch like `f/fast-recovery` so we can keep PRs small to review but not merge code into `main` until it's a working feature that can be released. (I realise this would be a no-op, but still if we don't complete the feature it will be unused code that will need later cleanup or completion.)

What do you think of that approach?

File: log.go [hunk 1]
```diff
@@ -190,3 +190,13 @@ func emitLogStoreMetrics(s LogStore, prefix []string, interval time.Duration, st
...
+
+type CommitTrackingLogStore interface {
+	SetCommitIndex(idx uint64) error
+	ReadCommitIndex() (uint64, error)
+}
+
+func isCommitTrackingLogStore(s LogStore) bool {
+	_, ok := s.(CommitTrackingLogStore)
+	return ok
+}
```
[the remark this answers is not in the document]
otoolep: I agree, wrapping interface assertions is not a style I've come across in Go.

lalalalatt:
> I also think it might be worth rebasing this PR on to a feature branch like `f/fast-recovery` so we can keep PRs small to review but not merge code into `main` until it's a working feature that can be released. (I realise this would be a no-op, but still if we don't complete the feature it will be unused code that will need later cleanup or completion.)

@banks Sure, that sounds like a good plan. Could you help me create the `f/fast-recovery` branch? Thanks!

log.go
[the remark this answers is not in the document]
lalalalatt: @banks @otoolep Thanks for the advice! I agree that directly using the type assertion at the call site is more idiomatic in Go. I’ll go ahead and remove that function.

otoolep:
Drive-by comment.

>@banks Sure, that sounds like a good plan. Could you help me create the f/fast-recovery branch? Thanks!

That's not the right way to think about the development flow (not unless this repo is managed in some unusual way). You create the branch in your own fork of this repo, and then generate a PR from that branch in your fork to the main branch in this repo.

banks:
@otoolep in general that is usually how GH works. In this case I wonder if we should have a long running branch here so that we can keep PRs small and review the changes in small parts rather than wait until the entire feature is built and have to review it all in one huge PR from the fork to here.

One way to do that would be for me to create a long-lived feature branch here, another would be for us to review PRs in a fork but that leaves all the interim review in a separate place 🤔 .

On reflection. I'm not sure what is gained by trying to make this many small PRs. Let's continue work here and see how large this PR gets as more of the code is built before we worry too much about making feature branches. Sorry for the suggestion  that wasn't very well thought through!

lalalalatt:
> On reflection. I'm not sure what is gained by trying to make this many small PRs yet. Let's continue work here and see how large this PR gets as more of the code is built before we worry too much about making feature branches. Sorry for the suggestion that wasn't very well thought through!

Ok, looks good!

api.go
```diff
@@ -697,6 +704,29 @@ func (r *Raft) tryRestoreSingleSnapshot(snapshot *SnapshotMeta) bool {
...
+	}
+	commitIndex, err := store.GetCommitIndex()
+	if err != nil {
+		return fmt.Errorf("failed to read commit index from store: %w", err)
+	}
+	if commitIndex == 0 {
+		return nil
+	}
+
+	r.processLogs(commitIndex, nil)
```
banks, line 726:
```suggestion
        // Update the in-memory state now we know what is committed on the leader
        r.setCommitIndex(commitIndex)
        // Apply all the known-committed logs to the FSM before we complete recovery.
	r.processLogs(commitIndex, nil)
```

api.go
```diff
@@ -606,6 +611,8 @@ func NewRaft(conf *Config, fsm FSM, logs LogStore, stable StableStore, snaps Sna
 	// to be called concurrently with a blocking RPC.
 	trans.SetHeartbeatHandler(r.processHeartbeat)
 
+	r.recoverFromCommitedLogs()
```
banks, line 614: There is a loop above that starts with:

```go
         // Scan through the log for any configuration change entries.
	snapshotIndex, _ := r.getLastSnapshot()
	for index := snapshotIndex + 1; index <= lastLog.Index; index++ {
```
(which GH doesn't let me comment on directly 🙄 ) that I wonder if we should incorporate somehow 🤔 .

It was already iterating the whole log to see if any logs are config changes so it would be nice to incorporate into the same loop somehow if we are restoring rather than read and deserialise every log entry twice. In fact `processLogs` will already do this config stuff to if we do call `recoverFromCommittedLogs`!

Also: 

```suggestion
	r.recoverFromCommittedLogs()
```
nit: but if we keep this name we should correct the typo in committed.

----

Finally, in writing the above I've realised that he config change loop in the existing code is possibly wrong and maybe even violate's Raft's correctness 😱 . I need to think about it some more but the whole reason we are not able to process all the logs through the FSM yet is because we don't know for sure if there were committed, yet in the loop above we are just blindly calling `processConfigurationLogEntry` on anything we find which assumes the log is committed and sets `lastCommittedConfiguration` to whatever you pass in!

I'm going to create a separate issue for this. Let's leave the behaviour unchanged for now here when not using fast recovery at least.

raft.go
```diff
@@ -1354,6 +1354,8 @@ func (r *Raft) processLogs(index uint64, futures map[uint64]*logFuture) {
 		applyBatch(batch)
 	}
 
+	r.persistCommitIndex(index)
+
```
banks, line 1358: I don't think this is the right place for this call. I think we should call it in `storeLogs` just before the actual call to `StoreLogs` on the storage since that is the write we want our log stores to be able to add this state to disk in. Calling it here will not do anything until the next StoreLogs call at least in the design I proposed in the issue.

banks (changes requested):
Hey @lalalalatt, thanks for working on this.

I spotted a couple of things that we should tweak here. I've added comments inline - let me know if anything is confusing. I didn't do a "full" final review yet but these should get it more inline with my original design. Feel free to explain if I'm misunderstanding anything though!

api.go
banks: On the last part. I spent a while thinking about the correctness of this and I think it's OK as it is 😅 .

We implement the "simplified" membership change protocol form the thesis (chapter 4) where only single node addition or removal are allowed at once and where we only allow a single uncommitted membership change to be in flight. Under these invariants, we don't need "joint consensus" because every quorum in both old and new configurations must still overlap. That means that in normal case, servers adopt new configurations immediately even before they are known to be committed and this is safe by that given logic:

>  If servers adopted Cnew only when they
learned that Cnew was committed, Raft leaders would have a difficult time knowing when a majority of the old cluster had adopted it. They would need to track which servers know of the entry’s
commitment, and the servers would need to persist their commit index to disk; neither of these
mechanisms is required in Raft. Instead, each server adopts Cnew as soon as that entry exists in its
log, and the leader knows it’s safe to allow further configuration changes as soon as the Cnew entry
has been committed. Unfortunately, this decision does imply that a log entry for a configuration
change can be removed (if leadership changes); in this case, a server must be prepared to fall back
to the previous configuration in its log.

Fun that it specifically mentions how servers do not need to persist commit index in the context of this PR 😄.

That is true for overall linearizable correctness, however I still think this is a useful addition. The raft thesis only (really)considers fully linearizable reads where all reads go through the leader and so there is no downside to a non-leader "accepting" requests before it's fully recovered since it's only handling for requests would be to redirect or forward to the leader. Raft already ensures that a leader doesn't serve requests until all committed logs have been applied. In systems such as Consul and Vault where "stale" reads are explicitly allowed in some cases to increase read scalability, it's certainly not ideal to have a period where a follower has started up and is accepting reads but has not applied all logs it has on it's disk even when it knew they were committed previously.

tl;dr, we're fine here 😄, but we should be careful to preserve the behaviour - even if we only replay up to `commitIndex` on the FSM, we _should_ still process any config changes in the rest of the log after that!

peterxcli: @banks Thanks for these lots of comments 😄

peterxcli: But,  for the current change, it is possible to apply same configurations twice, although its wouldn't violate the correctness, it still be wasteful. 
So I would reverse the order of `r.recoverFromCommitedLogs()` and "iterating the whole log to see if any logs are config changes", then, if the persisted commit are applied, then we only need to "iterating the ~~whole log~~ `r.logs[max(lastapplied, snapshotIndex) + 1 : ]` to see if any logs are config changes" and that prevent the duplication~

peterxcli:
> tl;dr, we're fine here 😅 (on the correctness issue, the other comments about this PR stand). We should be careful to preserve the behaviour in this PR - even if we only replay up to commitIndex on the FSM, we should still process any config changes in the rest of the log after that!

@banks Thanks for these lots of comments 😄
But, for the current change, it is possible to apply same configurations twice, although its wouldn't violate the correctness, it still be wasteful.
So I would reverse the order of r.recoverFromCommitedLogs() and "iterating the whole log to see if any logs are config changes", then, if the persisted commit are applied, then we only need to "iterating the whole log r.logs[max(lastapplied, snapshotIndex) + 1 : ] to see if any logs are config changes" and that prevent the duplication~

peterxcli:
> I don't think this is the right place for this call. I think we should call it in `storeLogs` just before the actual call to `StoreLogs` on the storage since that is the write we want our log stores to be able to add this state to disk in. Calling it here will not do anything until the next StoreLogs call at least in the design I proposed in the issue.

Isn't the

peterxcli:
> I don't think this is the right place for this call. I think we should call it in `storeLogs` just before the actual call to `StoreLogs` on the storage since that is the write we want our log stores to be able to add this state to disk in. Calling it here will not do anything until the next StoreLogs call at least in the design I proposed in the issue.

I don't think storing commit index every time before `StoreLogs` is a good idea. 

For followers, when they are appending the log received from leader, the commit index may not increase, because leader may not receive majority agreement on the corresponding logs. 

For leader, as I mention previously, it persist logs by calling `log.StoreLogs()` when receiving new log from applyCh which client requested, but that doesn't mean the logs are committed, unless majority accept them.

My idea is: the commit index only update at two place:
1. receive higher `leaderCommit` when member is follower
2. leader's commitment detect majority have accepted a higher log index, then it would send a message to `c.commitCh`, then the leader main thread would receive that new commit index and actually update it in raft in-memory state.

They both trigger `r.processLogs` eventually, so we can just update the commit index there.

Maybe that's my misunderstanding 😅, if there is anything wrong, please correct me~

banks:
Hi @peterxcli 

The idea behind persisting commit index during StoreLogs is to make sure it doesn't cause worse performance. Writing to disk is the slowest part of raft in most implementations.

If we add a new write to disk anywhere then performance will decrease and that's not acceptable for our products at least and probably not to other users of this library.

So the whole design was intended to persist the _current known_ commit index along with the logs every time we write new logs - yes it's a few extra bytes but it's the `fsync` calls that are slow so if we can write in the same write it's virtually "free" for all the LogStore implemetations we use currently and doesn't impact performance.

> For followers, when they are appending the log received from leader, the commit index may not increase, because leader may not receive majority agreement on the corresponding logs.

This is true, but the `AppendEntries` RPC is _the_ method that a follower learns about _previous_ commits. In a given `AppendEntries` RPC, there will be one or more new logs (ignoring heartbeats for now), lets say those are logs `{1000, 1001}` and the leader will also include `CommitIndex = 998` which lets this follower know that everything up to 998 is comitted. My proposal is that we call `store.SetCommitIndex()` with the commit index the leader just sent in each AppenEntries RPC then we always have persisted the absolute most up to day commit info each follower knows, and it didn't cost any additional disk fsyncs.

If we only write this when we call `processLogs` on the follower as you propose then there are two possibilities depending on how the LogStore implements it:

 1. The `LogStore` buffers that index in memory until the next call to `StoreLogs` - that would _work_ and have the same performance as my proposal, but I think it's slightly worse than my proposal for two reasons:
    1. If commitIndex sent by the leader is not actually persisted until the _next_ RPC arrives which means that whenever the node restarts it can only restore a smaller number of logs even though it did "know" about and possibly apply more logs than it's persisted commitIndex before it restarted. This difference is likely small, but it's easy to avoid this.
    2. I find it much more surprising and hard to reason about if we actually don't persist the `commitIndex` we learned from the leader until the _next_ RPC (where we'll learn about a new commit index most likely).
 2. Or, the `LogStore` could write the `commitIndex` to disk synchronously when `SetCommitIndex` is called. This would be a whole extra fsync (or two for BoltDB or WAL implementations) which would likely cause a significant slow down in the speed we can write to raft. I don't think we should consider this option especially when there's an easy way to avoid it (my proposal).

> For leader, as I mention previously, it persist logs by calling log.StoreLogs() when receiving new log from applyCh which client requested, but that doesn't mean the logs are committed, unless majority accept them.

You're correct that all the logs we store when we call `StoreLogs` on the leader are not yet committed, in fact, with our current implementation we _guarantee_ that none of them are committed (except in the trivial case of a single node cluster) because we don't replicate to followers until after this. But again the leader would not be calling `SetCommitIndex` with the _new_ log index, it would just be calling it with whatever it's current in-memory `commitIndex` is which is guaranteed to be some way behind.

Again this is so that we can "re-use" the fsync on the `LogStore` and so add more disk IO or waiting on the critical path of writing data.

Some of the same arguments as above apply here: if we wait until `commitIndex` updates and `processLogs` is called on the leader we'll either have a more confusing model where the commitIndex is only persisted by the commit of the batch of logs _after_ the one we are storing, or we'll make it all slower by adding a new sync disk write just to persist this state.

If we were OK with adding more disk writes, I'd not have proposed this as an extension of `LogStore` at all - we already have a `StableStore` that we could just add a new KV pair to, but if we did that and updated it on every commit it would massively impact the throughput of a raft cluster since disk writes are the slowest part and we'd effectiely double the number of them on both leader and follower for every commit.

Does that make sense?

peterxcli:
@banks oh~~ It seems like I completely misunderstood your design.

> Again this is so that we can "re-use" the fsync on the `LogStore` and so avoid more disk IO or waiting on the critical path of writing data.

Got it, thanks~ 😄

> As an aside, it we were OK with adding more disk writes, I'd not have proposed this as an extension of `LogStore` at all - we already have a `StableStore` that we could just add a new KV pair to. But if we did that and updated it on every commit it would massively impact the throughput of a raft cluster since disk writes (and fsyncs) are the slowest part and we'd effectively double the number of them on both leader and follower for every commit.

That's true 😅

@banks Thanks for your clarification very much~

peterxcli:
Hi @banks,

I’ve just added some tests for the fast recovery feature, and everything seems to be working well. I’d really appreciate any feedback or advice you might have.

Thank you!

banks (changes requested):
Hey Peter, 

This is looking awesome, thanks for your hard work!

I have a couple of suggestions inline. I'm not super hard line on any of them if you think there's a good reason to reject so love to hear your thoughts but I'd lean towards them being an improvement that will help us maintain or reason about this in the future!

raft.go
```diff
@@ -1535,6 +1552,9 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 		}
 
 		if n := len(newEntries); n > 0 {
+			commitIndex := r.getCommitIndex()
```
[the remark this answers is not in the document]
peterxcli: Agree.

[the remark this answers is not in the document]
lalalalatt: I think 2b is great too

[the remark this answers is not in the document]
lalalalatt: @banks Thanks for the detailed breakdown! I totally agree that your approach is better for preventing the commit index lag in commitTrackingStore. I also really like option 2b—it feels cleaner and makes things more explicit~

peterxcli:
@banks Thanks for the advice of the stage commit index. 
I have make some changes on it.
Would like to see your further opinion on it, thx~

raft_test.go
```diff
@@ -1095,6 +1095,180 @@ func TestRaft_RestoreSnapshotOnStartup_Monotonic(t *testing.T) {
...
+	}
+
+	// Expect: snap.Index --- commitIdx --- lastIdx
+	lastIdx, err := r.logs.LastIndex()
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+	assert.LessOrEqual(t, snap.Index, commitIdx)
+	assert.LessOrEqual(t, commitIdx, lastIdx)
+}
```
banks, line 1270: I'm assuming these tests are failing right now because the InmemStore wasn't updated to match the new interface right? If they are passing for you then it might be worth a look!

banks (dismissed):
This is looking awesome Peter!

Thanks again. There is one suggested minor improvement in being more defensive which is maybe questionable but I think just about seems worth it to me. There's also a typo in the new interface name and the testing implementation didn't update with it so a couple of minor fixes and it's hopefully good to go!

I think once we have those fixed up, I'll try and get some other reviewers internally to look before we make a merge decision. We'll may also look to get the BoltDB and WAL implementations PRed so we can test the whole thing end-to-end before we merge any of these to be totally sure it actually works as expected.

inmem_store.go
```diff
@@ -131,3 +132,30 @@ func (i *InmemStore) GetUint64(key []byte) (uint64, error) {
...
+// NewInmemCommitTrackingStore returns a new in-memory backend that tracks the commit index. Do not ever
+// use for production. Only for testing.
+func NewInmemCommitTrackingStore() *InmemCommitTrackingStore {
+	i := &InmemCommitTrackingStore{
+		InmemStore: *NewInmemStore(),
+	}
+	return i
+}
+
+func (i *InmemCommitTrackingStore) SetCommitIndex(index uint64) error {
```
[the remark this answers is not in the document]
lalalalatt: Oh great, I would also change the `NewInmemCommitTrackingStore` to return the `CommitTrackingStore` interface to prevent this kind of mistake😁

raft_test.go
lalalalatt: It passes😱

lalalalatt: Hi @banks 
First I want to thanks that you reviewed so many changes and gave lots of advices and opinions. 😁

---

Back to this topic, I think this test case is ok🤔. 

Because basically the commit index stored in the store would always lower than the last log index in store (there is always one `StoreLogs` call lag).

As the comment "Expect: snap.Index --- commitIdx --- lastIdx" leaved, I think we can't sure what is the exact position of the `commitIndex` in every test, so I just test the interval only.
But now we find this would lead to another problem - we can't even detect the commit store interface and its implementation aren't match😱.
We can simply solve that detection issue by changing `assert.LessOrEqual ` to `assert.Less`, but that would another flaky problem because of the uncertainty of the commit index.

What do you think?

inmem_store.go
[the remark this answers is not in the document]
banks: returning interfaces is sometimes a bit of a pain because then you have to type assert if you need to use any method not in the interface. It might be fine.

The other common idiom for catching this kind of issues at compile time is to add this somewhere (usaully at top of file with constants):

```go

var _ CommitTrackingStore = &InmemCommitTrackingStore{}

```

That will be a compile error if the interface isn't implemented but is a no-op and doesn't actually allocate anything in real code because the result is unused.

raft_test.go
banks: 🤔 I think the assertion is fine as it is, the real issue seems to be the line above:

```
commitIdx, err := r.logs.(CommitTrackingLogStore).GetCommitIndex()
```

I'm not sure how that didn't just panic if the log store actually didn't implement the interface 🤷 .

I don't think we need to change the assertions.

raft_test.go
```diff
@@ -1095,6 +1095,184 @@ func TestRaft_RestoreSnapshotOnStartup_Monotonic(t *testing.T) {
...
+	r, err = NewRaft(&cfg, r.fsm, r.logs, r.stable, r.snapshots, trans2)
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+	c.rafts[0] = r
+
+	store, ok := r.logs.(CommitTrackingLogStore)
+	if !ok {
+		t.Fatal("err: raft log store does not implement CommitTrackingLogStore interface")
+	}
```
banks, line 1260: 🤔 This is fine, but I'd probably have left it like it was before. The behaviour of an unchecked type assert like we had before (i.e. no `ok` assignment) would be to panic with a message that's basically the same as the error here. That should have been sufficient to fail the test anyway. The most puzzling thing to me is that it didn't. Perhaps you didn't actually run this locally before the interface implementation was fixed and then it was masked by a different error in CI (which is sadly often flaky).

We can leave it like this for now though I guess at least it doesn't stop the entire test run which is arguably nicer although something that was a programming mistake that would never work is fine to panic on in tests IMO.

banks (commented):
Nice thanks @peterxcli for all the work on this so far. I think it's looking good. I'm going to ask some other teams here to take a look before we make a final decision on whether this can be merged or whether we feel like there are potential risks or issues that still need mitigating.

schmichael (commented):
I am more of a consumer of the raft library (via Nomad) than a contributor, so please take my comments with a grain of salt! I don't have nearly the context on this change and the related issue as others!

On the Nomad side we've long hoped to solve the staleness-during-recovery aspect of this issue by using autopilot's health callback: https://github.com/hashicorp/nomad/issues/15560#issuecomment-1766894103

AFAICT these two approach would be *complimentary:* this PR would speed up recovery, while the other linked idea would allow the application layer to forward stale queries on until a restoring server is healthy.

The only potential for conflict I can see is that if NewRaft blocks for substantially longer, Nomad would probably prefer the linked approach alone: this would allow log recovery to be asynchronous with the server's other work. It could serve RPCs (even if it just forwards them to the leader until recovery completes).

That being said obviously it's up to Nomad to make that judgement call (e.g. defining what "substantially longer" even means!) and doesn't need to block this improvement.

---

My only other concern is the API contract this creates seems extremely brittle:

> The implementation MUST NOT persist this value separately from the log entries.

Would it be safer to just expect log stores to implement a new store log entrypoint since we're having to type assert on every StoreLog call anyway?

```go
type StoreLogsTx struct {
	CommittedIndex uint64
	Logs []*Log
}

type CommitTrackingLogStore interface {
	LogStore

	StoreLogsTx(tx StoreLogsTx) error
	GetCommitIndex() (uint64, error)
}
```

Ignoring the terrible naming of `StoreLogsTx`, this API seems like it models the actual atomicity fast recovery requires.

Using a struct would let us add more state to atomic log transactions in the future without API breakages, but I don't know if that's a desirable feature. I'm just borrowing patterns I've used elsewhere to minimize API compat breaks.

To avoid type assertions on every StoreLogs call, you could even use an adapter to turn every LogStore into a CommitTrackingLogStore that noops the commit index handling.

Just an idea! Sorry if this is all noise and nonsense.

api.go
```diff
@@ -697,6 +705,39 @@ func (r *Raft) tryRestoreSingleSnapshot(snapshot *SnapshotMeta) bool {
...
+	if err != nil {
+		r.logger.Error("failed to get commit index from store", "error", err)
+		panic(err)
+	}
+
+	lastIndex, err := r.logs.LastIndex()
+	if err != nil {
+		r.logger.Error("failed to get last log index from store", "error", err)
+		panic(err)
+	}
```
dhiaayachi, line 731: Shouldn't we fallback to the non fast recovery path instead of panic?

log.go
```diff
@@ -190,3 +190,24 @@ func emitLogStoreMetrics(s LogStore, prefix []string, interval time.Duration, st
...
+	// The implementation MUST NOT persist this value separately from the log entries.
+	// Instead, it should stage the value to be written atomically with the next
+	// StoreLogs call.
+	//
+	// GetCommitIndex MUST never return a value higher than the last index in the log,
+	// even if a higher value has been staged with this method.
+	//
+	// idx is the new commit index to stage.
+	StageCommitIndex(idx uint64) error
+	GetCommitIndex() (uint64, error)
```
dhiaayachi, line 212: How would a `GetCommitIndex()` could be implemented in a real store? Would it read the latest stored log and return the commit index associated to it? What if it don't find any because those logs were stored using a store that don't support fast-recovery?

log.go
```diff
@@ -190,3 +190,24 @@ func emitLogStoreMetrics(s LogStore, prefix []string, interval time.Duration, st
...
+	//
+	// The implementation MUST NOT persist this value separately from the log entries.
+	// Instead, it should stage the value to be written atomically with the next
+	// StoreLogs call.
+	//
+	// GetCommitIndex MUST never return a value higher than the last index in the log,
+	// even if a higher value has been staged with this method.
+	//
+	// idx is the new commit index to stage.
+	StageCommitIndex(idx uint64) error
```
dhiaayachi, line 211: Wondering if we can get rid of `StageCommitIndex` in favour of adding a nillable field to the Log struct. 
In all cases if a store don't support fast-recovery it could safely ignore that field and from raft perspective leaving the field to nil signal to the store that we don't want to write the commit index. It also make it clear that a the commit-index is associated with each log.

dhiaayachi (commented):
Thank you @lalalalatt and @banks for working on this!

I added few comments inline, the gist of it is the following:

- I'm not sure about the expected behaviour of `GetCommitIndex` in some cases, like when it read data that don't have commit index.
- I think we can simplify the API by making the commit Index a field in the log struct and removing the staging interface, I don't think that adding that field will break compatibility 
- Left a comment about error handling in `recoverFromCommittedLogs`
- We should add few tests related to error modes and when the persisted log don't have commit index but we activated using fast-recovery.

config.go
```diff
@@ -235,6 +235,16 @@ type Config struct {
...
 
+	// FastRecovery controls if the Raft server should use the fast recovery
+	// mechanism. Fast recovery requires a LogStore implementation that
+	// support commit tracking. When such a store is used and this config
+	// enabled, raft nodes will replay all known-committed logs on disk
+	// before completing `NewRaft` on startup. This is mainly useful where
+	// the application allows relaxed-consistency reads from followers as it
+	// will reduce how far behind the follower's FSM is when it starts. If all reads
+	// are forwarded to the leader then there won't be observable benefit from this feature.
+	FastRecovery bool
```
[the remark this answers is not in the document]
banks: This is great point actually. I'm not sure "Fast" does capture the semantics in any case really: mostly startup will take marginally to a lot longer, but on the plus side the FSM will actually startup in the same state it was before the node restarted which is probably what most users of the library assumed was the case already!

> If my understanding is accurate

Yeah I think you perfectly described the tradeoff. I think `CommittedLogRecovery` is a reasonable name for the thing and we can explain the semantics in the comments.

api.go
banks: 🤔 Great question.

For transient errors it's probably better to hard stop rather than silently weaken the expected guarantees. Someone who has enabled this feature (and provided a supported logstore) should be able to assume that by the time raft has started (without error), their FSM is at least as up-to-date as it was before a restart. Silently falling back seems like it makes it impossible to really trust that guarantee and may mean similar bugs that we are trying to prevent are still possible with no way to mitigate them (even if they are rarer).

In most use cases where each server is using this library as a core of it's functionality (i.e. all our products) and only uses a single raft group initialised during startup, a panic is probably reasonable too because then we crash and the supervisor will restart and if it was a transient error then great and if not it's no different to what will happen later when we are unable to read logs.

That said, in writing this I realised it's probably too strong of a decision to make in a library here: it would be possible for a server process to manage multiple raft instances for example and a fatal error in one of them shouldn't terminate the whole process. So I think I'd vote for making these cases return a hard error from `NewRaft` like we do if we fail to read a log or snapshot during startup rather than panic.

log.go
banks: For BoltDB, I imagine commit index would be a single KV in a separate bucket from logs so it would just read that and return it.

For WAL I anticipated extending the format slightly so that each commit entry in the log stores the most recently staged commit index and then re-populated that into memory when we open the log an scan it like we do with indexes.

If there is no commit index stored, we should just return `0, nil` which is always safe and has the same behavior as current code I think right?

config.go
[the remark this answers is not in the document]
lalalalatt: @schmichael  @banks How about `RecoveryCommitted`?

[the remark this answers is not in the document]
peterxcli: thanks!

api.go
```diff
@@ -697,6 +705,39 @@ func (r *Raft) tryRestoreSingleSnapshot(snapshot *SnapshotMeta) bool {
...
+func (r *Raft) recoverFromCommittedLogs() error {
+	if !r.fastRecovery {
+		return nil
+	}
+
+	// If the store implements CommitTrackingLogStore, we can read the commit index from the store.
+	// This is useful when the store is able to track the commit index and we can avoid replaying logs.
+	store, ok := r.logs.(CommitTrackingLogStore)
+	if !ok {
+		return nil
```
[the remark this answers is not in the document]
lalalalatt: thanks!

banks:
Thanks @schmichael and @dhiaayachi for great and thoughtful reviews.

I've left some inline responses to some thing but wanted to pull out the interface design discussion here as you both brought it up and made two alternative suggestions:

> Would it be safer to just expect log stores to implement a new store log entrypoint since we're having to type assert on every StoreLog call anyway?

Yeah the staging thing ended up kinda janky. I thought it was cleaner at first as it avoided coming up with two alternative ways to actually store the logs... but by the time all the correctness assumptions are documented it certainly seems a bit odd.

Your proposal is probably cleaner and I'm almost sold that it's a better approach, but I do have two thoughts that I'm not quite sure how much weight to give though:

 1. I know there are other abstractions we (might) want to make over log storage - for example my async logs branch where I create a new interface and we switch between either using the `StoreLogs` or `StoreLogsAsync` methods depending on support/mode. We might never land that and could probably make it work even if we do introduce a new `StoreLogsWithCommitIndex` or something now but it certainly doesn't feel like a very nicely composable set of solutions 🤔 . I don't want to end up with each log store having 6+ different variants of `Store` to cover each possible combination of optional features! Although it's weird, one benefit of the current design is that it composes with any other changes to store logs.
 2. I agree the interface is more straightforward if we passed the commit index in along with the logs, but we would still need to document and require that implementations handle them in a specific way (i.e. atomically with respect to both crash safety and read calls). In practice that's not any easier to implement

> Wondering if we can get rid of `StageCommitIndex` in favour of adding a nillable field to the Log struct.

I tried to avoid adding it to the LogStruct for a couple of reasons:
 1. Log structs escape the boundaries of the API and I don't love the fact that applications would get direct access to a field that seems useful called `CommitIndex` but is extremely unintuitive what it means: it's not directly related to the actual index of the log it's in nor even related to whether or not that log is committed. Worst of all, _it will be different on different nodes_. Because replication batches might be different at different followers. It would be pretty surprising to expose log structs for the same exact term and index but have different nodes have different metadata for that lock.
 2. Typically we store batches of logs so either we'd pay the (small) cost of encoding and storing that extra uint64 for every log when we only need it once per batch at most. If we set every log in the batch to have the same commit index then compressible encodings can't omit it, but if we don't then it's even more surprising to the user that the logs come out with "random" ones having it set but not others (again different on different nodes).

Now the first could be mitigated by always stripping or nulling that field back out in Raft before we return it to the application, but in general it feels like the metadata is attached in the wrong place. Both of our LogStore implementation have efficient ways to store one extra int of metadata alongside a batch of logs without resorting to encoding it in the log itself so it seems like that is a better interface to target to me?

I don't think it's a terrible idea, it just doesn't seem as clean - more like (ab)using the existing type we have to pass extra data around even when it doesn't really make sense there instead of having a new interface with the right semantics?

Curious about the thoughts here though?

log.go
dhiaayachi: > If there is no commit index stored, we should just return 0, nil which is always safe and has the same behavior as current code I think right?

I agree! I think that should be documented though. Because the API allow erroring on `GetCommitIndex()` it could easily be mistaken as a possible error case.

api.go
dhiaayachi: I guess returning an error from `NewRaft` would allow the library user to decide if they would like to do a fallback or fail starting the application, that is reasonable I think.

I think we should implement this as a specific error type that we publish on the API to allow calling `errors.Is` on it without relying on string compare.

peterxcli:
> Would it be safer to just expect log stores to implement a new store log entrypoint since we're having to type assert on every StoreLog call anyway?

Is it reasonable to add a new field in Raft struct and use it to cache the pointer to CommitTrackingLogStore at the start of Raft process? Then we could use the store directly and no type assertion anymore.
But as the store variant become more and more, it would be a mass😅...

schmichael:
> the store provided value SHOULD NOT exceed the last index in the persisted logs in the store

You made me realize something: I think `SHOULD` is key there. It's *not* a `MUST`. If the committed index persisted exceeds the highest index persisted log then the remaining logs must be streamed from a peer. AFAICT the persisted commit index is only needed to ensure *uncommitted* logs aren't replayed.

This implies the existing API is fine and not really brittle. A misbehaving implementation that writes the commit index on every call distinctly from StoreLogs is a performance issue (since its doing extra IOPs and those IOPs might be "wasted" if they're for a commit index ahead of what StoreLogs has persisted), not a correctness issue (since on recovery the delta from last-persisted-log to persisted-commit-index will be streamed).

If my understanding is correct then we need to relax the `MUST`s in `StageCommitIndex` to `SHOULD`s or similar.

That being said I'm not sure if relaxing the atomicity complicates the implementation of Raft or LogStores! If requiring atomicity simplifies the implementation: we're probably best off doing that since that's what we already have in front of us. :)

If it does not complicate implementations, I vote we change `StageCommitIndex -> SetCommitIndex` or similar and ship this as-is.

My idea to avoid runtime type assertions by turning minimal-logstore-implementations into maximal-logstore-implementations via noop-adapters could even be done as a followup PR with very little-to-no debt incurred by first accepting this approach.

---

Regardless I am quite pleased with the implementation and options here. There's not an option mentioned I would oppose, so as soon as CI is green I'm happy to :+1: and have 1 less cook in the kitchen.

config.go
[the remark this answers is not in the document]
schmichael: Naming is the worst. :sweat_smile: Looking around a bit we're not totally consistent, but I think generally:
1. `Recovery` refers to remediating a failure of some kind ([`RecoverCluster`](https://pkg.go.dev/github.com/hashicorp/raft#RecoverCluster) and [many comments](https://github.com/search?q=repo%3Ahashicorp%2Fraft%20recover&type=code))
2. [`Restore`](https://pkg.go.dev/github.com/hashicorp/raft#Raft.Restore) is the term used for this process: replaying logs on startup.

So I think switching `Recovery -> Restore` is appropriate.

After that I think I prefer `RestoreCommittedLogs` a tiny bit over `RestoreCommitted` since `Committed` alone is a tiny bit ambiguous. I would accept either though!

dhiaayachi:
I agree, if the commit index is higher we can always fallback to the current implementation but in this case we can apply the same logic to any error mode specific to this feature and chose to return an error from NewRaft and let the library user decide what to do with it.

Also it will be hard to determine that error without processing most of the logs, as we read them in sequence from the oldest to the newest.

dhiaayachi:
@banks @lalalalatt Same here, I don't have a strong opinion about the API and I'm ok to ✅  this as is. The only blocking items for me are:

- relaxing the panic when restoring to return an error instead
- document the expected behaviour when GetCommitIndex is called and no commit index is found
- add some error mode tests

api.go
```diff
@@ -697,6 +705,39 @@ func (r *Raft) tryRestoreSingleSnapshot(snapshot *SnapshotMeta) bool {
...
+		return
+	}
+
+	// If the store implements CommitTrackingLogStore, we can read the commit index from the store.
+	// This is useful when the store is able to track the commit index and we can avoid replaying logs.
+	store, ok := r.logs.(CommitTrackingLogStore)
+	if !ok {
+		r.logger.Warn("fast recovery enabled but log store does not support it", "log_store", fmt.Sprintf("%T", r.logs))
+		return
+	}
```
tgross, line 720: As long as we're considering returning an error below instead of panicking, I think we should consider doing so here as well. This is going to be a "programmer error" rather than a runtime error -- the consumer of the library should be ensuring they're passing a compatible combination of log store and `FastRecovery` configuration.

raft.go
```diff
@@ -1262,6 +1262,9 @@ func (r *Raft) dispatchLogs(applyLogs []*logFuture) {
 		r.leaderState.inflight.PushBack(applyLog)
 	}
 
+	commitIndex := r.getCommitIndex()
```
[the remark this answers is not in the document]
lalalalatt: @schmichael 
See the https://github.com/hashicorp/raft/blob/bdac45bbccfdfa13a6e90ad660473a2f16ea13c2/raft.go#L1555-L1558

lalalalatt:
To support error assertion for `NewRaft` error when making cluster, I add a error return field in `makeCluster`, but it requires a bunch of change for existing test to handle the returned error.
I think it is necessary cause I guess there would be more and more log store variant appearing, and their error mode test are needed.

- https://github.com/hashicorp/raft/pull/613/commits/cfffcb5b7e506de502c3c52e9c88ea87958e2737

peterxcli:
To support error assertion for `NewRaft` error when making cluster, I add a error return field in `makeCluster`, but it requires a bunch of change for existing test to handle the returned error.
I think it is necessary cause I guess there would be more and more log store variant appearing, and their error mode test are needed.

- https://github.com/hashicorp/raft/pull/613/commits/cfffcb5b7e506de502c3c52e9c88ea87958e2737

testing.go
```diff
@@ -717,20 +717,21 @@ WAIT:
...
+	MonotonicLogs      bool
+	CommitTrackingLogs bool
 }
 
 // makeCluster will return a cluster with the given config and number of peers.
 // If bootstrap is true, the servers will know about each other before starting,
 // otherwise their transports will be wired up but they won't yet have configured
 // each other.
-func makeCluster(t *testing.T, opts *MakeClusterOpts) *cluster {
+func makeCluster(t *testing.T, opts *MakeClusterOpts) (*cluster, error) {
```
schmichael, line 734: AFAICT this was added to support 1 test that asserts a specific error is returned. Since this causes tons of other test churn I have a slight preference for reverting the error return here and special casing the 1 test that needs to check the error return of NewRaft. Not a blocker though.

schmichael (dismissed):
:tada:

log.go
```diff
@@ -190,3 +190,28 @@ func emitLogStoreMetrics(s LogStore, prefix []string, interval time.Duration, st
...
+
+	// GetCommitIndex returns the latest persisted commit index from the latest log entry
+	// in the store at startup.
+	// It is ok to return a value higher than the last index in the log (But it should never happen).
```
dhiaayachi, line 215:
```suggestion
	// GetCommitIndex should not return a value higher than the last index in the log. If that happens, the last index in the log will be used.
```

dhiaayachi: I think we should also document here that GetCommitIndex need to return 0,nil when no commit index is found in the log.

dhiaayachi (commented):
Thank you again for the prompt update @lalalalatt!
I left a minor suggestion for a comment, other then that I think from testing perspective we can add more tests that cover the following cases:

- the logstore return 0, nil all the time (no commit index is in the store). 
- GetCommitIndex return a value that is bigger then the last index

lalalalatt:
Hi, @banks, any new progress on this?

banks:
Thanks @lalalalatt and sorry it's been a while on this one, time passes! I think @dhiaayachi and @schmichael and @tgross did a great job following up after my initial review so I've dismissed my concerns which seems to be outdated now.

I've not had a chance to do another thorough review yet so I'll not approve right now but I've unblocked the merge if one of the others was already at a point ready to approve.

banks:
@peterxcli thanks for the links.

I think their approach is essentially the same as this one - note that this is why I pushed this design towards `CommitTrackingLogStore` as an extension of log store rather than just storing it in the existing `StableStore`. Either way it lets you persist the extra data in the logs without additional fsyncs on every `AppendEntries`.

Their specific approach of have an actual log record type for it instead of having `LogStore` guarantee atomic commitment of that metadata is interesting and is an alternative we could consider - it would require no changes from logstores for example. I think it would need slightly more change to raft to introduce a new log type, and there would be compatibility issues to consider where new leader can't replicate those messages to old followers because they would panic... That was one of the reasons I proposed an out-of-band metadata solution rather than a new log type I think because it avoid the need for a complex compatibility dance that could break upgrades.

tgross (dismissed):
I've had a chance to re-review and I'm :+1: on merging, once the merge conflicts have been resolved.

lalalalatt:
@tgross Resolved, PTAL. Thanks!

tgross (approved):
LGTM

schmichael (approved):
Thanks for sticking with this @lalalalatt and everyone else. I'm +1 on merging.

A note to others: there's currently no https://github.com/hashicorp/raft-boltdb (AFAICT), so it's not possible to functionally test this in products like HashiCorp Nomad.

That being said, I'm fine with merging first so we can treat the API contract as finalized.

schmichael:
No clue why I didn't merge this before but better late than never.

PR #560, opened 2023-06-27
add CommitIndex API

niebayes:
This API is useful for implementing the read index optimization as described in the Raft paper.

PR #625, opened 2024-10-16
Support checking assertion of Leadership

otoolep:
Potential fix for https://github.com/hashicorp/raft/issues/621

This allows clients of this library to check if the current node, assuming it is the Leader, has explicitly asserted leadership in the current term. "Asserting" is defined has having committed at least one log entry in the current term, as per the Raft paper.

This change adds a new channel, which is only serviced when in Leadership mode. This channel will respond with the _asserted_ status via a _Future_. Clients are recommended to call this function only until they have confirmed that the leadership has been asserted. Once asserted for a given term the state can be cached until the end of the term.

This code is necessary to properly implement the Read Optimization outlined in the Raft paper and dissertation. The dissertation explains that it is not sufficient to simply check the Leader's commit index. Instead a client must first confirm that the Leader has explicitly asserted leadership by committing a log entry in the current term. Only then is commit index valid.


otoolep:
This is a follow up to https://github.com/hashicorp/raft/pull/560

otoolep:
Would you appreciate any help on the test failure, don't immediately see how it could be related to this change.

otoolep:
Looks like a racy test, passes now. 

Any input on this PR?

otoolep:
Gentle ping on this PR.

PR #630, opened 2024-12-09
chore: Add flag to skip legacy duplicate telemetry

kavirajk:
Currently we have some legacy metrics with `peer_id` in the metrics suffix (in addition to same metrics with `peer_id`d as label)
1. `raft_replication_appendEntries_rpc_peer0`
2. `raft_replication_appendEntries_logs_peer0`
3. `raft_replication_heartbeat_peer0`
4. `raft_replication_installSnapshot_peer0`

These metrics may have additional `_count` or `_sum` metrics. And each metrics are multiplicative. Meaning if I have 10 peers, these metrics will be 10x.

This PR adds a flag `noLegacyTelemetry` (default: false) which by setting to `true` you can skip those duplicate metrics.

### Before (`weaviate_internal` is some custom prefix)
Screenshot 2024-12-09 at 23 37 21

### After
Screenshot 2024-12-09 at 23 28 28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # UNRELEASED
 
+IMPROVEMENETS
+
+* chore: Add a flag to skip legacy duplicate telemetry. [GH-630](https://github.com/hashicorp/raft/pull/630)
+
 # 1.7.0 (June 5th, 2024)
 
 CHANGES
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 IMPROVEMENETS
 
-* chore: Add a flag to skip legacy duplicate telemetry. [GH-630](https://github.com/hashicorp/raft/pull/630)
+* Added a flag to skip legacy duplicate telemetry. [GH-630](https://github.com/hashicorp/raft/pull/630)
 
 # 1.7.0 (June 5th, 2024)
 
--- a/api.go
+++ b/api.go
@@ -218,8 +218,8 @@ type Raft struct {
 	// prevote feature is disabled if set to true.
 	preVoteDisabled bool
 
-	// noLegacyTelemetry allow to skip the legacy metrics to avoid duplicates.
-	// legacy metrics are which has `_peer_name` as metric suffix instead as labels.
+	// noLegacyTelemetry allows to skip the legacy metrics to avoid duplicates.
+	// legacy metrics are those that have `_peer_name` as metric suffix instead as labels.
 	// e.g: raft_replication_heartbeat_peer0
 	noLegacyTelemetry bool
 }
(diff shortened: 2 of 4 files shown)


api.go
```diff
@@ -217,6 +217,11 @@ type Raft struct {
 	// preVoteDisabled control if the pre-vote feature is activated,
 	// prevote feature is disabled if set to true.
 	preVoteDisabled bool
+
+	// noLegacyTelemetry allow to skip the legacy metrics to avoid duplicates.
```
pkazmierczak, line 221:
```suggestion
	// noLegacyTelemetry allows to skip the legacy metrics to avoid duplicates.
```

config.go
```diff
@@ -237,6 +237,11 @@ type Config struct {
 
 	// skipStartup allows NewRaft() to bypass all background work goroutines
 	skipStartup bool
+
+	// NoLegacyTelemetry allow to skip the legacy metrics to avoid duplicates.
+	// legacy metrics are which has `_peer_name` as metric suffix instead as labels.
+	// e.g: raft_replication_heartbeat_peer0
```
pkazmierczak, line 243:
```suggestion
	// NoLegacyTelemetry allows to skip the legacy metrics to avoid duplicates.
	// legacy metrics are those that have `_peer_name` as metric suffix instead 
	// of labels.
	// e.g: raft_replication_heartbeat_peer0
```

pkazmierczak (dismissed):
LGTM, thanks @kavirajk. Mind adding a changelog entry?

config.go
```diff
@@ -237,6 +237,11 @@ type Config struct {
...
+	NoLegacyTelemetry bool
```
banks, line 244: Minor thing, but we'd prefer to move this above the private internal-only `skipStartup` option. We typically keep these internal-only details grouped at the end of the struct just to make it clearer at a glance which configs are public.

banks (commented):
Thanks @kavirajk this is a useful addition and something we have not gotten around to doing for way to long 😅 .

I made one super minor request that doesn't impact behavior. I think @pkazmierczak also requested a changelog update which is hopefully simple.

If you're able to do those I think we'd be good to merge this. If not we can probably take care of that in a spare minute, but I'm a fan of commits/attribution staying with the person who did the hard work if possible!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # UNRELEASED
 
+IMPROVEMENETS
+
+* chore: Add a flag to skip legacy duplicate telemetry. [GH-630](https://github.com/hashicorp/raft/pull/630)
```
pkazmierczak, line 5:
```suggestion
* Added a flag to skip legacy duplicate telemetry. [GH-630](https://github.com/hashicorp/raft/pull/630)
```

pkazmierczak (commented):
Minor suggestion about the phrasing of the changelog entry, but otherwise lgtm!

pkazmierczak (approved):
LGTM, thanks for the contribution @kavirajk!

PR #633, opened 2025-01-09
Add linter rules to ensure we don't accidentally emit metrics directly instead of through the compat package

mkeeler:
My previous PR switched the repo over to emitting metrics via github.com/hashicorp/go-metrics/compat. In discussions with other engineers it came up that if new metrics are added it would be good to ensure that they use the compat package as well. To that end, this PR introduces linter rules to ensure only the compat package is used.


schmichael (approved):
:heart: :heart: :heart: linting guardrails

PR #636, opened 2025-03-11
Add a log line on active node when sending snapshot to a follower

kubawi:
This change is motivated by a request from Vault's support to make it easier to understand the progress of replication. 

Currently, we do some logging on the standby nodes, but our support would like to be able to see what is happening in the logs earlier, especially in Vault deployments when the active node is slow/overwhelmed or when it has a big database.

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -337,12 +337,12 @@ func (r *Raft) sendLatestSnapshot(s *followerReplication) (bool, error) {
 	peer := s.peer
 	s.peerLock.RUnlock()
 
-	r.logger.Info("installing snapshot on", "peer", peer, "id", snapID, "size", req.Size)
+	r.logger.Info("installing snapshot on", "peer", peer.ID, "id", snapID, "size", req.Size)
 	// Make the call
 	start := time.Now()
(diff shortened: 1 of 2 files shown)


replication.go
```diff
@@ -337,6 +337,7 @@ func (r *Raft) sendLatestSnapshot(s *followerReplication) (bool, error) {
 	peer := s.peer
 	s.peerLock.RUnlock()
 
+	r.logger.Info("installing snapshot on", "peer", peer, "id", snapID, "size", req.Size)
```
schmichael, line 340:
```suggestion
	r.logger.Info("installing snapshot on", "peer", peer.ID, "id", snapID, "size", req.Size)
```

Also do you mind adding `peer.ID` and `snapID` to the other two log lines in this func? It's nice when related log lines have the appropriate context to tie them all together.

kubawi: Good shout, Michael, thanks! Fixed in 9c2797b04868eff65dbfb18fae761aedd75c4c06

schmichael (approved):

raskchanky (approved):

PR #640, opened 2025-04-04
Add WaitCommitted

kolesnikovae:
Hello team,
 
Just wanted to say thanks for the awesome work on this project – it's been a pleasure to work with. We're using raft in a new experimental component we're building over at [Grafana Pyroscope](https://github.com/grafana/pyroscope), and it's been super helpful.

As part of our integration, we encountered a use case where quorum replication is sufficient for correctness, and blocking until FSM application introduces unnecessary latency. This PR proposes a small addition that enables that workflow in a clean and backward-compatible way: the new `WaitCommitted()` method in the `ApplyFuture` interface. Unlike `ApplyFuture.Error()`, which blocks until the command has been committed and applied to the local FSM, `WaitCommitted()` allows users to wait only until the log entry has been replicated to a quorum and marked committed – without blocking on FSM application.

I'm happy to implement the feature in a different way if needed – the code in the PR is just one option I considered. Testing for the new API is also not particularly extensive at the moment, but I'd be more than willing to fill in the gaps if the general idea is accepted.

### DIFF
--- a/future.go
+++ b/future.go
@@ -36,6 +36,12 @@ type IndexFuture interface {
 type ApplyFuture interface {
 	IndexFuture
 
+	// WaitCommitted blocks until the log entry has been committed to quorum.
+	// It does not wait for FSM application.
+	// The error returned follows the same semantics as the Error method,
+	// except for errors that occur after the log entry has been committed.
+	WaitCommitted() error
+
 	// Response returns the FSM response as returned by the FSM.Apply method. This
 	// must not be called until after the Error method has returned.
 	// Note that if FSM.Apply returns an error, it will be returned by Response,
@@ -85,6 +91,10 @@ func (e errorFuture) Response() interface{} {
 
 func (e errorFuture) Index() uint64 {
 	return 0
+}
+
+func (e errorFuture) WaitCommitted() error {
+	return e.err
 }
 
 // deferError can be embedded to allow a future
@@ -151,9 +161,15 @@ type bootstrapFuture struct {
 // the log is considered committed.
 type logFuture struct {
 	deferError
-	log      Log
-	response interface{}
-	dispatch time.Time
+	log       Log
+	response  interface{}
+	dispatch  time.Time
+	committed chan struct{}
+}
+
+func (l *logFuture) init() {
+	l.committed = make(chan struct{})
+	l.deferError.init()
 }
 
 func (l *logFuture) Response() interface{} {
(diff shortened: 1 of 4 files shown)


raft.go
```diff
@@ -569,6 +569,7 @@ func (r *Raft) runLeader() {
 	// maintain that there exists at most one uncommitted configuration entry in
 	// any log, so we have to do proper no-ops here.
 	noop := &logFuture{log: Log{Type: LogNoop}}
+	noop.init()
```
kolesnikovae, line 572: It seems that the no-op command issued by the leader at the beginning of a term is not properly initialized, which means that some errors may be silently ignored:

```go
func (d *deferError) respond(err error) {
	if d.errCh == nil {
		return
	}
	if d.responded {
		return
	}
	d.errCh <- err
	close(d.errCh)
	d.responded = true
}
```

raft.go
```diff
@@ -818,6 +819,7 @@ func (r *Raft) leaderLoop() {
 				groupReady = append(groupReady, e)
 				groupFutures[idx] = commitLog
 				lastIdxInGroup = idx
+				close(commitLog.committed)
```
kolesnikovae, line 822: I went back and forth between a "safer" option with a nil check (similar to the future error channel) and a more direct approach with an unconditional `close`. I opted for the latter because I couldn't identify a case where the same `logFuture` would be committed twice, which, in that case, would likely lead to a bigger issue.

future.go
```diff
@@ -36,6 +36,12 @@ type IndexFuture interface {
 type ApplyFuture interface {
 	IndexFuture
 
+	// WaitCommitted blocks until the log entry has been committed to quorum.
+	// It does not wait for FSM application.
+	// The error returned follows the same semantics as the Error method,
+	// except for errors that occur after the log entry has been committed.
+	WaitCommitted() error
```
kolesnikovae, line 43: However, there's an implicit dependency on the `applyCh` buffer size, which is controlled via the and `BatchApplyCh` and `MaxAppendEntries` configuration parameters (defaults to 64, max 1024, if `BatchApplyCh` is set to true). Once the buffer is full, `Apply` blocks.

```
	// MaxAppendEntries controls the maximum number of append entries
	// to send at once. We want to strike a balance between efficiency
	// and avoiding waste if the follower is going to reject because of
	// an inconsistent log.
```

I'd like to control the `applyCh` buffer size explicitly (as 1024 is fairly conservative) via a separate option, e.g., `AppendBufferSize` which would default to `MaxAppendEntries` and could not be less than that. It could replace the `BatchApplyCh` option.

PR #641, opened 2025-04-08
IND-2885 Linting

KaushikiAnand:
- I have added linting to go-tests.yml.
- The linting was done to ensure that the codebase is consistent and maintainable and helps in improving the code quality and reducing errors.

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"net"
 	"os"
 	"sync"
@@ -314,9 +313,7 @@ func (n *NetworkTransport) CloseStreams() {
 	// entry.
 	for k, e := range n.connPool {
 		for _, conn := range e {
-			if err := conn.Release(); err != nil {
-				log.Print(err)
-			}
+			conn.Release()
 		}
 
 		delete(n.connPool, k)
@@ -442,9 +439,7 @@ func (n *NetworkTransport) returnConn(conn *netConn) {
 	if !n.IsShutdown() && len(conns) < n.maxPool {
 		n.connPool[key] = append(conns, conn)
 	} else {
-		if err := conn.Release(); err != nil {
-			log.Print(err)
-		}
+		conn.Release()
 	}
 }
 
@@ -517,13 +512,7 @@ func (n *NetworkTransport) InstallSnapshot(id ServerID, target ServerAddress, ar
 	if err != nil {
 		return err
 	}
-	var errors error
-	defer func() {
-		if err := conn.Release(); err != nil && errors == nil {
-			errors = err
-			return
-		}
-	}()
+	defer conn.Release()
 
 	// Set a deadline, scaled by request size
 	if n.timeout > 0 {
@@ -783,17 +772,13 @@ func decodeResponse(conn *netConn, resp interface{}) (bool, error) {
 	// Decode the error if any
 	var rpcError string
 	if err := conn.dec.Decode(&rpcError); err != nil {
-		if err := conn.Release(); err != nil {
-			return false, err
-		}
+		conn.Release()
 		return false, err
 	}
 
 	// Decode the response
 	if err := conn.dec.Decode(resp); err != nil {
-		if err := conn.Release(); err != nil {
-			return false, err
-		}
+		conn.Release()
 		return false, err
 	}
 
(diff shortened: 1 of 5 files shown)


net_transport.go
```diff
@@ -441,7 +442,9 @@ func (n *NetworkTransport) returnConn(conn *netConn) {
 	if !n.IsShutdown() && len(conns) < n.maxPool {
 		n.connPool[key] = append(conns, conn)
 	} else {
-		conn.Release()
+		if err := conn.Release(); err != nil {
+			log.Print(err)
```
abhijeetviswa, line 446: Same as above.

net_transport_test.go
```diff
@@ -635,7 +635,9 @@ func TestNetworkTransport_InstallSnapshot(t *testing.T) {
 
 				// Try to read the bytes
 				buf := make([]byte, 10)
-				rpc.Reader.Read(buf)
+				if _, err := rpc.Reader.Read(buf); err != nil {
+					t.Errorf("unable to read bytes: %v", err)
```
abhijeetviswa, line 639: Suggestion: Can we wrap the error instead of formatting it into the string i.e `%v` -> `%w`?

raft_test.go
```diff
@@ -1859,7 +1859,7 @@ func TestRaft_Barrier(t *testing.T) {
 	// Ensure all the logs are the same
 	c.EnsureSame(t)
 	if len(getMockFSM(c.fsms[0]).logs) != 100 {
-		t.Fatalf(fmt.Sprintf("Bad log length: %d", len(getMockFSM(c.fsms[0]).logs)))
+		t.Fatalf("%s", fmt.Sprintf("Bad log length: %d", len(getMockFSM(c.fsms[0]).logs)))
```
abhijeetviswa, line 1862: Question: Why not use `t.Fatal(...)` instead?

net_transport.go
```diff
@@ -793,19 +808,25 @@ func decodeResponse(conn *netConn, resp interface{}) (bool, error) {
 func sendRPC(conn *netConn, rpcType uint8, args interface{}) error {
 	// Write the request type
 	if err := conn.w.WriteByte(rpcType); err != nil {
-		conn.Release()
+		if err := conn.Release(); err != nil {
+			return err
+		}
```
tgross, line 813: What's a circumstance where we're going to get an error here and want that to be the return value instead of the original error? (Similar for all these changes.)

net_transport.go
```diff
@@ -872,7 +896,9 @@ func (n *netPipeline) AppendEntries(args *AppendEntriesRequest, resp *AppendEntr
 
 	// Add a send timeout
 	if timeout := n.trans.timeout; timeout > 0 {
-		n.conn.conn.SetWriteDeadline(time.Now().Add(timeout))
+		if err := n.conn.conn.SetWriteDeadline(time.Now().Add(timeout)); err != nil {
+			return nil, err
+		}
```
tgross, line 901: How does this change behavior in the caller? We're changing the error handling in _many_ places in this PR.  We're doing this for the sake of linting without much in the way of consideration for how it changes the expected behavior. "It silences linters" isn't a good enough reason to make this change without that consideration.

(This comment applies to a lot of these.)

tgross (changes requested):
We need to be _very_ careful about introducing behavior changes and swallowing important errors in production code for the sake of silencing lints. If we can't justify each change aside from lints I'm a -1 on this PR.

mukeshjc:
@tgross 
Thank you for the review. We have discussed this internally and plan to take below approach for moving forward:

- For all instances of teardown functions _(eg: conn.Resolve(), sink.Cancel() etc.)_  that are flagged by the linter, we will [exclude them to suppress the findings](https://golangci.github.io/legacy-v1-doc/usage/linters/#errcheck) since they aren't needed for teardown type of functions.
- Many `errcheck` specific lint failures were for unit test files, so we will [add config to ignore test files from that linter](https://golangci-lint.run/usage/configuration/#linters-configuration:~:text=rules%3A,%2D%20errcheck). 

Thoughts?

tgross:
> * For all instances of teardown functions (eg: conn.Resolve(), sink.Cancel() etc.) that are flagged by the linter, we will [exclude them to suppress the findings](https://golangci.github.io/legacy-v1-doc/usage/linters/#errcheck) since they aren't needed for teardown type of functions.

That's a reasonable approach. Another option would be to figure out which teardown functions we _do_ need to care about and get the added error context wrapped in the original error we return. But that's certainly more work and needs more attention.

> * Many errcheck specific lint failures were for unit test files, so we will [add config to ignore test files from that linter](https://golangci-lint.run/usage/configuration/#linters-configuration:~:text=rules%3A,%2D%20errcheck).

Why? It seems totally safe to add the extra checks for unit tests, right?

mukeshjc:
> Why? It seems totally safe to add the extra checks for unit tests, right?

The idea is fixing these linting checks in unit test files doesn't improve the actual library. I'm happy to include them too, no strong opinions here. We will keep them.

tgross:
> The idea is fixing these linting checks in unit test files doesn't improve the actual library. I'm happy to include them too, no strong opinions here. We will keep them.

This is a little all-or-nothing. Either the lint is useful, in which case you should apply it _where it's safe to apply_ or it's not useful at all. I think it _is_ useful, so applying it to test code in particular could be handy.

Your change in e1a5e6a70bdf9503ddb2e97c1fc78a00f2437546 leaves a number of cases in the production code, however.

mukeshjc:
> Your change in [e1a5e6a](https://github.com/hashicorp/raft/commit/e1a5e6a70bdf9503ddb2e97c1fc78a00f2437546) leaves a number of cases in the production code, however.

We plan to re-visit the need for error wrapping of teardown functions after syncing with the group in #tech-raft slack channel. Based on the feedback we will keep/remove the suppressed findings.

PR #644, opened 2025-05-14
Enhance network transport debugging

VictorLowther:
Adds a couple of things for debuggability and stability:

1. When a follower fails to process an incoming RPC request for any reason, print up to 100 bytes of the remaining buffered bytes in the incoming connection.  This gives more detail for being able to resolve weird network issues where the client has msgpack decoding issues or is getting outright gibberish in the RPC request without needing client at remote sites to have to resort to packet captures or hand-decoding TLS encrypted streams.
2. Have the leader reset buffers and codecs for pooled network connections whenever they are pulled from the pool, and arrange for things to panic if anything tries to use a pooled conn when it has not been checked out.

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -54,6 +54,16 @@ const (
 	minInFlightForPipelining = 2
 )
 
+type zs struct{}
+
+func (zs) Read(buf []byte) (int, error) {
+	panic("Read should never be called")
+}
+
+func (zs) Write(buf []byte) (int, error) {
+	panic("Write should never be called")
+}
+
 var (
 	// ErrTransportShutdown is returned when operations on a transport are
 	// invoked after it's been terminated.
@@ -61,6 +71,7 @@ var (
 
 	// ErrPipelineShutdown is returned when the pipeline is closed.
 	ErrPipelineShutdown = errors.New("append pipeline closed")
+	xplod               = zs{}
 )
 
 // NetworkTransport provides a network based transport that can be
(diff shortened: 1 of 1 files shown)


net_transport.go
```diff
@@ -626,6 +635,11 @@ func (n *NetworkTransport) handleConn(connCtx context.Context, conn net.Conn) {
 			if err != io.EOF {
 				n.logger.Error("failed to decode incoming command", "error", err)
 			}
+			if r.Buffered() > 0 {
+				snippet := [100]byte{}
+				cnt, _ := r.Read(snippet[:])
+				n.logger.Error("remaining read buffer", "sz", cnt, "data", string(snippet[:cnt]))
```
schmichael, line 641: I don't think it's safe to log unknown payloads as they may contain secrets (although it seems _exceeding_ unlikely given the first 100 bytes are likely command headers).

I think if you want more context for debugging we should add it to the errors returned by handleCommand so context can be added on a case-by-case basis.

VictorLowther: I mean fair, but I have no idea what the case by case basis would look like -- my primary concern is tracking down exactly where gibberish is being injected.  I suspect it is the inappropriate buffer use addressed elsewhere in this PR, but have not been able to pin it down.

schmichael: Ah sorry I see the complication: other than the initial `ReadByte` (which could use some extra context but is unlikely to be the problem), all other errors would come directly from `codec.Decoder` and largely outside of our control to annotate.

The one exception there would be wrapping Decode error return values with the `rpcType` which might be a helpful bit of context.

To add more debugging beyond that I think we need to make it configurable (and `off` by default) so that production environments can avoid leaking secrets to logs.

schmichael: (There's also some opportunity to annotate the response encoding errors with more context, but I suspect that's unrelated to the immediate problem you're trying to debug: https://github.com/hashicorp/raft/blob/f207c248e7ab7b8f6c3d7d648f11737b53e22352/net_transport.go#L755-L761)

PR #669, opened 2026-03-19
fix flake in test for restarted followers not causing an election

tgross:
While working on #666, I encountered a test flake caused by a race condition in `TestRaft_FollowerRemovalNoElection`, which is also obscuring the test intent. After we restart the follower, typically the `c.ConnectFully` call will complete quickly enough that the follower never enters a candidate state, so we're not exercising the intended behavior. But if the `c.ConnectFully` call takes too long, the follower can miss heartbeats and start an election. This is expected behavior and its request for an election will be rejected. But in that case, the follower will not have settled on the leader by the time we check it and the test will fail when it shouldn't.

Add an extra sleep before we reconnect the cluster to force the restarted follower to become a candidate so that we're exercising the intent of the test, but then wait until the follower shows an event that it's got a leader before continuing to the remaining assertions.

Note that to hit this flake you need to run this test a _lot_. Ex. with `go test -v -failfast -count=30 . -run TestRaft_FollowerRemovalNoElection`. The work I'm doing in #666 changes timing slightly, and this caused it to happen maybe 10% of the time instead of 5%.

### DIFF
--- a/raft_test.go
+++ b/raft_test.go
@@ -3106,14 +3106,37 @@ func TestRaft_FollowerRemovalNoElection(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error restarting follower: %v", err)
 	}
+	n.RegisterObserver(NewObserver(c.observationCh, false, nil))
+
 	c.rafts[i] = n
 	c.trans[i] = n.trans.(*InmemTransport)
 	c.fsms[i] = n.fsm.(*MockFSM)
+	time.Sleep(500 * time.Millisecond)
 	c.FullyConnect()
-	// There should be no re-election during this sleep
-	time.Sleep(250 * time.Millisecond)
 
-	// Let things settle and make sure we recovered.
+	// Wait for the restarted follower to settle.
+	ch := c.WaitEventChan(t.Context(), func(o *Observation) bool {
+		if o == nil {
+			return false
+		}
+		if o.Raft.localID == follower.localID {
+			if obs, ok := o.Data.(LeaderObservation); ok {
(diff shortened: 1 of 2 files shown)


raft_test.go
```diff
@@ -3106,14 +3106,37 @@ func TestRaft_FollowerRemovalNoElection(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error restarting follower: %v", err)
 	}
+	n.RegisterObserver(NewObserver(c.observationCh, false, nil))
+
 	c.rafts[i] = n
 	c.trans[i] = n.trans.(*InmemTransport)
 	c.fsms[i] = n.fsm.(*MockFSM)
+	time.Sleep(500 * time.Millisecond)
```
schmichael, line 3116: Hm, not sure my comment is really helpful, but I do think it's a good habit to explain why sleeps are necessary.

```suggestion

	// Give follower time to become a candidate before connecting
	time.Sleep(500 * time.Millisecond)
```

raft_test.go
```diff
@@ -3106,14 +3106,37 @@ func TestRaft_FollowerRemovalNoElection(t *testing.T) {
...
 	c.FullyConnect()
-	// There should be no re-election during this sleep
-	time.Sleep(250 * time.Millisecond)
 
-	// Let things settle and make sure we recovered.
+	// Wait for the restarted follower to settle.
```
schmichael:
```suggestion
	// Wait for the restarted follower to recognize leader.
```

schmichael (dismissed):
Fine as-is

PR #683, opened 2026-05-19
fix liveness issue when leader's disk stalls

### LINKED ISSUE #503: When the leader’s raft.LogStore hangs indefinitely, the network will hang indefinitely instead of re-electing
I recently encountered an issue in my production deployment of https://robustirc.net/, where the network was not making any Raft progress anymore.

It turns out that one of my servers has an issue with its (local NVMe) storage, which manifests itself in hanging indefinitely. There are no read or write errors, any disk access just hangs.

When that server happens to currently be the Raft leader when the issue occurs, **the entire Raft network will just hang indefinitely**. By this, I mean the leader will still participate in the Raft protocol (last contact times do update on the Raft followers), but applying new messages to Raft will not work (will timeout) and, crucially, the Raft network never even starts electing a new leader.

This was a surprising failure mode to me, and I wonder if that’s intentional (out of scope for Raft) or an issue with the current implementation?

To reproduce, I cloned the Raft example https://github.com/yongman/leto and modified it like so:

```diff
diff --git i/store/store.go w/store/store.go
index 6997242..2d63d77 100644
--- i/store/store.go
+++ w/store/store.go
@@ -9,10 +9,14 @@ package store
 import (
 	"encoding/json"
 	"errors"
+	fmt "fmt"
 	"log"
 	"net"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"sync"
+	"syscall"
 	"time"
 
 	"github.com/hashicorp/raft"
@@ -46,6 +50,51 @@ func NewStore(raftdir, raftbind string) (*Store, error) {
 	}, nil
 }
 
+type hangingDB struct {
+	raft.LogStore
+
+	hangingMu sync.Mutex
+	hanging   bool
+}
+
+func (h *hangingDB) currentlyHanging() bool {
+	h.hangingMu.Lock()
+	defer h.hangingMu.Unlock()
+	return h.hanging
+}
+
+func (h *hangingDB) hang() {
+	h.hangingMu.Lock()
+	defer h.hangingMu.Unlock()
+	h.hanging = true
+}
+
+func (h *hangingDB) StoreLog(log *raft.Log) error {
+	if h.currentlyHanging() {
+		time.Sleep(24 * time.Hour)
+	}
+	return h.LogStore.StoreLog(log)
+}
+
+func (h *hangingDB) StoreLogs(logs []*raft.Log) error {
+	if h.currentlyHanging() {
+		time.Sleep(24 * time.Hour)
+	}
+	return h.LogStore.StoreLogs(logs)
+}
+
+func newHangingDB(logStore raft.LogStore) *hangingDB {
+	h := &hangingDB{LogStore: logStore}
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGUSR1)
+	go func() {
+		sig := <-c
+		fmt.Printf("got signal: %v\n", sig)
+		h.hang()
+	}()
+	return h
+}
+
 func (s *Store) Open(bootstrap bool, localID string) error {
 	config := raft.DefaultConfig()
 	config.LocalID = raft.ServerID(localID)
@@ -73,7 +122,7 @@ func (s *Store) Open(bootstrap bool, localID string) error {
 	}
 
 	// raft system
-	r, err := raft.NewRaft(config, s.fsm, boltDB, boltDB, ss, transport)
+	r, err := raft.NewRaft(config, s.fsm, newHangingDB(boltDB), boltDB, ss, transport)
 	if err != nil {
 		return err
 	}

```

Then, after bringing up 3 nodes as described in the README, I send SIGUSR1 to the leader node, and now all commands hang, but no re-election happens.

The same is reproducible in https://robustirc.net/ with the robustirc-localnet command, but might be a bit more elaborate to test than the more self-contained leto example.

Given that it happens in two different projects built on top of hashicorp/raft, I don’t think it’s a bug with my code itself, but maybe both projects are using hashicorp/raft slightly wrong?

Any recommendations for how to handle this failure mode?

### LINKED ISSUE #612: Leader stopped sending appendEntries to the follower, but heartbeat was working
Hi, Thanks for the great library! I really appreciate the efforts you guys have taken to build this project.

Issue Description:
We recently encountered an issue, leader stopped sending appendEntries to the follower but could send the heartbeat.
The screenshot below, taken from the metrics,  shows the situation. Only one follower node was affected, and all other nodes were fine.

AppendEntried_heartbeat

Let me describe the issue in more detail, and share my findings so far on the issue.
```
At 09:44:36.106, Follower tries to storeLogs from Index: 151153297 to 151153341.
At 09:44:36.115, Follower received the error, "failed to append to logs error=unable to store logs in logStore, 
                          err=Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction"
                          // note that our logStore is backed by MySQL. 
At 09:44:36.126, On Follower, raft: failed to get previous log previous-index=151153341 last-index=151153296 error=log not found
At 09:44:36.136, On Follower, raft: failed to get previous log previous-index=151153342 last-index=151153296 error=log not found
```

The above timeline represents the events that occur at the time of the issue. At the same time, I didn't get any warn/error from the leader.

Following is the Stats on the Follower node at 9.45:29.236
```
{
	"applied_index":"151153288",
	"commit_index":"151153288",
	"fsm_pending":"0",
	"last_contact":"40.69452ms",
	"last_log_index":"151153296",
	"last_log_term":"34105",
	"last_snapshot_index":"151134079",
	"last_snapshot_term":"34105",
	"latest_configuration":"<Sorry, I can't share this. But I've double checked the follower node is available in the list>",
	"latest_configuration_index":"0",
	"num_peers":"8",
	"protocol_version":"3",
	"protocol_version_max":"3",
	"protocol_version_min":"0",
	"server_id":"FollowerNode",
	"snapshot_version_max":"1",
	"snapshot_version_min":"0",
	"state":"Follower",
	"term":"34105"
}
```

I'm referring to the code below,
```
    // Update s based on success
	if resp.Success {
		// Update our replication state
		updateLastAppended(s, &req)

		// Clear any failures, allow pipelining
		s.failures = 0
		s.allowPipeline = true
	} else {
		atomic.StoreUint64(&s.nextIndex, max(min(s.nextIndex-1, resp.LastLog+1), 1))
		if resp.NoRetryBackoff {
			s.failures = 0
		} else {
			s.failures++
		}
		r.logger.Warn("appendEntries rejected, sending older logs", "peer", peer, "next", atomic.LoadUint64(&s.nextIndex))
	}
```

I've the following questions.
1. When "failed to append to logs" error occurred on the follower node, on the leader `atomic.StoreUint64(&s.nextIndex, max(min(s.nextIndex-1, resp.LastLog+1), 1))` should set `s.nextIndex`
to 151153297, and try to replicate again from Index=151153297. But in this case "appendEntries rejected, sending older logs" error didn't occur on the leader node. Does that mean the leader received `resp.Success=true` when an error occurred?

2. Does the leader consider it has replicated logs till Index 151153341 at the follower node? As from the logs on the follower, it is looking for previousLog at index=151153341

3. I don't understand the reason why replication stopped from Leader to Follower, i don't see "removed peer, stopping replication" in the logs on the leader node.

Note: when the leadershipTransfer happened again, a new leader was able to restart the replication.

### LINKED ISSUE #614: Node unable to persist log, but keeps being elected
Hello,

Recently, we faced an issue where we encounter a case where the leader node had a bad persistent store:
- leader node was unable to persist logs -> demotes itself -> starts election -> becomes leader -> unable to persists logs -> ... (the cycle repeats)
  - Demotion is caused by https://github.com/hashicorp/raft/blob/main/raft.go#L1271

This cycle caused unstable leadership during the period. For us, this cycle persisted for 10 mins until another node was finally elected leader.

Wondering if there are recommendations or good practices for handling such cases? Given that Hashicorp runs your own cloud offerings too.

Also, wondering if there is an optimisation that we can do here in the library? I understand there's some nuances to this. 
- Based on my understanding, the current way to fend against this is that heartbeat timeouts has a form of randomness. 
- Given a cluster: node A (leader), node B, node C:
  - When node A demotes itself, because of randomness, node C has a chance to timeout earlier and becomes a candidate before node A becomes one
  - However, if node B doesn't timeout, it will still think that node A is the leader, and will always reject node C's vote request. 
  - In such cases, node A has the natural advantage in winning elections. This is not preferred when node A has a persistent store issue

tgross:
If the leader's disk stalls, we can't replicate logs to followers. But the heartbeat RPCs will still be sent concurrently, and this resets the heartbeat timeout on the followers. This is intentional in order to avoid leadership flapping in the face of temporary disk stalls, but has the side effect of preventing recovery from a more persistent disk stall.

Have the leader set a timestamp at the start of each replication for each follower, and clear it on either success or failure. The heartbeater will check this timestamp if set, and if has been set for too long, delay the next heartbeat so that the leader can eventually step down.

Fixes: https://github.com/hashicorp/raft/issues/666
Fixes: https://github.com/hashicorp/raft/issues/503
Fixes: https://github.com/hashicorp/raft/issues/612
Fixes: https://github.com/hashicorp/raft/issues/614

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2013, 2025
+// Copyright IBM Corp. 2013, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package raft
@@ -10,7 +10,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/hashicorp/go-metrics/compat"
+	metrics "github.com/hashicorp/go-metrics/compat"
 )
 
 const (
@@ -73,6 +73,14 @@ type followerReplication struct {
 	// lastContactLock protects 'lastContact'.
 	lastContactLock sync.RWMutex
 
+	// lastReplicationStart is updated to the current time whenever a
+	// replicateTo method call is started, and is cleared when the replicateTo
+	// method returns. This is used by heartbeats to check if replication has
+	// stalled for too long.
+	lastReplicationStart time.Time
+	// lastReplicationStartLock protects 'lastReplicationStart'.
+	lastReplicationStartLock sync.RWMutex
+
 	// failures counts the number of failed RPCs since the last success, which is
 	// used to apply backoff.
 	failures uint64
@@ -132,6 +140,30 @@ func (s *followerReplication) setLastContact() {
 	s.lastContactLock.Unlock()
 }
 
+// resetLastReplicationStart clears the marker for the start of the last replication
+// attempt
+func (s *followerReplication) resetLastReplicationStart() {
+	s.lastReplicationStartLock.Lock()
+	s.lastReplicationStart = time.Time{}
+	s.lastReplicationStartLock.Unlock()
+}
+
+// setLastReplicationStart sets the marker for the start of the last replication
+// attempt to the current time
+func (s *followerReplication) setLastReplicationStart() {
+	s.lastReplicationStartLock.Lock()
+	s.lastReplicationStart = time.Now()
+	s.lastReplicationStartLock.Unlock()
+}
+
+// getLastReplicationStart gets the start of the last replication attempt
+func (s *followerReplication) getLastReplicationStart() time.Time {
+	s.lastReplicationStartLock.RLock()
+	t := s.lastReplicationStart
+	s.lastReplicationStartLock.RUnlock()
+	return t
+}
+
 // replicate is a long running routine that replicates log entries to a single
 // follower.
 func (r *Raft) replicate(s *followerReplication) {
@@ -140,17 +172,22 @@ func (r *Raft) replicate(s *followerReplication) {
 	defer close(stopHeartbeat)
 	r.goFunc(func() { r.heartbeat(s, stopHeartbeat) })
 
+	defer s.resetLastReplicationStart()
+
 RPC:
 	shouldStop := false
 	for !shouldStop {
+		s.resetLastReplicationStart()
 		select {
 		case maxIndex := <-s.stopCh:
 			// Make a best effort to replicate up to this index
 			if maxIndex > 0 {
+				s.setLastReplicationStart()
 				r.replicateTo(s, maxIndex)
 			}
 			return
 		case deferErr := <-s.triggerDeferErrorCh:
+			s.setLastReplicationStart()
 			lastLogIdx, _ := r.getLastLog()
 			shouldStop = r.replicateTo(s, lastLogIdx)
 			if !shouldStop {
@@ -159,6 +196,7 @@ RPC:
 				deferErr.respond(fmt.Errorf("replication failed"))
 			}
 		case <-s.triggerCh:
+			s.setLastReplicationStart()
 			lastLogIdx, _ := r.getLastLog()
 			shouldStop = r.replicateTo(s, lastLogIdx)
 		// This is _not_ our heartbeat mechanism but is to ensure
@@ -167,12 +205,14 @@ RPC:
 		// can't do this to keep them unblocked by disk IO on the
 		// follower. See https://github.com/hashicorp/raft/issues/282.
 		case <-randomTimeout(r.config().CommitTimeout):
+			s.setLastReplicationStart()
 			lastLogIdx, _ := r.getLastLog()
 			shouldStop = r.replicateTo(s, lastLogIdx)
 		}
 
 		// If things looks healthy, switch to pipeline mode
 		if !shouldStop && s.allowPipeline {
+			s.resetLastReplicationStart()
 			goto PIPELINE
 		}
 	}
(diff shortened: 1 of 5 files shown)


replication.go
```diff
@@ -73,6 +73,14 @@ type followerReplication struct {
 	// lastContactLock protects 'lastContact'.
 	lastContactLock sync.RWMutex
 
+	// lastReplicationStart is updated to the current time whenever a
+	// replicateTo method call is started, and is cleared when the replicateTo
+	// method returns. This is used by heartbeats to check if replication has
+	// stalled for too long.
+	lastReplicationStart time.Time
+	// lastReplicationStartLock protects 'lastReplicationStart'.
+	lastReplicationStartLock sync.RWMutex
```
schmichael, line 82: Since the critical section is only a load or store, we could use a ` sync.Pointer` instead. Your choice matches existing code (`lastContact`) though, so either way is fine.

replication.go
```diff
@@ -409,6 +449,30 @@ func (r *Raft) heartbeat(s *followerReplication, stopCh chan struct{}) {
 		s.peerLock.RUnlock()
 
 		start := time.Now()
+
+		lastReplicationStart := s.getLastReplicationStart()
+		if !lastReplicationStart.IsZero() {
+			maxLastReplication := r.config().HeartbeatTimeout * 10
```
schmichael, line 455: Let's always comment magic numbers (especially since I can't remember exactly why we picked this one)

```suggestion
			// Replication timeout should be relatively long to avoid
			// costly leadership elections due to temporary replication
			// stalls.
			maxLastReplication := r.config().HeartbeatTimeout * 10
```

maybe?

schmichael (commented):
Apologies if we already covered this and my fuzzy memory just forgot, but it might be nice to find some place to record this in the code itself:

Since the goal is to delay heartbeats in response to leader disk slowness, why is replication the place to do it? Won't that include a lot of other conditions as well such as follower disk stalls and snapshotting? Since `replicateTo` may send an entire snapshot, might this approach cause heartbeat failures during large snapshots?

Apologies if these are naive questions, I'm in this code extremely infrequently!

PR #699, opened 2026-07-15
add deadline in net_transport handleConn

zjj:
- Add read deadline in handleConn to prevent connections from hanging indefinitely

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -619,6 +619,10 @@ func (n *NetworkTransport) handleConn(connCtx context.Context, conn net.Conn) {
 		default:
 		}
 
+		if n.timeout > 0 {
+			conn.SetReadDeadline(time.Now().Add(n.timeout))
+		}
+
 		if err := n.handleCommand(r, dec, enc); err != nil {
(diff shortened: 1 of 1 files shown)


schmichael:
I just noticed this after closing #700! I'm curious why you closed it as at a glance it appears to be a place we don't respect the configured netio timeout.

zjj:
> I just noticed this after closing #700! I'm curious why you closed it as at a glance it appears to be a place we don't respect the configured netio timeout.

Previously, I thought it was just a simple timeout configuration issue, but later it seemed a bit more complicated. It appears that having a TCP long connection blocked on a read operation is inherently fine. Increasing ReadTimeout might only be appropriate when you are absolutely certain that incoming data is guaranteed to arrive. On the other hand, setting a WriteTimeout every time you write data makes total sense. However, when dealing with operations like io.Copy(conn.w, data) or connections wrapped by bufio, the situation seems to become much more complex.

zjj:
@schmichael i created a new pr https://github.com/hashicorp/raft/pull/703

Issue #695, opened 2026-07-09
Data divergence (Safety violation) due to Async Heartbeat


rohanpadhye:
Hello,

When testing this library with [Antithesis](https://antithesis.com/), we encountered a serious data-divergence bug where some nodes on the Raft cluster commit different values to the FSM as compared to other nodes.  On further investigation, we found that the bug violates at least three critical safety properties from Figure 3 of the Raft paper:

- Log Matching ❌
- Leader Completeness ❌
- State Machine Safety ❌

In some cases, the same bug can manifest in a different way, violating another crucial property:
- Election Safety ❌

Version tested: Latest `main` committed on May 19, 2026 (4c8f61ac9255bb95fb3b8319dfcf0ae53ab325b6)

## Impact

**Safety** / **Date Integrity** - Cluster nodes can disagree on log entries with the same index/term, the values of committed entries at a given index, the committed FSM state, and in some cases even who the leader is for a given term.

## Triggers

1. A network partition that causes a new leader to be elected in the majority partition.
2. A network restore that causes the old leader to receive heartbeats from the new leader
3. Concurrently, the old leader to be handling RPCs such as AppendEntries (e.g., servicing a client request) or RequestVote (e.g., another node timing out)

## Symptom

Nodes disagree about the FSM value after some index.

For example, consider a 3-node cluster with a "chain-of-blocks" workload; that is, the FSM simply stores a running hash of all input commands ("blocks" of random data) that are applied to it. One sample symptom from the client's view is as follows:

Applying random blocks in sequence up to index 89, everything is fine:

```
[node1]: 2026/05/11 19:41:53 Applied block 89 (752358328303c56397112eac5446f69e) state: f7c73ec4 => cc97c2fc
[node3]: 2026/05/11 19:41:54 Applied block 89 (752358328303c56397112eac5446f69e) state: f7c73ec4 => cc97c2fc
[node2]: 2026/05/11 19:41:54 Applied block 89 (752358328303c56397112eac5446f69e) state: f7c73ec4 => cc97c2fc
```

At block 90, nodes diverge:
```
[node3]: 2026/05/11 19:42:01 Applied block 90 (2af01d1606f25b6237a659a02bdd8c1f) state: cc97c2fc => c45dc6f9
[node1]: 2026/05/11 19:42:02 Applied block 90 (270c66a1b96b16b17e0bbc2b4deb004b) state: cc97c2fc => 8dbda39d
[...]

PR #701, opened 2026-07-17
Heartbeat term bump races with the leader loop, diverges logs

goingforstudying-ctrl:
Was reading through the Antithesis report in #695 and the async-heartbeat section of docs/divergence.md, and the race finally clicked for me. Figured I'd take a stab at it since it's been sitting open.

The problem: heartbeat AppendEntries get handled on the transport I/O thread (the fast-path, so the election timeout doesn't have to wait on disk latency). But a heartbeat that carries a newer term mutates `currentTerm` and flips the node to Follower just like any other AppendEntries — and that check-and-set runs with no synchronization against the main raft thread. So you can get this interleaving:

1. Node A is leader for term T, takes a client apply, and `dispatchLogs()` reads `getCurrentTerm()` → T.
2. Before A persists the entry, a heartbeat from B (the real leader for term T+1) lands on the I/O thread and bumps `currentTerm` to T+1, sets state to Follower.
3. The main thread doesn't notice and goes ahead writing `(data=X, term=T+1)` and starts replicating it — a term A no longer leads.

Now some nodes apply X at that index, others apply whatever B says, and since everything claims term T+1 the logs commit-diverge. That's the Log Matching / State Machine Safety violation from the issue.

The fix is two parts:

- Serialize the higher-term step-down in `appendEntries()` with a dedicated mutex (`rpcTransitionLock`) so the check-and-set of `(state, currentTerm)` is atomic with respect to both the transport thread and the main thread. This keeps the async heartbeat fast-path — no extra disk stall on the election timeout, which is the whole reason heartbeats are fast-pathed in the first place — it just stops the transition from tearing.
- Belt-and-suspenders in `dispatchLogs()`: re-read the state right before persisting, and if we've been superseded (no longer Leader) refuse to append and answer the future with `ErrLeadershipLost` so the caller retries against the real leader, instead of writing a divergent entry.

I went back and forth on whether the mutex alone was enough. I think it is for the term-tearing itself, but the dispatch guard closes the wider window where the apply was already dequeued before the heartbeat arrived, so I kept both. Happy to drop the second one if folks feel it's redundant.

For tests: `TestRaft_AppendEntries_ConcurrentTermBump` hammers `appendEntries` from 32 goroutines with distinct higher terms and fails on the torn "still Leader but term bumped" state without the lock (I verified it fails on the unpatched code and passes with it). `TestRaft_DispatchLogs_AfterLeadershipLost` pins the dispatch guard. Both run clean under `-race`.

Ran `gofmt`, `go vet`, `go build ./...`, and the raft test suite under `-race` locally on go1.25. Everything green except `TestRaft_LeadershipTransferLeaderReplicationTimeout`, which flaked for me on the base commit too (fails ~3/4 of the time with or without my change — looks timing-related, not related to this path).

This is the same thing reported in #695. I couldn't reproduce the full Antithesis scenario without their harness, so this targets the root cause they identified rather than the end-to-end partition replay.

### DIFF
--- a/api.go
+++ b/api.go
@@ -140,6 +140,21 @@ type Raft struct {
 	// on.
 	candidateFromLeadershipTransfer atomic.Bool
 
+	// rpcTransitionLock serializes the "observe a higher term -> step down to
+	// follower" transition that can be triggered concurrently from two places:
+	// the main raft thread (via processRPC) and the transport I/O thread (via
+	// the async heartbeat fast-path, processHeartbeat). HashiCorp Raft handles
+	// heartbeat AppendEntries on the transport thread to keep the election
+	// timeout decoupled from disk latency, but a heartbeat carrying a newer
+	// term mutates currentTerm and state just like any other AppendEntries.
+	// Without serialization that mutation can interleave with the main thread
+	// in the middle of dispatching a log entry or handling another RPC, which
+	// is exactly the race that produces the data-divergence safety violation
+	// described in https://github.com/hashicorp/raft/issues/695. The lock makes
+	// the check-and-set of (state, currentTerm) atomic with respect to both
+	// callers so a node can never act on a term it has already superseded.
+	rpcTransitionLock sync.Mutex
+
 	// Stores our local server ID, used to avoid sending RPCs to ourself
 	localID ServerID
 
--- a/api.go
+++ b/api.go
@@ -140,19 +140,8 @@ type Raft struct {
 	// on.
 	candidateFromLeadershipTransfer atomic.Bool
 
-	// rpcTransitionLock serializes the "observe a higher term -> step down to
-	// follower" transition that can be triggered concurrently from two places:
-	// the main raft thread (via processRPC) and the transport I/O thread (via
-	// the async heartbeat fast-path, processHeartbeat). HashiCorp Raft handles
-	// heartbeat AppendEntries on the transport thread to keep the election
-	// timeout decoupled from disk latency, but a heartbeat carrying a newer
-	// term mutates currentTerm and state just like any other AppendEntries.
-	// Without serialization that mutation can interleave with the main thread
-	// in the middle of dispatching a log entry or handling another RPC, which
-	// is exactly the race that produces the data-divergence safety violation
-	// described in https://github.com/hashicorp/raft/issues/695. The lock makes
-	// the check-and-set of (state, currentTerm) atomic with respect to both
-	// callers so a node can never act on a term it has already superseded.
+	// rpcTransitionLock serializes term/state transitions between the main
+	// raft thread and the async heartbeat I/O thread.
 	rpcTransitionLock sync.Mutex
 
 	// Stores our local server ID, used to avoid sending RPCs to ourself
--- a/api.go
+++ b/api.go
@@ -140,8 +140,14 @@ type Raft struct {
 	// on.
 	candidateFromLeadershipTransfer atomic.Bool
 
-	// rpcTransitionLock serializes term/state transitions between the main
-	// raft thread and the async heartbeat I/O thread.
+	// rpcTransitionLock serializes transitions of the (state, currentTerm)
+	// pair between the main raft thread and the transport I/O thread, which
+	// can step the node down concurrently via a fast-path heartbeat. Writers
+	// that move state and/or currentTerm as part of one transition must hold
+	// the lock for the whole transition, and appendEntries additionally holds
+	// it across the term check that decides to transition, so concurrent
+	// step-downs cannot interleave. The atomic accessors (getState,
+	// getCurrentTerm) may be read without the lock.
 	rpcTransitionLock sync.Mutex
 
 	// Stores our local server ID, used to avoid sending RPCs to ourself
--- a/raft.go
+++ b/raft.go
@@ -1247,17 +1247,6 @@ func (r *Raft) dispatchLogs(applyLogs []*logFuture) {
 	term := r.getCurrentTerm()
 	lastIndex := r.getLastIndex()
 
-	// A heartbeat on the I/O thread can flip us to Follower mid-dispatch.
-	// Refuse to append under a term we no longer lead.
-	if r.getCurrentTerm() != term {
-		r.logger.Warn("aborting log dispatch, leadership lost before append",
-			"term", term, "currentTerm", r.getCurrentTerm())
-		for _, applyLog := range applyLogs {
-			applyLog.respond(ErrLeadershipLost)
-		}
-		return
-	}
-
 	n := len(applyLogs)
 	logs := make([]*Log, n)
 	metrics.SetGauge([]string{"raft", "leader", "dispatchNumLogs"}, float32(n))
@@ -1489,8 +1478,7 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 	r.rpcTransitionLock.Lock()
 	if a.Term > r.getCurrentTerm() || (r.getState() != Follower && !r.candidateFromLeadershipTransfer.Load()) {
 		// Ensure transition to follower
-		r.setState(Follower)
-		r.setCurrentTerm(a.Term)
+		r.transitionToFollowerLocked(a.Term)
 		resp.Term = a.Term
 	}
 	r.rpcTransitionLock.Unlock()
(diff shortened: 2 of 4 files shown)


raft.go
tgross: The problem with the async heartbeat is, of course, that we're _not_ only calling this from the main thread. This gets called from the transport thread for heartbeats too. The `processHeartbeat` has the same comment.

Because heartbeats _only_ have a `Term` and `Leader` field filled in, it looks like we don't have to worry about tearing in the `a.PrevLogEntry` , `a.Entries`, or `a.LeaderCommitIndex` blocks below because they'll only be executing on the main thread?

raft.go
```diff
@@ -1476,13 +1487,15 @@ func (r *Raft) appendEntries(rpc RPC, a *AppendEntriesRequest) {
 	}
 
 	// Increase the term if we see a newer one, also transition to follower
-	// if we ever get an appendEntries call
+	// if we ever get an appendEntries call.
+	r.rpcTransitionLock.Lock()
 	if a.Term > r.getCurrentTerm() || (r.getState() != Follower && !r.candidateFromLeadershipTransfer.Load()) {
 		// Ensure transition to follower
 		r.setState(Follower)
 		r.setCurrentTerm(a.Term)
```
tgross: There are a _lot_ of other places where we call `r.setState(Follower)` and `r.setCurrentTerm(...)` on the main thread. Can't a heartbeat come in and fail to serialize against any of those, not just `appendEntries`?

tgross (commented):
Thanks for this PR @goingforstudying-ctrl. I'm still reviewing this in more depth but I've taken a first pass thru and had some questions. I'll try to come back with more over the next few working days.

goingforstudying-ctrl:
Thanks for the close look. On the heartbeat threading question: right, with the NetworkTransport the heartbeat fast path calls processHeartbeat (and through it appendEntries) directly from the connection's I/O goroutine, so the "must only be called from the main thread" comments above both functions are stale. The heartbeat request only carries Term and Leader (the heartbeat() constructor in replication.go never sets the other fields), so on the I/O thread the PrevLogEntry, Entries, and LeaderCommitIndex blocks can't execute, and the transition block is the only shared-state write beyond setLeader. That's why the lock sits only around that block.

On the other transition sites: the lock as written doesn't serialize against them. requestVote, the preVote and vote handling in runCandidate, and the leader-loop step-downs all run on the main thread and still do setState(Follower) followed by setCurrentTerm without taking the lock, so a transport-thread heartbeat can interleave with any of those. And because setCurrentTerm persists to stable storage before updating the in-memory term, an interleaving heartbeat can persist a term lower than one the node already observed (for example requestVote bumps to 8 and the heartbeat's older bump lands last, writing 7). The window is narrow, but the invariant isn't airtight.

I can extend the same lock to the other transition sites, or funnel them through one stepDown helper, if that matches the direction you'd prefer. Also happy to update the two stale main-thread comments either way.

raft.go
goingforstudying-ctrl: The main-thread-only comments on appendEntries and processHeartbeat were stale: the fast path calls processHeartbeat from the transport I/O thread. heartbeat() only sets Term and Leader, so the PrevLogEntry/Entries/LeaderCommitIndex blocks can't execute on that path. The transition block is the only other shared-state write there (setLeader takes leaderLock itself), which is why the lock is scoped to that block. Fixed both comments in the latest push.

raft.go
goingforstudying-ctrl: The lock only serialized against the heartbeat path in appendEntries. requestVote, installSnapshot, and runCandidate's preVote/vote handling each did setState(Follower) + setCurrentTerm without it. Since setCurrentTerm persists to stable storage before updating the in-memory term, a racing heartbeat could persist a term lower than one the node already observed. The latest push routes all of them through a transitionToFollower helper under the lock, and serializes electSelf's term bump the same way.

raft_test.go
```diff
@@ -3098,6 +3161,116 @@ func TestRaft_InstallSnapshot_InvalidPeers(t *testing.T) {
...
+	r.leaderState.inflight = list.New()
+	// dispatchLogs also touches leaderState.commitment after storing logs.
+	r.leaderState.commitment = newCommitment(make(chan struct{}), Configuration{}, 0)
+
+	// Hold lastLock so dispatchLogs blocks inside getLastIndex, bump the
+	// term as a concurrent heartbeat would, then unblock it.
+	r.lastLock.Lock()
+	bumped := make(chan struct{})
+	go func() {
+		defer close(bumped)
```
tgross: This construct is usually done with [`sync.WaitGroup.Go`](https://pkg.go.dev/sync#WaitGroup.Go)... but it also doesn't look like it's exercising the code path the comment says it does?  This goroutine could completely exit before we ever hit `r.dispatchLogs`, and we're blocking on `bumped` _after_ we make that call anyways so that doesn't do anything for us. 

This is always an awkward thing to setup a test for, with only two "good" options I've seen:
* We can spin on `TryLock` and keep unlocking it until we get `false` back, or
* We can use [`time.AfterFunc`](https://pkg.go.dev/time#AfterFunc) before we call `r.dispatchLogs` with enough delay that we're always going to hit the lock in `r.dispatchLogs` before the timeout elapses

The second option there is how I typically see this done.

raft_test.go
```diff
@@ -3098,6 +3161,116 @@ func TestRaft_InstallSnapshot_InvalidPeers(t *testing.T) {
...
+		// Wait until dispatchLogs is queued on lastLock, then simulate the
+		// heartbeat term bump and release the lock so the guard re-reads
+		// the new term.
+		r.setCurrentTerm(4)
+		r.setState(Follower)
```
tgross: Ideally we'd use the `transitionToFollower` here

api.go
```diff
@@ -140,6 +140,10 @@ type Raft struct {
 	// on.
 	candidateFromLeadershipTransfer atomic.Bool
 
+	// rpcTransitionLock serializes term/state transitions between the main
+	// raft thread and the async heartbeat I/O thread.
+	rpcTransitionLock sync.Mutex
```
tgross: I generally like to have a mutex call out which specific fields need to be accessed with the mutex in place. Right now the implementation is ambiguous about whether this needs to be held when setting the term or the state or only when setting both. 

Ex. we hold it over `getCurrentTerm()` and `setCurrentTerm()` in some places but not others, even though `r.currentTerm` is an atomic field. That we have `setCurrentTerm()` implemented on `Raft` and touching both disk and `RaftState`, whereas `getCurrentTerm()` is only on the `RaftState` is kind confusing here, so the locks are even harder to puzzle out.

raft.go
```diff
@@ -1249,6 +1247,17 @@ func (r *Raft) dispatchLogs(applyLogs []*logFuture) {
...
+	// A heartbeat on the I/O thread can flip us to Follower mid-dispatch.
+	// Refuse to append under a term we no longer lead.
+	if r.getCurrentTerm() != term {
+		r.logger.Warn("aborting log dispatch, leadership lost before append",
+			"term", term, "currentTerm", r.getCurrentTerm())
+		for _, applyLog := range applyLogs {
+			applyLog.respond(ErrLeadershipLost)
+		}
+		return
+	}
```
tgross: Both `term := getCurrentTerm()` and `lastIndex := getLastIndex()` are just reading in-memory atomic values. So we're comparing the term against the term we got just an instant before without interleaving disk accesses that could make it meaningful. I don't think this block is actually earning its keep.

goingforstudying-ctrl: Removed. Both reads are adjacent in-memory atomics and the test only exposed the race window by wedging lastLock, so the block was not catching a real race. term and lastIndex stay since the append path uses them; TestRaft_DispatchLogs_AfterLeadershipLost was deleted with the block.

raft_test.go
goingforstudying-ctrl: There was no ordering guarantee: the goroutine could bump the term and release lastLock before dispatchLogs ever reached it, and the <-bumped after the call synchronized nothing. The guard is gone in 49e9ce9, so I removed the test with it.

raft_test.go
goingforstudying-ctrl: Done. 49e9ce9 routes the appendEntries step-down through transitionToFollowerLocked while holding rpcTransitionLock across the term check.

api.go
goingforstudying-ctrl: Done. The comment now spells out that the lock serializes the (state, currentTerm) pair, that writers must hold it for the whole transition including the appendEntries term check, and that the atomic accessors (getState, getCurrentTerm) don't need it.

